PR #129, opened 2013-12-07
updated the benchmarks of the join and string concatenation

yorkie:
see https://gist.github.com/yorkie/7837575, please.

### DIFF
--- a/README.md
+++ b/README.md
@@ -214,8 +214,11 @@
       'with this, you would get nowhere ' +
       'fast.';
     ```
-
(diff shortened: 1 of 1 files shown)


ljharb:
Given that this is a style guide, and performance should be both the very least and last concern for any codebase, and also, should cause style guides to be ignored as necessary to achieve optimization goals, I don't think performance considerations are appropriate to mention in the guide.

Thanks for your contribution!

PR #158, opened 2014-04-24
Remove unnecessary extra newline that confused some markdown previewers

unkillbob:
The markdown previewer for Atom.io didn't like the double newline inside the code block, it treated it as 2 separate code blocks (which threw out formatting for the rest of the file).

Github seems to handle it, but the extra newline seemed redundant anyway.

### DIFF
--- a/README.md
+++ b/README.md
@@ -137,7 +137,6 @@
     ```javascript
     var someStack = [];
 
-
     // bad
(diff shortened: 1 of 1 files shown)


iamstarkov:
prob its better to create issue for atom.io md previewer, @unkillbob what do you think?

goatslacker:
Good to do both.

PR #161, opened 2014-04-24
Note on escaping single quotes in strings.

unkillbob:


ljharb:
The style guide currently recommends using ES6 template string literals for this case. In general, you'd also want to use typographically correct curly quotes (`“”` / `‘’`) which wouldn't require escaping.

PR #162, opened 2014-04-24
Avoid unnecessary else statements when returning within an if clause.

unkillbob:


evanrs:
How do you feel about ternaries?
```javascript
  return someCondition ? something : somethingElse;
```

I limit my use of  `return` to the very beginning and end of a function.
```javascript
function someMutation(val1, val2) {
  if(!val1 || !val2) { return false; }

  // …
  // logic logic logic
  // …

  return someValue;
}
```
When it's a callback:
```javascript
function someCallback(err, val, done) {
  if (err) { return done(err); }
  if (!val) { return done(new Error("Missing value")); }

  // …
  // logic logic logic
  // …

  return done(null, someValue);
}
```

I find this simplifies how the function can be written. But, I tend to employ a ternary to vet the final return value.
``` javascript
function (...args) {
 // some initial exit conditions
 // some logic yielding a new value
 return isThisRight ? yup : nope;
}
```

Does this seem reasonable?

netpoetica:
> I limit my use of return to the very beginning and end of a function.

I agree with this - I usually leave some comment like `// bail when ____`, usually referring to this as a bailout.

Again, the ternary looks nice now but in evolving software, I think it will get ugly. This isn't a perfect example (as in, not realistic) but illustrative of the point:

```js
function buildPerson() {
  var bPersonCreated = false;
  // do stuff
  return bPersonCreated ? person : null;
}
```
New Requirement: If a Person is not created return an error object
```js
function buildPerson() {
  var bPersonCreated = false;
  // do stuff
  return bPersonCreated ? person : new Error('Person was not created.');
}
```
And so on, and so forth, until you end up with something like this:
```js
function buildPerson() {
  var bPersonCreated = false;
  // do stuff
  return bPersonCreated ? {
   status: "success",
   error: null,
   handle: person
  } : { 
    status: "failure",
    handle: null,
    error: new Error('Person was not created.')
  };
}
```

hshoff:
> I don't think this should be an always-or-never case.

Going to agree with @ssorallen from over an year ago. Fun discussion, though! Happy to re-open this if it picks up again.

ssorallen:
@hshoff :beers:

PR #164, opened 2014-04-24
Justify using a leading semi colon before an IIFE.

unkillbob:
Add justification and some further reading on using a leading semi colon before an IIFE to the Semicolons section. This wasn't immediately obvious to me so thought it could do with the extra detail.

### DIFF
--- a/README.md
+++ b/README.md
@@ -907,12 +907,14 @@
       return name;
     })();
 
-    // good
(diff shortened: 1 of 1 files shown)


ArmorDarks:
In 2017 good probably should sound like `stop mindlessly concatenating JS files, use JS builders like Webpack or JSPM for goodness sake and never worry about this issue!`

PR #217, opened 2014-11-20
Update README.md

markpop:
add Loop


hshoff:
Hi @markpop, I'm going to close this in favor of https://github.com/airbnb/javascript#11.1 which says:

> Don't use iterators. Prefer JavaScript's higher-order functions like map() and reduce() instead of loops like for-of.
>
> Why? This enforces our immutable rule. Dealing with pure functions that return values is easier to reason about than side-effects.

Sorry for the super long delay on this! :beers:

PR #226, opened 2014-12-30
add named methods on prototype declarations

k-yak:


ljharb:
@k-yak can you check the "allow edits" checkbox on the right hand side of this PR?

PR #229, opened 2015-01-22
Punctuation tweak

ClimbsRocks:
The which clause depends on and logically flows as part of the sentence above it, so I joined them: 

image

image


hshoff:
:+1: 

Thanks for checking out the guide!

PR #230, opened 2015-01-23
put the dot with fn name when chaining

justjake:
Reasons:
- styleguide already uses leading-dot style.
- leading dot emphasizes that the line is a method call, not a new statement. Consider the readability without whitespace:
```javascript
foo.
something().
else(); // <-- github JS parser error!

foo
.something()
.else();
```
- to spite @horaceko :snowman: 
- no one in semicolon-terminated languages would ever use the trailing-dot style.

### DIFF
--- a/README.md
+++ b/README.md
@@ -814,11 +814,20 @@
     })(this);↵
     ```
 
-  - Use indentation when making long method chains.
+  - Use indentation when making long method chains. Use a leading dot, which
+    emphasizes that the line is a method call, not a new statement.
 
     ```javascript
     // bad
     $('#items').find('.selected').highlight().end().find('.open').updateCount();
+
+    // bad
+    $('#items').
+      find('selected').
+        highlight().
+        end().
+      find('.open').
+        updateCount();
 
     // good
     $('#items')


iamnirav:
:+1: also to spite @horaceko

horaceko:
:-1: to this PR but :+1: to spite @horaceko

nomcopter:
:+1:

hshoff:
@justjake could you change `Put the dot with the function name.` to `Put the dot with the function name.?

justjake:
@hshoff k

hshoff:
:beers:

rafaelsales:
I think both options should be valid. Ruby style guide consider both ok:

https://github.com/bbatsov/ruby-style-guide#consistent-multi-line-chains
```ruby
 Adopt a consistent multi-line method chaining style. There are two popular styles in the Ruby community, both of which are considered good—leading . (Option A) and trailing . (Option B). [link]

(Option A) When continuing a chained method invocation on another line keep the . on the second line.

# bad - need to consult first line to understand second line
one.two.three.
  four

# good - it's immediately clear what's going on the second line
one.two.three
  .four
(Option B) When continuing a chained method invocation on another line, include the . on the first line to indicate that the expression continues.

# bad - need to read ahead to the second line to know that the chain continues
one.two.three
  .four

# good - it's immediately clear that the expression continues beyond the first line
one.two.three.
  four
A discussion on the merits of both alternative styles can be found [here](https://github.com/bbatsov/ruby-style-guide/pull/176).
```

Two reasons I like using dot in the first line of each line:

  1. When pasting code, it works, because on a new line, the REPL knows the sentence did not end yet. You can reproduce that by pasting the following code in node:

  ```js
  const foo = { something: function() { return { else: function () {} } } }
  foo.
   something().
   else()
  ```

  2. When indentation is wrong, it's easier to see the sentence continues in the next line.

ljharb:
@rafaelsales idiomatic ruby styles can be problematic for a number of reasons, so that's not a point in its favor.

Our guide requires either a) everything fits on one line, or b) each thing goes on its own line.

In other words,
```
one.two.three
  .four
```

is invalid - it must either be `one.two.three.four` or:

```
one
  .two
  .three
  .four
```

In addition, our guide and linter rules *require* the dot to begin the next line - we never permit the dot to trail on the end.

rafaelsales:
You could have summarized your answer in one line: "Our rule is the one you can find in the readme"

I was trying to start a discussion here showing the other style also has advantages over the current one, thus shouldn't be considered "bad" as the README says. Instead it should be offered as just another option certain teams can adopt and still be able to say "we follow airbnb styleguide"

ljharb:
Fair enough :-)

I don't think that the other style has advantages, and I highly dislike it in Ruby as well. As far as I'm aware, the only reason that style became popular in Ruby is because some old version of Ruby or earlier (i forget which) didn't support the "leading dots" style. Now that Ruby supports it, I think that's the style that should be used in both JS and Ruby.

As to your specific reasons:
 1. REPL use cases are not common, and should not be optimized for. That might be a nice perk of "trailing dots", but it's not something that overrides the readability of "leading dots".
 1. The linter catches when indentation is wrong, so there's no need to inflict that burden on humans.

If someone wants to be able to say "we follow the airbnb style guide", they have to cleave to our rules - and that includes using "leading dots".

PR #234, opened 2015-02-01
add Apartmint to list of organizations

davychiu:


hshoff:
Thanks for checking out the guide! :beers:

PR #235, opened 2015-02-02
Combine variable declarations.

DannyMoerkerke:
Perhaps it would be an idea to combine variable declarations, for example

    var foo = 1;
    var bar = 2;

would then become:

    var foo = 1,
        bar = 2;


nkbt:
@DannyMoerkerke see https://github.com/airbnb/javascript/pull/220

DannyMoerkerke:
Makes sense, thanx!

goatslacker:
Related #245

PR #237, opened 2015-02-04
Update string concatenation, using \ is good instead of bad

Fonger:
According to the test and the links provided in the document, longer string concatenation is very slow.
You should `\` to improve the performance.
However the good example is using `+` instead of `\`. This is obviously an error.


goatslacker:
String concatenation is preferred over slash because spaces after the slash could cause a silent bug. If you really want multi-line strings you should be using template literals.

Also, concatenating strings isn't very slow, it's quite fast. Unless you're writing a mission critical quicksort algorithm there are other micro-optimizations you could be using rather than this one which may lead to unexpected bugs.

PR #239, opened 2015-02-08
Adding comments about spacing between blocks

TheSavior:
All of the examples in the style guide (and most javascript on the web) follow this pattern already, but I haven't seen it officially defined.

### DIFF
--- a/README.md
+++ b/README.md
@@ -858,28 +858,29 @@
 
     ```javascript
     // bad
-    if (true) {
-      var a = 2;
+    if (foo) {
+      return bar;
     }
-    var b = 3;
+    return baz;
 
     // good
-    if (true) {
-      var a = 2;
+    if (foo) {
+      return bar;
     }
-    var b = 3;
+
+    return baz
 
     // bad
-    var o = {
+    var obj = {
       foo: function() {
       },
       bar: function() {
       }
     };
-    return o;
+    return obj;
 
     // good
-    var o = {
+    var obj = {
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -854,6 +854,43 @@
...
+    ```javascript
+    // bad
+    if (true) {
+      var a = 2;
+    }
+    var b = 3;
+
+    // good
+    if (true) {
+      var a = 2;
```
nkbt: 1. Not a good idea to declare a var inside if block under `good` sign
2. Actually, no difference from `bad`

TheSavior: Sure, It's a bad example. Haha, I can fix that, but the point here is to discuss the spacing between lines after the closing curly braces. How do you feel about that? :)

nkbt: I personally prefer to separate logical blocks with an extra line and functions with two. So I would support this in general. The only issue is to figure out what that logical block is and how to generalize the recommendation.
I also consider this to be very minor and non-strict recommendation, so never raise it on code reviews. Unless it is a very bad case.

TheSavior: Agreed on it being pretty minor. I implemented it for JSCS and it will be landing in the next version which should help to reason about the rule as well.

You can view the tests for the rule here:
https://github.com/TheSavior/node-jscs/blob/require-padding-newlines-after-block-declarations/test/rules/require-padding-newlines-after-blocks.js

hshoff:
:+1:

PR #242, opened 2015-02-12
(Fix) Add semicolon to whitespace code example

omarduarte:
# About
A semicolon appears to be unintentionally left out of a syntax code example related to 'Whitespace: Leave a blank line after blocks and before the next statement'. 

## In this commit
Add semicolon at the end of line 871's statement.

### DIFF
--- a/README.md
+++ b/README.md
@@ -868,7 +868,7 @@
       return bar;
     }
 
-    return baz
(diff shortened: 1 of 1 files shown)


hshoff:
:+1:

PR #246, opened 2015-02-22
Added entry for self invoking constructor

sarbbottam:
to facilitate invoking constructors with or without new keyword

### DIFF
--- a/README.md
+++ b/README.md
@@ -1324,6 +1324,22 @@
     };
     ```
 
+  - Self invoking constructors to facilitate invoking constructors with or without ``new`` keyword.
+
+    ```javascript
+    function Jedi(options) {
+      options || (options = {});
+      if (!(this instanceof Jedi)) {
+        return new Jedi(options);
+      }
+      this.name = options.name || 'no name';
+    }
+
+    var jedi = new Jedi({name: 'Jedi'});
+
+    var jedi = Jedi({name: 'Jedi'})
+    ```
+
 **[⬆ back to top](#table-of-contents)**
 
 


README.md
```diff
@@ -1324,6 +1324,22 @@
     };
     ```
 
+  - Self invoking constructors to facilitate invoking constructors with or without ``new`` keyword.
+
+    ```javascript
+    function Jedi(options) {
+      options || (options = {});
+      if (!(this instanceof Jedi)) {
```
goatslacker: This is not something one should get used to.

When you start using ES6 classes (which we are) calling class Jedi as `Jedi({name: 'Jedi'})` will throw an Error.

sarbbottam: Thanks @goatslacker, was not aware of the [Cannot call a class as a function](https://babeljs.io/repl/#?experimental=true&playground=true&evaluate=true&loose=false&spec=false&code=class%20Employee%20%7B%0A%20%0A%20%20constructor(name)%20%7B%0A%20%20%20%20this.name%20%3D%20name%3B%0A%20%20%7D%0A%0A%20%20get%20name()%20%7B%0A%20%20%20%20return%20this._name.toUpperCase()%3B%0A%20%20%7D%0A%0A%20%20set%20name(name)%7B%0A%20%20%20%20if(name)%7B%20%0A%20%20%20%20%20%20this.name%20%3D%20name%3B%0A%20%20%20%20%7D%0A%20%20%7D%0A%7D%0A%0Avar%20employee%20%3D%20Employee('John')%3B%0A) error.

PR #249, opened 2015-03-11
If your file exports a single class, your filename should be exactly the name of the class.

justjake:
@hshoff @spikebrehm @goatslacker


spikebrehm:
:+1:

ascott:
:ship:

PR #252, opened 2015-03-16
Fixes #251 - Rename Conditional Expressions section

### LINKED ISSUE #251: Conditional Expressions & Equality section should be named Comparison operators
Section 9 is named Conditional Expressions & Equality, however, its subject matter seems to be covering comparison operators. 

From the MDN:
Comparison operators: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Comparison_Operators 
Conditional expressions: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Conditional_Operator

fvanderwielen:
As per issue #251, renames section Conditional Expressions & Equality to Comparison Operators & Equality, as the section is not about the Elvis operator :)

### DIFF
--- a/README.md
+++ b/README.md
@@ -15,7 +15,7 @@
   1. [Properties](#properties)
   1. [Variables](#variables)
   1. [Hoisting](#hoisting)
-  1. [Conditional Expressions & Equality](#conditional-expressions--equality)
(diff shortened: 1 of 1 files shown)


goatslacker:
Thanks!

PR #260, opened 2015-03-27
String concat code blocks were not equivalent.

joeframbach:
If the array is empty, the "bad" code will return `<ul></ul>` and the "good" code will return `<ul><li></li></ul>`.

Although I would prefer this over the for loop:

    function inbox(messages) {
      items = messages.map(function(message) {
        return '<li>' + message.message + '</li>';
      });

      return '<ul>' + items.join('') + '</ul>';

### DIFF
--- a/README.md
+++ b/README.md
@@ -250,10 +250,10 @@
       items = [];
 
       for (i = 0; i < length; i++) {
-        items[i] = messages[i].message;
+        items[i] = '<li>' + messages[i].message + '</li>';
       }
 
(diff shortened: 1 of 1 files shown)


goatslacker:
:+1: for no for-loop.

goatslacker:
@hshoff keep it for-loopy or change to a more functional approach?

hshoff:
Would love to go functional approach with `.map` but we still support IE8 and we'd have to talk about adding a polyfill or something similar. Let's keep the for loop for now and avoid the for loops in the ES6 guide.

Issue #136, opened 2014-01-06
Unclear code example in variables section


Daniel-Hug:
At the bottom of the [variables section](https://github.com/airbnb/javascript#variables), this code example is given:

```
// bad
function() {
  var name = getName();

  if (!arguments.length) {
    return false;
  }

  return true;
}

// good
function() {
  if (!arguments.length) {
    return false;
  }

  var name = getName();

  return true;
}
```

At the top of that code block it says:
> Assign variables at the top of their scope. This helps avoid issues with variable declaration and assignment hoisting related issues.

It's not clear what point the code example is trying to make, as it doesn't reflect the statement made in that comment. Plus, the `name` variable is defined but never used.

goatslacker:
Closing in favor of #265

PR #265, opened 2015-03-28
Fix examples in README.md

wcwung:
The 'bad' and 'good' labels were wrongly placed for the variable declaration examples. An `if` statement before the variable declaration was labeled as 'good' while the following example with a variable declaration before the `if` statement was labeled as 'bad'.

### DIFF
--- a/README.md
+++ b/README.md
@@ -446,25 +446,25 @@
 
       return name;
     }
-
+    
     // bad
     function() {
-      var name = getName();
-
       if (!arguments.length) {
         return false;
       }
+
+      var name = getName();
 
       return true;
     }
 
     // good
     function() {
+      var name = getName();
+
       if (!arguments.length) {
         return false;
       }
-
-      var name = getName();
 
       return true;
     }


garrettjoecox:
@wcwung +1

goatslacker:
I think this might be intentional actually. cc @hshoff

README.md
```diff
@@ -446,26 +446,26 @@
 
       return name;
     }
-
+
```
goatslacker: remove trailing whitespace

hshoff:
@goatslacker is correct, this was intentional. It's just missing an explanation. It was supposed to say declare variables underneath short-circuit/early returns.

`getName` gets evaluated here for no reason here:
```js
// bad
function() {
  var name = getName();

  if (!arguments.length) {
    return false;
  }

  return true;
}
```

Either of the following is fine, though I tend to prefer the first:

```js
// good
function() {
  if (!arguments.length) {
    return false;
  }

  var name = getName();

  return true;
}

// good
function() {
  var name;

  if (!arguments.length) {
    return false;
  }

  name = getName();

  return true;
}
```

hshoff:
After reading through https://github.com/airbnb/javascript/issues/136 again, I think we should just remove the oversimplified example of early returning without arguments.

jakeoverall:
@hshoff I think the example should be left in with the explanation of variable declaration vs variable assignment. 

As a general rule of explanation variables should be declared at the top of their scope even if not assigned since they get hoisted anyways. 

It is a bad habit to get in by declaring them later as this can cause issues. If I was doing code review and saw something like this I would correct it. 

The declaration will occur either way however the assignment will not be made uselessly if arguments are provided.

nkbt:
ES6 is on the way. Always use `let` or better `const` and forget about this awkward rule.

goatslacker:
I was going to make a comment about how irrelevant this is going to be soon but then again we plan on supporting ES5 for those folks who want an ES5 only style guide.

PR #268, opened 2015-03-30
Resolves issue #136

jakeoverall:
Unclear code example in variables section. Variables should be declared at the top of their scope to avoid declaration hoisting.

### DIFF
--- a/README.md
+++ b/README.md
@@ -446,7 +446,9 @@
 
       return name;
     }
-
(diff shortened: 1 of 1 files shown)


hshoff:
Hi @jakeoverall thanks for the PR. We're discussing this over in https://github.com/airbnb/javascript/pull/265

Issue #259, opened 2015-03-26
.eslintrc


iamstarkov:
There are _.jscsrc_ and _.jshintrc_ in the wild performed in Airbnb style. Is there any _.eslintrc_ implementation?

justin808:
Wondering the same thing, especially if one will support ES6/JSX.

RobLoach:
Related: https://github.com/airbnb/javascript/issues/247

justin808:
FYI:  I just converted this tutorial project to use the current style guide with eslint, ES6, and ReactJs
https://github.com/justin808/react-webpack-rails-tutorial/pull/25. You can see the toolchain of npm packages to make eslint work. The only rule I disagree with is a 80 char line limit. 100 seems much more reasonable.

ziflex:
so, where can I get config file with these rules? :)

nkbt:
@ziflex apparently https://github.com/airbnb/javascript/blob/master/linters/.eslintrc

ziflex:
@nkbt ooops, sorry, didn't see :(

PR #272, opened 2015-04-01
Add ESLint support

RobLoach:
[ESLint](http://eslint.org/) is an alternative to JSHint. This is just a rough start, feel free to add more.

References: https://github.com/airbnb/javascript/issues/259

### DIFF
--- a/linters/eslintrc
+++ b/linters/eslintrc
@@ -1,8 +1,6 @@
 env: 
   browser: true
   node: true
-  amd: true
-  jquery: true
 
 rules:
   no-new-object: 2
--- a/linters/eslintrc
+++ b/linters/eslintrc
@@ -33,6 +33,7 @@ rules:
   space-before-function-paren: [2, "never"]
   no-spaced-func: 2
   no-wrap-func: 2
+  one-var: [2, "never"]
   camelcase: 2
   new-cap: 2
   consistent-this: [1, "_this"]
--- a/linters/eslintrc
+++ b/linters/eslintrc
@@ -1,0 +1,44 @@
+env: 
+  browser: true
+  node: true
+  amd: true
+  jquery: true
+
+rules:
+  no-new-object: 2
+  no-reserved-keys: 2
+  no-array-constructor: 2
+  quotes: [2, "single"]
+  indent: [2, 2]
+  no-multi-str: 2
+  no-multi-spaces: 2
+  no-shadow-restricted-names: 2
+  dot-notation: 2
+  block-scoped-var: 2
+  vars-on-top: 2
+  no-unused-vars: 2
+  eqeqeq: 2
+  use-isnan: 2
+  no-eq-null: 2
+  brace-style: 2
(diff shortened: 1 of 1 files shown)


linters/eslintrc
```diff
@@ -0,0 +1,44 @@
+env: 
+  browser: true
+  node: true
+  amd: true
+  jquery: true
```
nkbt: I reckon these things trigger errors, when you are trying to do `var $ = require('jquery')` (fails because you are trying to re-define global `$`). So I usually keep all 3rd party libs as `false` even when I actually have them.

RobLoach: Should we just remove the `env` key entirely?

goatslacker: we can leave env with browser and node.

hshoff: Let's leave `browser` and `node`

hshoff:
LGTM after the change to `env`. Thanks for this  @RobLoach!

nkbt:
I've just realized (when pasted to my `.eslintrc`), that this is some funky non-json format. I would expect it to be something like this instead:

```json
{
  "env": {
    "browser": true,
    "node": true
  },

  "rules": {
    "no-new-object": 2,
    "no-reserved-keys": 2,
    "no-array-constructor": 2,
    "quotes": [2, "single"],
    "indent": [2, 2],
    "no-multi-str": 2,
    "no-multi-spaces": 2,
    "no-shadow-restricted-names": 2,
    "dot-notation": 2,
    "block-scoped-var": 2,
    "vars-on-top": 2,
    "no-unused-vars": 2,
    "eqeqeq": 2,
    "use-isnan": 2,
    "no-eq-null": 2,
    "brace-style": 2,
    "spaced-line-comment": 2,
    "valid-jsdoc": 2,
    "space-infix-ops": 2,
    "eol-last": 2,
    "padded-blocks": [2, "never"],
    "space-before-blocks": 2,
    "comma-dangle": [2, "never"],
    "comma-style": 2,
    "semi": [2, "always"],
    "semi-spacing": 2,
    "no-extra-boolean-cast": 2,
    "space-before-function-paren": [2, "never"],
    "no-spaced-func": 2,
    "no-wrap-func": 2,
    "camelcase": 2,
    "new-cap": 2,
    "consistent-this": [1, "_this"],
    "func-names": 2,
    "key-spacing": 2,
    "no-dupe-keys": 2,
    "strict": 2,
  }
}
```

nkbt:
Missing:
`  "one-var": [2, "never"],`

RobLoach:
> I've just realized (when pasted to my .eslintrc), that this is some funky non-json format. I would expect it to be something like this instead:

ESLint supports either [YAML](http://yaml.org/) (via [js-yaml](https://github.com/nodeca/js-yaml)), or JSON. Both formats produce the same result.

> Missing:
    "one-var": [2, "never"],

Added `one-var` in there, thanks for the note :+1: .

hshoff:
:clap:

PR #278, opened 2015-04-03
Iterators and Generators: sum should be 15, not 10

caseyjenks:
5+4+3+2+1 = 15

Great guide by the way! Loving it so far as I'm getting more into Ember CLI and ES6 over the last few months.

### DIFF
--- a/README.md
+++ b/README.md
@@ -786,16 +786,16 @@
       sum += num;
     }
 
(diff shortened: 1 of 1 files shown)


goatslacker:
Thanks! Good catch. Such bad math.

PR #284, opened 2015-04-04
Change wording in Iterators and Generators

lukewestby:
I found the term "function constructors" to be confusing as the functions the section refers to are not constructors. I think "higher-order functions" or something similar would be more clear.

Thanks!

### DIFF
--- a/README.md
+++ b/README.md
@@ -767,7 +767,7 @@
 
 ## Iterators and Generators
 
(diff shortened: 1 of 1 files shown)


goatslacker:
Oh geez that's supposed to say functional constructs.

lukewestby:
That sounds good to me. Thanks for taking a look!

PR #289, opened 2015-04-05
adding a name to reference for TODO or FIXME

sh6khan:


README.md
```diff
@@ -729,12 +729,14 @@
...
 
+  - Include the name of someone you can reference to aid in either the `FIXME` or `TODO`. Usually, if you are the one writing the `TODO` or `FIXME`, you should reference yourself.
+
   - Use `// FIXME:` to annotate problems.
 
     ```javascript
     function Calculator() {
 
-      // FIXME: shouldn't use a global here
+      // FIXME (John Adams): shouldn't use a global here
```
goatslacker: Totally superfluous but can we use different names here?

How about

` Leia Amidala`

and

`Han Solo`

in keeping with the Star Wars theme.

PR #291, opened 2015-04-05
Reference people in TODO and FIXME

sh6khan:


hshoff:
Hi @sh6khan, thanks for checking out the guide! Love that you used Star Wars names. 

The style guide is what Airbnb engineers use internally and we don't follow this practice (we do it optionally). It's certainly helpful and is good practice for some teams/code bases. 

As we can't merge this because Airbnb only does this optionally, I'd recommend forking the guide for your team and adding this rule.

PR #292, opened 2015-04-05
Fix minor punctuation errors in README

sunshinewyin:
(1) BEFORE: Never name a parameter `arguments`, this will take precedence over the `arguments` object.
AFTER: Never name a parameter `arguments`. This will take precedence over the `arguments` object.

(2) BEFORE: Variable declarations get hoisted to the top of their scope, their assignment does not.
AFTER:
Variable declarations get hoisted to the top of their scope, but their assignment does not.

(3) BEFORE: for multiline comments
AFTER: for multi-line comments

### DIFF
--- a/README.md
+++ b/README.md
@@ -301,7 +301,7 @@
     }
     ```
 
-  - Never name a parameter `arguments`, this will take precedence over the `arguments` object that is given to every function scope.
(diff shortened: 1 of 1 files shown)


hshoff:
Thanks! :+1:

PR #298, opened 2015-04-08
Fix inconsistent use of direct assignment in place of Array#push

tesk9:
Style guide on Arrays specifies that one should avoid using direct assignment. Section on using Array#join in place of string concatenation now follows this recommendation.

### DIFF
--- a/README.md
+++ b/README.md
@@ -250,7 +250,7 @@
       items = [];
 
       for (i = 0; i < length; i++) {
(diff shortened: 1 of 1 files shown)


goatslacker:
You probably want to use direct assignment in this particular case since these seems like a micro-optimization and by using `push` you're losing some of the perf benefits.

PR #300, opened 2015-04-09
Add `esnext` to the jsHint configuration

tomekwi:
This way we'll allow ES6 syntax.

### DIFF
--- a/linters/jshintrc
+++ b/linters/jshintrc
@@ -13,6 +13,9 @@
   // Define globals exposed by Node.js.
   "node": true,
 
+  // Allow ES6.
+  "esnext": true,
+
   /*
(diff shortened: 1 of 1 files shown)


goatslacker:
Nice didn't know this was possible, are there ES6 options available to JSHINT? We should update the whole file.

tomekwi:
What I usually do is uncheck [`"strict": true`](https://github.com/tomekwi/javascript/blob/f37a4155b489f44476c372fb3d300fdafb531041/linters/jshintrc#L47) – because strict mode is implied in ES6 modules.

It works OK 99% of times, though they do have some quirks with parsing ES6 here and there.

hshoff:
:+1:

PR #303, opened 2015-04-11
Fix hoisting examples

tomekwi:
Hoisting only happens to `var`. As kangax describes in [Why `typeof` is no longer “safe”](http://es-discourse.com/t/why-typeof-is-no-longer-safe/15/3).

Also, I wonder if we need the whole section, since you recommend to [avoid using `var` altogether](https://github.com/airbnb/javascript/blob/b492/README.md#references).

### DIFF
--- a/README.md
+++ b/README.md
@@ -318,7 +318,7 @@
   - To convert an array-like object to an array, use Array#from.
 
     ```javascript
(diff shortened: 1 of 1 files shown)


tomekwi:
Sorry, there's a typo. I shouldn't have used the GitHub editor. I'll open a new clean issue.

PR #310, opened 2015-04-14
push instead of array index assignment

ck1125:
Think this might be more compliant ;) Nice rules BTW.


goatslacker:
Thanks for your contribution, this was addressed before:

#298

If you'd like we'd accept a pull request which includes a comment on that line why we're using array index rather than pushing.

Issue #313, opened 2015-04-17
Suggestion convert eslintrc to JSON


ButuzGOL:
I copied eslintrc and started do some changes to my project
And find out that I am not quit good with YAML and examples that I found in google is in JSON
So I converted it into JSON and it became much easier

nkbt:
I use this pretty strict one

```javacript
{
  "env": {
    "browser": true,
    "node": true,
    "es6": true
  },

  "plugins": ["react"],

  "ecmaFeatures": {
    "arrowFunctions": true,
    "binaryLiterals": true,
    "blockBindings": true,
    "classes": true,
    "defaultParams": true,
    "destructuring": true,
    "forOf": true,
    "generators": true,
    "modules": true,
    "objectLiteralComputedProperties": true,
    "objectLiteralDuplicateProperties": true,
    "objectLiteralShorthandMethods": true,
    "objectLiteralShorthandProperties": true,
    "octalLiterals": true,
    "regexUFlag": true,
    "regexYFlag": true,
    "spread": true,
    "superInFunctions": true,
    "templateStrings": true,
    "unicodeCodePointEscapes": true,
    "globalReturn": true,
    "jsx": true
  },

  "rules": {

    //
    //Possible Errors
    //
    // The following rules point out areas where you might have made mistakes.
    //
    "comma-dangle": 2, // disallow or enforce trailing commas
    "no-cond-assign": 2, // disallow assignment in conditional expressions
    "no-console": 1, // disallow use of console (off by default in the node environment)
    "no-constant-condition": 2, // disallow use of constant expressions in conditions
    "no-control-regex": 2, // disallow control characters in regular expressions
    "no-debugger": 2, // disallow use of debugger
    "no-dupe-args": 2, // disallow duplicate arguments in functions
    "no-dupe-keys": 2, // disallow duplicate keys when creating object literals
    "no-duplicate-case": 2, // disallow a duplicate case label.
    "no-empty": 2, // disallow empty statements
    "no-empty-class": 2, // disallow the use of empty character classes in regular expressions
    "no-ex-assign": 2, // disallow assigning to the exception in a catch block
    "no-extra-boolean-cast": 2, // disallow double-negation boolean casts in a boolean context
    "no-extra-parens": 0, // disallow unnecessary parentheses (off by default)
    "no-extra-semi": 2, // disallow unnecessary semicolons
    "no-func-assign": 2, // disallow overwriting functions written as function declarations
[...]

PR #316, opened 2015-04-19
[linters] update .eslintrc. fixes: #313, #259

hshoff:
Finally went through all of the eslint rules.

**Fixes**: 
  - [x] https://github.com/airbnb/javascript/issues/313
  - [x] https://github.com/airbnb/javascript/issues/259

@goatslacker


linters/.eslintrc
```diff
@@ -0,0 +1,168 @@
+{
+  "parser": "babel-eslint",
```
nkbt: With babel-eslint I ran into issue with some stylistic checks. After babel modifies the raw code, it cannot pass some checks. That might be fixed already and worth doublechecking.

linters/.eslintrc
```diff
@@ -0,0 +1,168 @@
...
+    "no-unused-vars": [2, {          // http://eslint.org/docs/rules/no-unused-vars
+      "vars": "local",
+      "args": "after-used"
+    }],
+    "no-use-before-define": 2,       // http://eslint.org/docs/rules/no-use-before-define
+
+/**
+ * Possible errors
+ */
+    "comma-dangle": [2, "never"],    // http://eslint.org/docs/rules/comma-dangle
```
goatslacker: we should change this to always and modify our rules.

linters/.eslintrc
```diff
@@ -0,0 +1,168 @@
...
+    "no-cond-assign": [2,"always"],  // http://eslint.org/docs/rules/no-cond-assign
```
goatslacker: nit, whitespace after ,

linters/.eslintrc
nkbt: This rule is actually correct here. Should be `"never"`

---

The following patterns **are** considered warnings when configured `"never"`:

```javascript
var foo = {
    bar: "baz",
    qux: "quux",
};

var arr = [1,2,];

foo({
  bar: "baz",
  qux: "quux",
});
```
The following patterns **are not** considered warnings when configured `"never"`:

```javascript
var foo = {
    bar: "baz",
    qux: "quux"
};

var arr = [1,2];

foo({
  bar: "baz",
  qux: "quux"
});
```

goatslacker: Yes exactly, we should change it to always and change the rules in the style guide :)

Trailing commas are a good thing!

Hopefully this makes it into ES7: https://github.com/jeffmo/es-trailing-function-commas

nkbt: I would personally disagree with this. But anyway this PR is not the right place to discuss it. Best way is to open a new issue (unless there is/was one already).

hshoff: @goatslacker let's update it for the es6 guide. Should keep the trailing comma rule for es5 since it causes unexpected behavior.

linters/.eslintrc
hshoff: :sparkles: :eyes: :eyes:

linters/.eslintrc
goatslacker: oh definitely, I thought this was meant for ES6 though since it has all the options.

hshoff: @goatslacker good point. Let's do a followup PR once we merge ES6.

PR #318, opened 2015-04-20
Remove excess whitespace in Whitespace section

jensechu:
I believe that two spaces after the commas in these lines are extraneous. I couldn't find anything in the styleguide where this pattern was used, or documented.

:beers:


goatslacker:
Thanks! Good eyes!

PR #322, opened 2015-04-23
Clarifying what rule is being enforced.

kzsh:
Capitalization of constructors is not the rule in question here.  It makes it more clear what the offending difference is between `//bad` and `//good` to have consistent constructor capitalization.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1147,17 +1147,12 @@
     // bad
     var OBJEcttsssss = {};
     var this_is_my_object = {};
+    var o = {};
     function c() {}
-    var u = new user({
-      name: 'Bob Parr'
-    });
 
     // good
     var thisIsMyObject = {};
     function thisIsMyFunction() {}
-    var user = new User({
-      name: 'Bob Parr'
-    });
     ```
 
   - Use PascalCase when naming constructors or classes.


README.md
```diff
@@ -1148,7 +1148,7 @@
     var OBJEcttsssss = {};
     var this_is_my_object = {};
     function c() {}
-    var u = new user({
+    var u = new User({
```
kzsh: u is bad, not the fact that "user" should be "User"

The constructor rule is shown in the next example below (PascalCase)

goatslacker:
We should just kill this example from this section

goatslacker:
@kzsh do you want to amend changes or shall I?

kzsh:
Hey @goatslacker, I'll take care of it now.  Just one thought, though.  It might be worth keeping the `var u` part to be more clear that it's not just somehow applied to functions.  What about `var o = {};` after the other two?  I'll amend it that way, and if you prefer the other way, I'll amend it again.

README.md
```diff
@@ -1147,17 +1147,12 @@
     // bad
     var OBJEcttsssss = {};
     var this_is_my_object = {};
+    var o = {};
```
goatslacker: single letter names is mentioned above

goatslacker:
Can you rebase down to a single commit too?

kzsh:
Ok, let me know if that works out.

goatslacker:
Thanks :beers:

PR #324, opened 2015-04-25
Allow dot-notation keywords in eslint

nkt:
ES6 gives us promises, which have `.catch` method. Other Promise implementation also uses `try` and `finally` methods, which are also reserved keywords ([bluebird](https://github.com/petkaantonov/bluebird/blob/master/API.md)).


hshoff:
Good catch! :+1: 

http://eslint.org/docs/rules/no-reserved-keys will cover the rule in [Objects](https://github.com/airbnb/javascript#objects) about not using reserved words as object keys.

AveVlad:
@nkt красава

PR #331, opened 2015-04-29
React/JSX Style Guide

alvinsng:
Lets share our internal React/JSX style guide!

@hshoff @goatslacker @spikebrehm

### DIFF
--- a/react/README.md
+++ b/react/README.md
@@ -13,18 +13,19 @@
   1. [Props](#props)
   1. [Parentheses](#parentheses)
   1. [Tags](#tags)
+  1. [Methods](#methods)
   1. [Ordering](#ordering)
 
 ## Basic Rules
 
   - Only include one React component per file.
   - Always use JSX syntax.
-  - Do not use `React.createElement` unless you're initializing the app from a `.js` file.
+  - Do not use `React.createElement` unless you're initializing the app from a file that does not transform JSX.
 
 ## Naming
 
-  - **Extensions**: Use `.jsx` extension for React components.
-  - **Filename**: Use PascalCase for filenames. E.g., `ReservationCard.jsx`.
+  - **Extensions**: Use `.js` extension for React components.
+  - **Filename**: Use PascalCase for filenames. E.g., `ReservationCard.js`.
   - **Reference Naming**: Use PascalCase for React components and camelCase for their instances:
     ```javascript
     // bad
@@ -40,13 +41,13 @@
     const reservationItem = <ReservationCard />;
     ```
 
-    **Component Naming**: Use the filename as the component name. So `ReservationCard.jsx` should have a reference name of ReservationCard. However, for root components of a directory, use index.jsx as the filename and use the directory name as the component name:
+    **Component Naming**: Use the filename as the component name. So `ReservationCard.js` should have a reference name of ReservationCard. However, for root components of a directory, use index.js as the filename and use the directory name as the component name:
     ```javascript
     // bad
-    const Footer = require('./Footer/Footer.jsx')
+    const Footer = require('./Footer/Footer.js')
 
     // bad
-    const Footer = require('./Footer/index.jsx')
+    const Footer = require('./Footer/index.js')
 
     // good
     const Footer = require('./Footer')
@@ -71,17 +72,17 @@
     ```
 
 ## Alignment
-  - Follow these alignment styles for JSX syntax
+  - Follow these alignment styles for js syntax
 
     ```javascript
     // bad
-    <Foo bar="bar"
-         baz="baz" />
+    <Foo superLongParam="bar"
+         anotherSuperLongParam="baz" />
 
     // good
     <Foo
-      bar="bar"
-      baz="baz"
+      superLongParam="bar"
+      anotherSuperLongParam="baz"
     />
 
     // if props fit in one line then keep it on the same line
@@ -89,8 +90,8 @@
 
     // children get indented normally
     <Foo
-      bar="bar"
-      baz="baz"
+      superLongParam="bar"
+      anotherSuperLongParam="baz"
     >
       <Spazz />
     </Foo>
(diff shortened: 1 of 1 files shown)


iamnirav:
:grin: :+1:

spikebrehm:
:+1:

react/README.md
```diff
@@ -0,0 +1,218 @@
+# Airbnb React/JSX Style Guide
+
+*A mostly reasonable approach to React and JSX*
```
hshoff: `L` => `l`

alvinsng: ah thanks will do

react/README.md
```diff
@@ -0,0 +1,218 @@
...
+    ```javascript
+    // bad
+    <Foo bar="bar"
+         baz="baz" />
+
+    // good
+    <Foo
+      bar="bar"
+      baz="baz"
+    />
```
hshoff: I'm not sold on the `/>` goes to new line, I know the benefit is the same as the additional trailing comma, but it adds noising lines that making scanning a `render` call harder. So the tradeoff is easier to read vs. nice git diffs.

react/README.md
```diff
@@ -0,0 +1,218 @@
...
+    // bad
+    <Foo
+      bar="bar"
+      baz="baz" />
+
+    // good
+    <Foo
+      bar="bar"
+      baz="baz"
+    />
```
hshoff: I'm not sold on the `/>` goes to new line. I know the benefit is the same as the additional trailing comma, but it adds noising lines that making scanning a `render` call harder. So the tradeoff is easier to read vs. nice git diffs.

react/README.md
```diff
@@ -0,0 +1,218 @@
...
+
+    // if props fit in one line then keep it on the same line
+    <Foo bar="bar" />
+
+    // children get indented normally
+    <Foo
+      bar="bar"
+      baz="baz"
+    >
```
nkbt: Close on the same line. This one looks pretty ugly.

react/README.md
```diff
@@ -0,0 +1,218 @@
...
+## Alignment
+  - Follow these alignment styles for JSX syntax
+
+    ```javascript
+    // bad
+    <Foo bar="bar"
+         baz="baz" />
+
+    // good
+    <Foo
```
nkbt:
```javascript
// better
<Foo bar="bar" baz="baz" />
```

spikebrehm: I suppose what's missing here is our justification: that changes are easier to make with each prop on its own line.

react/README.md
```diff
@@ -0,0 +1,218 @@
...
+
+## Basic Rules
+
+  - Only include one React component per file.
+  - Always use JSX syntax.
+  - Do not use `React.createElement` unless you're initializing the app from a `.js` file.
+
+## Naming
+
+  - **Extensions**: Use `.jsx` extension for React components.
```
nkbt: Can't see why *.jsx extension is necessary. *.js works perfectly. When you remove or add some component into your file, does it mean you have to rename that file? That does not sound cool for me. They are all just JS after all transpilation. 
It is like renaming your files to `.es6.js` or something.

react/README.md
```diff
@@ -0,0 +1,218 @@
...
+    <Foo bar='bar' />
+
+    // good
+    <Foo bar="bar" />
+
+    // bad
+    <Foo style={{ left: "20px" }} />
+
+    // good
+    <Foo style={{ left: '20px' }} />
```
nkbt: Extra spaces `<Foo style={{left: '20px'}} />`

react/README.md
```diff
@@ -0,0 +1,218 @@
...
+    // bad
+    <Foo
+     />
+
+    // good
+    <Foo />
+    ```
+
+## Props
+  - Alphabetically order your props.
```
nkbt: agrguable. If you use jsxstyles for example and add your styling attributes to the tag - it is a mess.
The best order - is **logical** order and not alphabetical.

What is the best logical order - up to developer + code reviewers.

react/README.md
nkbt: yeah, closing angles on new lines look really bad

react/README.md
goatslacker: No the whitespace is intentional

react/README.md
goatslacker: +1 I don't think this is necessary any more since we moved to babel

react/README.md
goatslacker: I used to care about this because if your component has a bunch of props there's no question on where should a new prop go or where to find a prop but it's not that important. I'm neutral on this

react/README.md
alvinsng: This is for when all the props do not fit in a single line. If it fits in a single line then see line 87 below.

react/README.md
alvinsng: they do look bad but we want to maintain git blame, so when another prop is added, it does not need to touch two lines.

iamnirav: I remember having this conversation—my view is that it looks weird now because we're not used to it, but there's nothing inherently wrong with it.

react/README.md
goatslacker: Maybe we should amend the example below to include both props?

react/README.md
```diff
@@ -0,0 +1,241 @@
...
+
+    // good
+    <Foo
+      bar="bar"
+      baz="baz"
+    />
+    ```
+
+## Methods
+  - Do not use underscore prefix for internal methods of a react component
```
hshoff: missing `.`

react/README.md
```diff
@@ -0,0 +1,241 @@
...
+  5. getDefaultProps
+  6. getInitialState
+  7. componentWillMount
+  8. componentDidMount
+  9. componentWillReceiveProps
+  10. shouldComponentUpdate
+  11. componentWillUpdate
+  12. componentWillUnmount
+  13. *clickHandlers or eventHandlers* like onClickSubmit() or onChangeDescription()
+  14. *gettter methods for render* like getSelectReason() or getFooterContent()
```
hshoff: `gettter` => `getter`

react/README.md
hshoff: Not a blocker for merging. I'll follow up in a separate issue.

alvinsng:
Thanks everyone for the review!

Issue #345, opened 2015-05-18
eslint react plugin


nkt:
[There is a great plugin](https://github.com/yannickcr/eslint-plugin-react) for eslint which adds some validation rules for React.
Bellow rules according yours style guide:
```json
"react/display-name": 0,
"react/jsx-boolean-value": 2,
"react/jsx-quotes": [2, "double"],
"react/jsx-no-undef": 2,
"react/jsx-sort-props": 0,
"react/jsx-sort-prop-types": 0,
"react/jsx-uses-react": 2,
"react/jsx-uses-vars": 2,
"react/no-did-mount-set-state": [2, "allow-in-func"],
"react/no-did-update-set-state": 2,
"react/no-multi-comp": 2,
"react/no-unknown-property": 2,
"react/prop-types": 2,
"react/react-in-jsx-scope": 2,
"react/self-closing-comp": 2,
"react/wrap-multilines": 2,
"react/sort-comp": [2, {
  "order": [
    "displayName",
    "mixins",
    "statics",
    "propTypes",
    "getDefaultProps",
    "getInitialState",
    "componentWillMount",
    "componentDidMount",
    "componentWillReceiveProps",
    "shouldComponentUpdate",
    "componentWillUpdate",
    "componentWillUnmount",
    "/^on.+$/",
    "/^get.+$/",
    "/^render.+$/",
    "render"
  ]
}]
```

justjake:
Thank you! I'll open a PR to add this to our ESlint config.

justjake:
Merged these in. Closing.

PR #346, opened 2015-05-18
add eslint rules for JSX style from issue #345

justjake:
Adds React style to our `.eslintrc.` These rules were authored by @nkt

Should we keep these rules separate from our base `.eslintrc` or is it fine to put them in here?

### DIFF
--- a/linters/.eslintrc
+++ b/linters/.eslintrc
@@ -1,5 +1,8 @@
 {
   "parser": "babel-eslint",
+  "plugins": [
+    "react"
+  ],
   "env": {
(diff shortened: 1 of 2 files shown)


hshoff:
Fine to keep them in the existing `.eslintrc` since it is our internal file. Folks can fork or copy + paste and remove the jsx section easily enough.

If it becomes problem to maintain in the future we could introduce a build tool around this, but that's not necessary at the moment. 

:+1: thanks @nkt!

justjake:
PS issue is #345

PR #347, opened 2015-05-18
Amend the arrow function rules

### LINKED ISSUE #281: Parenthesis around single arguments in arrow functions
> Always use parentheses around the arguments. Omitting the parentheses makes the functions less readable and only works for single arguments.

> Why? These declarations read better with parentheses. They are also required when you have multiple parameters so this enforces consistency.
>
>  ```js
>  // bad
>  [1, 2, 3].map(x => x * x);
> 
>  // good
>  [1, 2, 3].map((x) => x * x);
> ```

This stuck out to me as odd. It's an intentional part of the spec to make it more readable for simple cases.

Anyways, I'm interested in discussing this.

goatslacker:
This allows one to omit parens whenever it fits into a single line, which is similar to our if statement rules.

Fixes #281 

cc: @hshoff for the Airbnb change
cc: @nkbt @thejameskyle

### DIFF
--- a/README.md
+++ b/README.md
@@ -590,7 +590,7 @@
     });
     ```
 
-  - [8.2](#8.2) <a name='8.2'></a> If the function body fits on one line and there is only a single argument, feel free to omit the braces and parenthesis, and use the implicit return. Otherwise, add the parenthesis, braces, and use a `return` statement.
+  - [8.2](#8.2) <a name='8.2'></a> If the function body fits on one line and there is only a single argument, feel free to omit the braces and parentheses, and use the implicit return. Otherwise, add the parentheses, braces, and use a `return` statement.
 
   > Why? Syntactic sugar. It reads well when multiple functions are chained together.
 
--- a/README.md
+++ b/README.md
@@ -590,7 +590,7 @@
     });
     ```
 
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -590,7 +590,7 @@
     });
     ```
 
-  - [8.2](#8.2) <a name='8.2'></a> If the function body fits on one line, feel free to omit the braces and use implicit return. Otherwise, add the braces and use a `return` statement.
+  - [8.2](#8.2) <a name='8.2'></a> If the function body fits on one line and there is only a single argument, feel free to omit the braces and parenthesis, and use the implicit return. Otherwise, add the parenthesis, braces, and use a `return` statement.
```
iamnirav: s/parenthesis/parentheses

hshoff:
:+1:

thejameskyle:
:+1:

README.md
```diff
@@ -598,24 +598,12 @@
 
     ```javascript
     // good
```
arianf: There are now two `// good` blocks and no bad blocks? was this the intended result? This is kind of confusing if so, maybe they should just have one `// good` comment.

nkbt:
@goatslacker awesome!

goatslacker:
Thanks for the feedback from both of you!

PR #348, opened 2015-05-19
Store configuration in function default parameters

Jabher:
E.g.

``` javascript
function call(url, timeout = 60) {
  return request.call(url, {timeout});
}
```

I think that really makes sense. 
Moreover, that's a JavaScript way - "everything is exposed, except what's hidden explicitly".

I've been using this technique for a long time and it really makes sense, as it's rather common situation — when you suddenly find out you need to override some closure-stored configuration variables during the call.

### DIFF
--- a/README.md
+++ b/README.md
@@ -567,6 +567,53 @@
   count();  // 3
   ```
 
+  - [7.9](#7.9) <a name='7.9'></a> If you have a configuration values that are used inside of function, use them as default parameters (if this makes sense).
+
+  > Why? That's how you will be able to keep legacy call format for your configuration and be able to modify configuration options during the call.
+
+  ```javascript
+  const requestTimeout = 60;
+
+  // bad
+  function call(url) {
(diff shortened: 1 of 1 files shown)


ljharb:
I'm concerned about this approach, primarily because f you later need to add a required argument, it becomes a breaking change (because somebody might have relied on the timeout value). However, I also am not in favor of exposing anything publicly when it's not strictly necessary.

README.md
```diff
@@ -567,6 +567,53 @@
...
+  const requestBaseUrl = 'http://some.api/';
+
+  // bad
+  function call(url) {
+    return request.call(baseUrl + url, {timeout: requestTimeout});
+  }
+
+  //good
+  function call(url, {timeout = requestTimeout, baseUrl = requestBaseUrl} = {}) {
+    return request.call(baseUrl + url, {timeout: requestTimeout});
```
ljharb: i think this is meant to be `{ timeout }`?

Issue #349, opened 2015-05-22
Add standalone ESLint shareable config


popenkomaksim:
Hello.
In [ESLint 0.21.0 released](http://eslint.org/blog/2015/05/eslint-0.21.0-released/) they introduces such thing as [shareable config](http://eslint.org/docs/developer-guide/shareable-configs.html).
It provide teams import external configs as npm packages. 
What do you think about publishing ESLint config as [standalone npm packadge](https://www.npmjs.com/search?q=eslintconfig), like "eslint-config-airbnb"?

popenkomaksim:
Wow! Incredible fast!

popenkomaksim:
To make config searchable via [standalone npm packadge](https://www.npmjs.com/search?q=eslintconfig), please, extend packadge.json keywords with ```eslintconfig```:
```
  "keywords": [
    "eslint",
    "eslintconfig"
    "config",
    "airbnb",
    "javascript",
    "styleguide"
  ],
```

justjake:
Done: https://github.com/justjake/javascript/commit/c3a94c15fd5385828f47e45b1e329a810d0ecfc3

popenkomaksim:
@justjake you rock!

PR #350, opened 2015-05-22
Create package eslint-config-airbnb

justjake:
This is a sharable ESLint config package. We could optionally just use our eslint config as the package `main` for `airbnb-style` but it'd be nicer for everyone if the eslint package had a cool name.

Solves for issue #349

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,0 +1,1 @@
+node_modules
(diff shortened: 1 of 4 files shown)


justjake:
@popenkomaksim you can try this out now: https://www.npmjs.com/package/eslint-config-airbnb

PR #352, opened 2015-05-25
Change Chinese Translation to a newer version

sivan:
The current Chinese version translated by adamlu is a modified version which as he said change/delete some rules from the original and it's base on the version 2 years ago. I've transalted a new version base on current style guide of es5 version.


hethune:
@hshoff ?

panxianhai:
@sivan is it better to translate a complete version?

1ternal:
I agree with @sivan, it's better not to modify the origin.

lijunle:
I think @sivan 's translation is newer and better.

For the space rule,

The old one still suggests 4 spaces indentation,
The new one suggests 2 spaces indentation.

hshoff:
@sivan thanks for newer/better translation! :beers:

PR #353, opened 2015-05-25
Better solution for multiline strings (es6 feature)

joaosamouco:
Added a better solution for `multiline strings`. This is a `es6` feature.

See: [Template Strings](https://babeljs.io/docs/learn-es6/#template-strings)


joaosamouco:
I just noticed that this isn't probably what we want to get in this case. This will give you a string with brake lines.

goatslacker:
Break lines and white space for the indentation.

PR #357, opened 2015-06-01
added trailing comma

arianf:
In some parts of this style guide, trailing commas are not used for other examples outside the comma section.

According to the [comma section](https://github.com/arianf/javascript#19.2) of this style guide, trailing commas should be used. But outside of that section, it is commonly not used in this style guide.

I think for consistency, when the section/topic of the style guide is not focused on the trailing comma's, they should be used. Especially because it is confusing to see a good example that doesn't include the trailing comma.

### DIFF
--- a/README.md
+++ b/README.md
@@ -146,13 +146,13 @@
     // bad
     const superman = {
       default: { clark: 'kent' },
-      private: true
+      private: true,
     };
 
     // good
     const superman = {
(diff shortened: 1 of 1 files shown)


christophehurpeau:
:+1:

justjake:
Sounds good to me

PR #359, opened 2015-06-08
add missing componentDidUpdate

ycavatars:
A simple pull request that adds missing `componentDidUpdate`.


goatslacker:
Thanks! :sunglasses:

PR #360, opened 2015-06-08
Update jsHint config

tomekwi:
Although `__proto__` can be misused, it’s now [part of the ES 2015 spec](http://people.mozilla.org/~jorendorff/es6-draft.html#sec-additional-properties-of-the-object.prototype-object). JsHint seems to lag behind a little – it still issues a warning even if `esnext` is on.

http://stackoverflow.com/a/29254531/2816199

### DIFF
--- a/linters/jshintrc
+++ b/linters/jshintrc
@@ -15,6 +15,7 @@
 
   // Allow ES6.
   "esnext": true,
+  "proto": true,
 
   /*
    * ENFORCING OPTIONS


ljharb:
It's part of the spec, but it's in "Annex B", which means "deprecated, but standardized since most everybody happens to implement it, and we don't want someone to make it worse".

`__proto__` should never be used - `Object.setPrototypeOf` is the blessed method to do this going forward.

tomekwi:
I actually only ever need `Object.getPrototypeOf`. But I didn’t see anything wrong with setting the prototype within an object literal upon creation, and `__proto__` seems the only way to do it.

Thanks for the note about “Annex B” – I didn’t know there’s such a thing.

I don’t insist on keeping it this way. I just thought it’d be more consistent to support the whole spec.

ljharb:
You're correct that `__proto__` is the only way to set the prototype of an object literal upon creation - otherwise, you'd need to do something like `Object.assign(Object.create(proto), {…})`.

ljharb:
At Airbnb, mutating the `[[Prototype]]` is something we try to avoid, so it's preferred that it not be an easy or trivial thing to do. However, we encourage you to fork the style guide and change it so it fits your needs!

tomekwi:
Thanks for the reply.

> At Airbnb, mutating the [[Prototype]] is something we try to avoid

I try to avoid it too, by all means. But I have nothing against this:

```js
const gorilla = {
  __proto__: monkey,
  big: true,
};
```

– which is not *mutating*.

tomekwi:
Actually, when trying to write JS in a functional style – without *mutating* objects – this is the only way to initialize an object with a prototype attached.

```js
const gorilla = {big: true};
Object.setPrototypeOf(gorilla, monkey);  // Mutation!
```

Besides, the previous example reads better to me. But that’s a personal opinion.

ljharb:
`const gorilla = Object.assign(Object.create(monkey), { big: true });` - there's definitely mutation, but no overriding of the prototype.

tomekwi:
Fair enough, it’ll stay as you wish then.

PR #369, opened 2015-06-23
Update README.md

alsotang:


README.md
```diff
@@ -314,6 +314,7 @@
 
     // good
     const itemsCopy = [...items];
+    const itemsCopy = items.slice();
```
ljharb: Certainly both of them copy the array - why prefer the older API construct over the newer syntactical one?

alsotang: I dont think newer one means better one.

ljharb: `Array.prototype.slice` could be overwritten, or items might not be an `Array` - but as long as items is an iterable, or an array-like object, the spread operator will *always* safely turn it into an `Array`. I think in this case, it's strictly better.

PR #371, opened 2015-06-24
Replace no-wrap-func (deprecated) for no-extra-parens

framini:
no-wrap-func is deprecated and has been superseded by the no-extra-parens rule.


justjake:
this happened, but thanks!

PR #372, opened 2015-06-24
Update .eslintrc with other react class properties

0xCMP:
In `eslint-plugin-react`'s page on their sorting rules they show the order I showed here. https://github.com/yannickcr/eslint-plugin-react/blob/master/docs/rules/sort-comp.md#rule-options

I noticed because according to the current rules `contextTypes` goes below `render` and I also noticed other methods were missing from the list. Since we're being specific we should also include the other methods and properties that `eslint-plugin-react` handles by default.


justjake:
Looks good to me. @goatslacker looks good to you?

goatslacker:
LGTM

PR #373, opened 2015-06-25
Added "prefer-const" rule

nkt:
Added new ES6 rule, introduced in eslint 0.23.0
 > This rule is aimed to flag variables that are declared using let keyword, but never modified after initial assignment.


ljharb:
👍

hshoff:
:+1:

PR #376, opened 2015-06-30
Change information in string sections

dimitardanailov:
I think is better explanation.


justjake:
thanks for checking out the guide, this is the style guide we use internally and we don’t follow that style rule. Feel free to fork the guide and add the rule to keep it consistent among your team

PR #378, opened 2015-07-01
Add a new rule 18.7

dimitardanailov:


ljharb:
If this has an associated `eslint` rule, should https://github.com/airbnb/javascript/blob/master/linters/.eslintrc be updated to include it?

justjake:
Why?

dimitardanailov:
@ljharb I can't find association rule for `eslint`. What do you think for this article / source: [requirePaddingNewlinesBeforeKeywords](http://jscs.info/rule/requirePaddingNewlinesBeforeKeywords.html)

@justjake, because I seen many times developers to be not consecutive.
Many times they will have line break between return and business logic, but again many times they will **follow** this rules. 

I've seen many times this example:
```
function methodA(firstParam, secondParam) {
    // Implementation of my business logic
    return sum; // Here we not have empty line.
}

function methodB(firstParam, secondParam) {
    // Implementation of my business logic

    return sum; // Here we have empty line.
}

function methodC(firstParam, secondParam) {
    // Implementation of my business logic
    return sum; // Here we not have empty line.
}

function methodD(firstParam, secondParam) {
    // Implementation of my business logic

    return sum; // Here we not have empty line.
}
```

hshoff:
Hello again @dimitardanailov! Unfortunately, this isn't a style rule that we enforce at Airbnb. Totally understand the use case you brought up and I bet there are many teams that do follow a style rule around this. One day in the future I could see the team re-opening and adding this, but for now Airbnb isn't ready to add this rule.

This is another one of those times we encourage folks to fork the guide and update it to match their team's style. :beers:

PR #379, opened 2015-07-01
Prevent NaN, when we try to parse string to int

dimitardanailov:
For more information:
1. [javascript parseInt return NaN for empty string](http://stackoverflow.com/questions/6736476/javascript-parseint-return-nan-for-empty-string)
2. [MDN - isNaN()](https://developer.mozilla.org/en/docs/Web/JavaScript/Reference/Global_Objects/isNaN)

### DIFF
--- a/README.md
+++ b/README.md
@@ -1564,10 +1564,12 @@
     const val = parseInt(inputValue);
 
     // good
-    const val = Number(inputValue);
+    // Prevent assign of NaN value
+    const val = Number(inputValue) || 0;
 
     // good
-    const val = parseInt(inputValue, 10);
+    // Prevent assign of NaN value
+    const val = parseInt(inputValue, 10) || 0;
     ```
 
   - [21.4](#21.4) <a name='21.4'></a> If for whatever reason you are doing something wild and `parseInt` is your bottleneck and need to use Bitshift for [performance reasons](http://jsperf.com/coercion-vs-casting/3), leave a comment explaining why and what you're doing.


ljharb:
Unless `NaN` is the value we want here - I see both having `NaN` and defaulting to `0` as real use cases, but I'm not sure why this change belongs in a "avoid typecasting and coercion" section.

dimitardanailov:
1) 
I am big fan of ["Defensive Programming"](https://www.youtube.com/watch?v=_3q4xayI-I4).

`ParseInt` operation always require `try` and `catch` construction, if you follow good patterns from `C#`,  `Java`, `C+`

Example:

### .Net

```csharp
using System;

public class Example
{
   public static void Main()
   {
      string[] values = { "+13230", "-0", "1,390,146", "$190,235,421,127",
                          "0xFA1B", "163042", "-10", "2147483647", 
                          "2147483648", "16e07", "134985.0", "-12034",
                          "-2147483648", "-2147483649" };
      foreach (string value in values)
      {
         try {
            int number = Int32.Parse(value); 
            Console.WriteLine("{0} --> {1}", value, number);
         }
         catch (FormatException) {
            Console.WriteLine("{0}: Bad Format", value);
         }   
         catch (OverflowException) {
            Console.WriteLine("{0}: Overflow", value);   
         }  
      }
   }
}
// The example displays the following output: 
//       +13230 --> 13230 
//       -0 --> 0 
//       1,390,146: Bad Format 
//       $190,235,421,127: Bad Format 
//       0xFA1B: Bad Format 
//       163042 --> 163042 
//       -10 --> -10 
//       2147483647 --> 2147483647 
//       2147483648: Overflow 
//       16e07: Bad Format 
//       134985.0: Bad Format 
//       -12034 --> -12034 
//       -2147483648 --> -2147483648 
//       -2147483649: Overflow     
```

Source: [Int32.Parse Method (String)](https://msdn.microsoft.com/en-us/library/b3h1hf19(v=vs.110).aspx)

### Java

```java
public static int parseInt(String s,
           int radix)
                    throws NumberFormatException
Parses the string argument as a signed integer in the radix specified by the second argument. The characters in the string must all be digits of the specified radix (as determined by whether Character.digit(char, int) returns a nonnegative value), except that the first character may be an ASCII minus sign '-' ('\u002D') to indicate a negative value or an ASCII plus sign '+' ('\u002B') to indicate a positive value. The resulting integer value is returned.
An exception of type NumberFormatException is thrown if any of the following situations occurs:

The first argument is null or is a string of length zero.
The radix is either smaller than Character.MIN_RADIX or larger than Character.MAX_RADIX.
Any character of the string is not a digit of the specified radix, except that the first character may be a minus sign '-' ('\u002D') or plus sign '+' ('\u002B') provided that the string is longer than length 1.
The value represented by the string is not a value of type int.
Examples:

 parseInt("0", 10) returns 0
 parseInt("473", 10) returns 473
 parseInt("+42", 10) returns 42
 parseInt("-0", 10) returns 0
 parseInt("-FF", 16) returns -255
 parseInt("1100110", 2) returns 102
 parseInt("2147483647", 10) returns 2147483647
 parseInt("-2147483648", 10) returns -2147483648
 parseInt("2147483648", 10) throws a NumberFormatException
 parseInt("99", 8) throws a NumberFormatException
 parseInt("Kona", 10) throws a NumberFormatException
 parseInt("Kona", 27) returns 411787
 
Parameters:
s - the String containing the integer representation to be parsed
radix - the radix to be used while parsing s.
Returns:
the integer represented by the string argument in the specified radix.
Throws:
`NumberFormatException` - if the String does not contain a parsable int.
```

Source: [Class Integer](http://docs.oracle.com/javase/7/docs/api/java/lang/Integer.html)

2)

> I'm not sure why this change belongs in a "avoid typecasting and coercion" section.

I not find better place / section.
Do you have any suggestion ?

hshoff:
Hi @dimitardanailov, at Airbnb we prefer to hang on to `NaN`s. We encourage you to fork the guide and update it to match your team's preference for the cases like this.

PR #384, opened 2015-07-02
Extended links to rules and plugins

bettiolo:
Added links to JSX plugin's documentation and some links to `env` documentation and  plugins repositories

### DIFF
--- a/linters/.eslintrc
+++ b/linters/.eslintrc
@@ -1,11 +1,11 @@
 {
-  "parser": "babel-eslint",
+  "parser": "babel-eslint",          // https://github.com/babel/babel-eslint
   "plugins": [
-    "react"
+    "react"                          // https://github.com/yannickcr/eslint-plugin-react
   ],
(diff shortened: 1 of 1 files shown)


hshoff:
:+1:

justjake:
Nice, thanks

ljharb:
O.o Isn't this file supposed to be valid JSON, which can't have comments? Can `eslint` parse this?

hshoff:
@ljharb eslint can still parse it...I was surprised as well.

Once I discoverd eslint works with invalid JSON, I thought it would be more helpful to have an annotated version of the `.eslintrc` in the repo so folks could easily learn more as they go.

justjake:
It isn't necessarily valid JSON. As you can see here, there's a little bit of extra work needed to use it from Node: https://github.com/airbnb/javascript/blob/master/packages/eslint-config-airbnb/index.js#L5

PR #395, opened 2015-07-04
Update to include warning against using eval.

Ryan-McBride:
Added 6.5 to strings section of Readme in order to explicitly warn against using eval(). Listed in ticket #118.

No other lines were edited or removed.

screen shot 2015-07-03 at 7 52 45 pm

PR #396, opened 2015-07-04
Update to warn against Function constructor

Ryan-McBride:
Added section 7.9 to warn against using the Function creator, as it opens similar vulnerabilities as eval.
Originally suggested in pull request #395.

No other lines were edited or removed.

screen shot 2015-07-04 at 10 19 36 am

### DIFF
--- a/README.md
+++ b/README.md
@@ -567,6 +567,14 @@
   count();  // 3
   ```
 
+- [7.9](#7.9) <a name='7.9'></a> Never use the Function constructor to create a new function.
+
+  > Why? Creating a function in this way evaluates a string similarly to eval(), which opens vulnerabilities.
+
+  ```javascript
+  // bad
+  var add = new Function("a", "b", "return a + b");
+  ```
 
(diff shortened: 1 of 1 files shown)


chrisngobanh:
Nitpick: Section 6.1 says to use single quotes instead of double quotes for strings.

Ryan-McBride:
Good call, fixed.

README.md
```diff
@@ -567,6 +567,14 @@
   count();  // 3
   ```
 
+- [7.9](#7.9) <a name='7.9'></a> Never use the Function constructor to create a new function.
+
+  > Why? Creating a function in this way evaluates a string similarly to eval(), which opens vulnerabilities.
+
+  ```javascript
+  // bad
+  var add = new Function('a', 'b', 'return a + b');
```
ljharb: Also `var subtract = Function('a', 'b', 'return a - b');` is bad also.

Ryan-McBride: Added to this PR.

goatslacker:
Thanks!

PR #398, opened 2015-07-04
Fixed typo in 13.4

Ryan-McBride:
A comment in the example of section 13.4 had 'unnecessary' spelled as 'unnessary'

No other lines were edited or removed.

screen shot 2015-07-04 at 10 49 01 am

### DIFF
--- a/README.md
+++ b/README.md
@@ -919,7 +919,7 @@
       return name;
     }
 
-    // bad - unnessary function call
+    // bad - unnecessary function call
     function(hasName) {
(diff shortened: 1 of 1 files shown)


ljharb:
It might be nice to combine all these typo corrections into one single PR, rather than one for each?

Ryan-McBride:
Yeah sorry I was splitting them up out of habit. I finished going through them now so there won't be any additional ones after this. Sorry about that.

PR #400, opened 2015-07-05
Update package.json

SangameshVK:
first one


ljharb:
Updating the version number in a package.json is something that only module maintainers should ever do - it shouldn't be updated in a pull request.

If you'd like to modify the version number, please feel free to fork it and do so!

PR #404, opened 2015-07-06
[strings] 80 char => 100 char

hshoff:
Let's be consistent with our Ruby style guide: https://github.com/airbnb/ruby#line-length

@goatslacker @ljharb @justjake

### DIFF
--- a/README.md
+++ b/README.md
@@ -403,7 +403,7 @@
     const name = 'Capt. Janeway';
     ```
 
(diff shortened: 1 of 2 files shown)


goatslacker:
-1 How about we make the ruby one 80? This

BMorearty:
+1 for 100 char line length.

ljharb:
None of us are on 9" terminals. I'd vote for 120 :-) but 100 is fine.

Complexity, not length, should be the thing we strive to reduce - whether it's in lines, or anything else.

hshoff:
Sorry @goatslacker, I'm going for it. :smile:

PR #407, opened 2015-07-07
use leading plus for string concatenation

num-lock:
i had some code where i occasionally had to add and remove strings, much like the typical work on multi-line arrays. using leading plus made much nicer diffs. same reason in section 19.2 for leading commas.

### DIFF
--- a/README.md
+++ b/README.md
@@ -417,9 +417,9 @@
     fast.';
 
     // good
-    const errorMessage = 'This is a super long error that was thrown because ' +
-      'of Batman. When you stop to think about how Batman had anything to do ' +
-      'with this, you would get nowhere fast.';
+    const errorMessage = 'This is a super long error that was thrown because '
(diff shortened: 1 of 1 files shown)


num-lock:
also note it might be possible to even move the trailing semicolon on its own line. not sure if that would be **too much just for the sake of clean diffs?** i left it out of the commit for now. i prefer it in my own code base at least, but just because i have a lot of multi-line strings that change over development time.

ljharb:
imo this is the same argument as leading vs trailing commas, and should be kept consistent with that rule.

These days, a multiline string should be done with an interpoliteral/template string literal anyways, not with quoted string concatenation.

hshoff:
+1 to @ljharb, use template strings if you can and most importantly maintain consistency with location of `+`.  Definitely fork the guide and update it for you're own teams preference on places that differ from Airbnb. We encourage that!

PR #415, opened 2015-07-11
Move babel-eslint to peer dependencies

nkbt:
The reason for this is that config is something that is mixed in into a main config. And main project not necessarily has babel-eslint in there. Internal dependency of eslint-config-airbnb is ignored (it is not a module, rules/keys are just mixed in by eslint), so when you run lint there is an error that "babel-eslint is not found".

### DIFF
--- a/packages/eslint-config-airbnb/package.json
+++ b/packages/eslint-config-airbnb/package.json
@@ -24,9 +24,11 @@
     "url": "https://github.com/airbnb/javascript/issues"
   },
   "homepage": "https://github.com/airbnb/javascript",
+  "peerDependencies": {
+    "babel-eslint": "3.x"
+  },
   "dependencies": {
     "airbnb-style": "2.0.0",
-    "babel-eslint": "3.1.7",
     "eslint": "0.21.2",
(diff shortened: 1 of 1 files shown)


packages/eslint-config-airbnb/package.json
```diff
@@ -24,9 +24,11 @@
     "url": "https://github.com/airbnb/javascript/issues"
   },
   "homepage": "https://github.com/airbnb/javascript",
+  "peerDependencies": {
+    "babel-eslint": "3.x"
```
nkbt: Using peer dependencies locking down dep version will not play well

iamstarkov:
peerDeps have been deprecated, so prob it will not work

nkbt:
@iamstarkov, not really. They are not deprecated. In 3.0 they just will not be automatically installed so developer can decide which version to install manually (if needed). Do I miss something here?

justjake:
babel-eslint is now on 4.x. Perhaps we can just leave off a version

justjake:
dependencies removed

PR #418, opened 2015-07-14
Proper ESLint React components order validation

kasperekt:
It fixes validation of React methods order.

For example, when I had:

```js
const Input = React.createClass({
  componentDidMount() {
    console.log('mounted');
  },

  getInitialState() {
    return {
      a: 1,
    };
  },
});
```

Linter showed no errors. That happened because of `/^get.+$/` regexp. It matched with getInitalState so it was allowed to place `getInitialState` below `componentDidMount`.

### DIFF
--- a/linters/.eslintrc
+++ b/linters/.eslintrc
@@ -208,7 +208,7 @@
         "componentDidUpdate",
         "componentWillUnmount",
         "/^on.+$/",
-        "/^get.+$/",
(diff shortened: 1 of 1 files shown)


hshoff:
Looks like this was fixed by @justjake here: https://github.com/airbnb/javascript/pull/479#discussion_r37468261

Thanks for the heads up @kasperekt!

PR #420, opened 2015-07-15
add constructor as an alias for componentWillMount

iamstarkov:
As far as we can use ES6+ syntax, we should add it to the styleguide.
See details here http://babeljs.io/blog/2015/06/07/react-on-es6-plus/

what do you think?

### DIFF
--- a/linters/.eslintrc
+++ b/linters/.eslintrc
@@ -197,6 +197,7 @@
         "mixins",
         "statics",
         "defaultProps",
+        "constructor",
         "getDefaultProps",
         "getInitialState",
         "getChildContext",


justjake:
I think with all ES6 classes constructor should come first.

iamstarkov:
I used to think so, but babel blog post states _"constructor now assumes the role previously filled by componentWillMount"_:

> All of the lifecycle methods but one can be defined as you would expect when using the new class syntax. The class' constructor now assumes the role previously filled by componentWillMount:
> 
> ```js
> // The ES5 way
> var EmbedModal = React.createClass({
>   componentWillMount: function() { … },
> });
>
> // The ES6+ way
> class EmbedModal extends React.Component {
>   constructor(props) {
>     super(props);
>     // Operations usually carried out in componentWillMount go here
>   }
> }
> ```

justjake:
Right, I understand that in the new React style, `constructor` has similar semantics to the old `componentWillMount`. However, I think it's better to consistently put the constructor method first in all ES6 classes, whether or not they are React components. Furthermore, since `constructor` also replaces `getInitialState` and `getDefaultProps` (which are both higher in the list) I don't think it's exactly equivalent to `componentWillMount`.

iamstarkov:
move constructor above `getInitialState` and `getDefaultProps`, but below `defaultProps` as far as they need to be defined as static properties

iamstarkov:
from my perspective now it looks good

justjake:
Looks good to me.

justjake:
sorry for the wait!

PR #421, opened 2015-07-15
Rename jshintrc to .jshintrc

chrisngobanh:
According to http://jshint.com/docs/, the "special file" to configure JSHint is named `.jshintrc`, not `jshintrc`.


chrisngobanh:
#134 is where the `jshintrc` file was added. The description of the PR says that "This adds a `.jshintrc` to the `linters/` directory", when that wasn't actually the case. This pull request corrects this.

justjake:
I'm split. On one hand, that is the proper name of the file for a project. On the other, we aren't actually linting any code in our linters directory, and users are expected to copy the file into their project anyways. Dotfiles are harder to work with: by default my editor hides them in the sidebar, and I can't tab complete them on the terminal. I would prefer to keep this as `jshintrc` and rely on users copying it the correct filename in their projects.

chrisngobanh:
@justjake I didn't realize that this was intentional. Seeing as how the other file in the directory is a dotfile, and the pull request I referenced says that it adds a `.jshintrc` file, I assumed that this was a mistake.

justjake:
it was probably a mistake back then -- i was expressing my personal opinion

ljharb:
I'd love to hear some more opinions on this - personally, I don't find dotfiles difficult to work with, and I already have to configure my editor to show dotfiles to work with `.eslintrc`, `.gitignore`, and many others. I'm a weak +1 in favor of this.

goatslacker:
+1 it's better to copy over the file without having to rename it.

Plus someone could forget to rename it and then it won't work for them.

chrisngobanh:
I think it should be changed to a dotfile because:

1. It needs to be a dotfile to work.
2. The other file in the directory is a dotfile. If we decide to not have it as a dotfile, then `.eslintrc` should be renamed.

ljharb:
@chrisngobanh that's a strong argument too. (would you mind freshly rebasing this on latest master, btw?)

chrisngobanh:
@ljharb Done!

Side question: Is the `jshintrc` file still being maintained? It looks out of date

ljharb:
Somehow I doubt it. We may end up removing the file anyways - jshint's ES6 support is limited.

PR #424, opened 2015-07-16
Add Waysact exception for single quote -> double quote

mattdaw:
Based on discussion in new-backend-dev, we prefer sticking to one style of quoting strings... rather than single quotes in JS and double quotes in JSX as suggested by this style guide.


mattdaw:
Guh, sorry, that was meant to be in a forked repo. Closing.

PR #425, opened 2015-07-16
Update React style guide with new rules

goatslacker:
This formalizes the rules that we've been following for class extends React.Component and also updates our ordering with whats in the eslintrc file.

### DIFF
--- a/react/README.md
+++ b/react/README.md
@@ -20,7 +20,27 @@
 
   - Only include one React component per file.
   - Always use JSX syntax.
-  - Do not use `React.createElement` unless you're initializing the app from a file that does not transform JSX.
+  - Do not use `React.createElement` unless you're initializing the app from a file that is not JSX.
+
+## Class vs React.createClass
+
+  - Use class extends React.Component unless you have a very good reason to use mixins.
+
+  ```javascript
+  // bad
+  const Listing = React.createClass({
+    render() {
+      return <div />;
+    }
+  });
+  
+  // good
+  class Listing extends React.Component {
+    render() {
+      return <div />;
+    }
+  }
+  ```
 
 ## Naming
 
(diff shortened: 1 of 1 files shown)


goatslacker:
Also related: #420

react/README.md
```diff
@@ -219,24 +239,77 @@
...
+    id: PropTypes.number.isRequired,
+    url: PropTypes.string.isRequired,
+    text: PropTypes.string,
+  };
+  
+  const defaultProps = {
+    text: 'Hello World',
+  };
+  
+  class Link extends Component {
```
spikebrehm: Alternative: I kind of like doing

```js
export default class Link extends Component {
  ...
}

Link.propTypes = propTypes;
```

thoughts?

spikebrehm:
Nice.

react/README.md
goatslacker: No opinion on the matter, should we enforce one way or another? It seems like this belongs in the module discussion on the main style guide.

ljharb: I like things that encourage people to have a single default export

iamnirav: seconded

justjake:
Where are the changes to the .eslint for this? Someone submitted a PR to put `constructor` in the eslint file here: 

I don't want to abandon their PR in favor of this one without the .eslint change from #420 or similar

goatslacker:
No changes to eslintrc in this PR, #420 takes care of adding constructor. The rest of the order is already present in our eslintrc.

justjake:
ok. LGTM

react/README.md
goatslacker: Looks like we already have that: https://github.com/airbnb/javascript#10.2

PR #431, opened 2015-07-20
Update ESlint deprecated rules

pascalduez:
- no-wrap-func -> no-extra-parens
- spaced-line-comment -> spaced-comment


justjake:
if there are no immediate objections I will merge this.

PR #432, opened 2015-07-21
Update Standard ECMA-262 and Add You Don't Know JS: ES6 book

lvarayut:
- In the Resources section, the Annotated ECMAScript 5.1 was replaced by
  Standard ECMA-262, which is a ECMAScript® 2015 Language Specification
- Add You Don't Know JS: ES6 & Beyond book by Kyle Simpson

### DIFF
--- a/README.md
+++ b/README.md
@@ -1984,7 +1984,7 @@
 
 **Read This**
 
-  - [Annotated ECMAScript 5.1](http://es5.github.com/)
+  - [Standard ECMA-262](http://www.ecma-international.org/ecma-262/6.0/index.html)
 
 **Tools**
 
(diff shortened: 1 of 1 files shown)


ljharb:
These are three separate changes - could you do them in three separate PRs?

README.md
```diff
@@ -982,7 +982,7 @@
     // declaration to the top of the scope,
     // which means our example could be rewritten as:
     function example() {
-      let declaredButNotAssigned;
+      var declaredButNotAssigned;
```
ljharb: Using `let` here, which is not hoisted, is the intention, since it's meant to represent what happens when `var` is hoisted on line 978.

lvarayut:
Sure, I'm doing it. :)

README.md
lvarayut: Correct me if I'm wrong. Based on the comment on line 981 -  `The interpreter is hoisting the variable declaration to the top of the scope, which means our example could be rewritten as`. It seems that the code from the example above would be rewritten in the perspective of interpreter, which shows the `var` will be hoisted to the top of the block.

I think that `let` isn't hoisted is showing on the line 992.

ljharb: right - but since `let declaredButNotAssigned` is now *at the top of the block*, it matches the semantic of the var in the previous snippet.

lvarayut: Alright, it's just a little bit strange for me, however, I will change it back to `let`.

lvarayut:
@ljharb Are commits looking good for you? If something should be fixed, don't hesitate to tell me.

ljharb:
These look good to me, thanks

PR #442, opened 2015-07-29
Get rid of `no-reserved-keys` in the .eslintrc

tomekwi:
After all, [rule 8.2](https://github.com/airbnb/javascript/tree/63bece1#3.2) allows them in an ES 2015+ environment.


hshoff:
Yup! Good catch.

Issue #438, opened 2015-07-28
Implicit return – one line vs. one expression


tomekwi:
Hello again. I have a subjective suggestion.

Point [8.2](https://github.com/airbnb/javascript/tree/63bece1234b446ec0f1354ca5d8b62e18914383c#8.2) says: *“If the function body fits on one line and there is only a single argument, feel free to omit the braces and parentheses, and use the implicit return. Otherwise, add the parentheses, braces, and use a return statement.”*. To be honest, this is the only rule I don’t keep to in my projects.

Simple function expressions – a set of arguments mapping to a single expression – are at the very heart of functional languages like JavaScript.

Consider this example:

```js
[/* … */]
  .map(number => `A string containing the ${number}.`)
  /* … */

// As time goes by, the code evolves…
[/* … */]
  .map((number) => (
    `The string containing the ${number} now became much longer. A common ` +
    'thing to do when a line of code outgrows its 80 character limit is to ' +
    'break it over multiple lines.'
  ))
  /* … */

// That looks more natural to the language than:
[/* … */]
  .map((number) => {
    return (
      `The string containing the ${number} now became much longer. A common ` +
      'thing to do when a line of code outgrows its 80 character limit is to ' +
      'break it over multiple lines.'
    );
  ))
  /* … */
```

I suggest changing that guideline to the following:

> If the function body ~~fits on one line and there is only a single argument~~ **consists of a single expression**, feel free to omit the braces ~~and parentheses,~~ and use the implicit return. **When your expression spans over multiple lines, wrap it in parentheses for better readability.** Otherwise, add the ~~parentheses,~~ braces and use a return statement.

ljharb:
Theoretically though, that would be a template string, which means the expression isn't broken up by the newlines within it, right?

tomekwi:
That’s just string concatenation there in the example.

But neither do multiline template strings keep to the “fits on one line” rule.

goatslacker:
This does fit well with React where you do something like:

```js
foo.map(Parent => (
  <Parent>
    <Child />
  </Parent>
))
```

justjake:
I think this is a good change. Submit a PR!

PR #443, opened 2015-07-29
Replace deprecated ESLint rules

tomekwi:
http://eslint.org/docs/rules/no-wrap-func
http://eslint.org/docs/rules/spaced-comment

Both have been deprecated and will be removed from ESLint soon.

### DIFF
--- a/linters/.eslintrc
+++ b/linters/.eslintrc
@@ -166,7 +166,7 @@
     "space-before-function-paren": [2, "never"], // http://eslint.org/docs/rules/space-before-function-paren
     "space-infix-ops": 2,            // http://eslint.org/docs/rules/space-infix-ops
     "space-return-throw-case": 2,    // http://eslint.org/docs/rules/space-return-throw-case
-    "spaced-line-comment": 2,        // http://eslint.org/docs/rules/spaced-line-comment
+    "spaced-comment": 2,             // http://eslint.org/docs/rules/spaced-comment
 
 /**
  * JSX style
--- a/linters/.eslintrc
+++ b/linters/.eslintrc
@@ -152,7 +152,7 @@
     "no-new-object": 2,              // http://eslint.org/docs/rules/no-new-object
     "no-spaced-func": 2,             // http://eslint.org/docs/rules/no-spaced-func
     "no-trailing-spaces": 2,         // http://eslint.org/docs/rules/no-trailing-spaces
-    "no-wrap-func": 2,               // http://eslint.org/docs/rules/no-wrap-func
+    "no-extra-parens": 2,            // http://eslint.org/docs/rules/no-extra-parens
     "no-underscore-dangle": 0,       // http://eslint.org/docs/rules/no-underscore-dangle
     "one-var": [2, "never"],         // http://eslint.org/docs/rules/one-var
     "padded-blocks": [2, "never"],   // http://eslint.org/docs/rules/padded-blocks


linters/.eslintrc
```diff
@@ -152,7 +152,7 @@
     "no-new-object": 2,              // http://eslint.org/docs/rules/no-new-object
     "no-spaced-func": 2,             // http://eslint.org/docs/rules/no-spaced-func
     "no-trailing-spaces": 2,         // http://eslint.org/docs/rules/no-trailing-spaces
-    "no-wrap-func": 2,               // http://eslint.org/docs/rules/no-wrap-func
+    "no-extra-parens": 2,            // http://eslint.org/docs/rules/no-extra-parens
```
ljharb: If we only want the `no-extra-parens` rule to apply to functions, we'll need the "functions" option.

linters/.eslintrc
```diff
@@ -166,7 +166,7 @@
     "space-before-function-paren": [2, "never"], // http://eslint.org/docs/rules/space-before-function-paren
     "space-infix-ops": 2,            // http://eslint.org/docs/rules/space-infix-ops
     "space-return-throw-case": 2,    // http://eslint.org/docs/rules/space-return-throw-case
-    "spaced-line-comment": 2,        // http://eslint.org/docs/rules/spaced-line-comment
+    "spaced-comment": 2,             // http://eslint.org/docs/rules/spaced-comment
```
ljharb: We should be careful with this one; it's currently unreliable with multiline comments that start with `/*!` or `/**` or similar.

tomekwi: According to <http://eslint.org/docs/rules/spaced-comment>:
* `/**` is OK.
* `/*!` seems to require `[2, {"markers": "!"}]`

linters/.eslintrc
tomekwi: But do we? I guess all of the filtered-out cases are against good practices: <http://eslint.org/docs/rules/no-extra-parens#options>. Don’t you think so?

linters/.eslintrc
ljharb: We should just experiment with it - i've found "markers" to not work as advertised.

linters/.eslintrc
ljharb: I strongly prefer using "extra" parens with boolean logic - ie, `a && (b || c)`, solely for clarity. This rule, without configuration, warns on this.

tomekwi: Me too. I agree.

Besides, it goes against my suggestion in #438:

```sh
$ cat <<——— > f.js
export default () => (
  'aaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa' +
  'aaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa'
);
———

$ eslint --rule '{"no-extra-parens": 2}' f.js 

f.js
  1:21  error  Gratuitous parentheses around expression  no-extra-parens

✖ 1 problem (1 error, 0 warnings)

$ eslint --rule '{"no-extra-parens": [2, "functions"]}' f.js
# ✓ Exits with 0
```

linters/.eslintrc
tomekwi: It seems to work OK as of `eslint@0.24.1`:

```sh
$ cat <<——— > f.js
/**
 * Should be OK
 */
/*!
 * Should be OK too
 */
———

$ npm install --global eslint@0.24.1

$ eslint --rule '{"spaced-comment": 2}' f.js

f.js
  4:0  error  Expected space or tab after /* in comment  spaced-comment

✖ 1 problem (1 error, 0 warnings)


$ eslint --rule '{"spaced-comment": [2, "always", {"markers": ["!"]}]}' f.js
# ✓ Exits with 0
```

tomekwi:
@ljharb thanks for your guidance. I’ve investigated the thing a bit and found out which rules should be OK. Let me know if that satisfies you guys and I’ll update the PR.

tomekwi:
Looks like it’s a duplicate of #431 by the way. Sorry @pascalduez! You have right of way. Feel free to move the discussion to your PR.

justjake:
Merged a prior dupe

PR #446, opened 2015-07-30
change second 18.5 to 18.6

tschuy:
There are two sections 18.5. This changes the second one to 18.6.


tschuy:
I don't know if you all would prefer to simply put the second 18.5 below the current 18.6 as 18.7, or shift them both down one number. I did the latter to keep the order the same.

PR #447, opened 2015-07-30
Support `restParams` in ESLint

tomekwi:
Otherwise we get this error ( https://github.com/eslint/eslint/issues/2346 ) when using the default parser on code like `(...args) => result`. Code which [couldn’t be more valid](https://github.com/airbnb/javascript/tree/0c29596a7eaf8f03c8a85428014075041098db0f#7.6).

### DIFF
--- a/linters/.eslintrc
+++ b/linters/.eslintrc
@@ -20,6 +20,7 @@
     "objectLiteralDuplicateProperties": false,
     "objectLiteralShorthandMethods": true,
     "objectLiteralShorthandProperties": true,
+    "restParams": true,
     "spread": true,
     "superInFunctions": true,
     "templateStrings": true,


hzoo:
Maybe just add

```yaml
env:
  es6: true
```

to cover all es6 features?

tomekwi:
:+1: Good idea!

I’m closing this PR then – I’ll try to put together another one once I catch up with my stuff and find a spare while.

justjake:
fyi, we have some ES6 features turned off (such as generators) because we don't want engineers at Airbnb using them.

tomekwi:
@justjake then maybe we should reopen this PR?

hzoo:
Didn't know how not using generators so :+1:

Issue #263, opened 2015-03-28
Branding


tomekwi:
What I like about [JS standard](https://github.com/feross/standard) is not only its simplicity (no config), but also branding. But Standard is just a bunch of opinions – it's far from this guide in terms of comprehensiveness.

But yes, I'm a visual thinker – and I do buy stuff by their looks :) But I guess so do many other developers and designers. Just as a good readme with a green travis badge looks reassuring, so does a well-known logo.

How about developing something similar to this:

[![js-standard-style](https://raw.githubusercontent.com/feross/standard/master/badge.png)](https://github.com/feross/standard)

tomekwi:
It could have the AirBnB logo and a simple slogan like

### “coded in JavaScript style”

You've got great designers out there, telling by https://www.airbnb.com :)

nkbt:
I do believe "JS Standard Code Style" should be used here instead of the other one which actually is "JS Opinionated Code Style"

hshoff:
Interesting. I wonder if it'd be nice to make a general style badge  that's extensible in a way so it's easy to tell based on the readme what style the repo follows whether it's Airbnb, Crockford, jQuery, idiomatic or Standard. I want to think about this some more and will try to follow up later this week.

tomekwi:
@hshoff, one option is to use <http://shields.io/>:

![](https://img.shields.io/badge/code%20style-airbnb-brightgreen.svg?style=flat-square)

@nkbt how about naming it

## “Quality JS code style”

IMO “airbnb” isn't a name which conveys a meaning. Using “JS Standard” code style feels cool – like it's the right thing to do. But using “google” code style, “airbnb” code style, “facebook” code style – or whatever – feels dull. There's no feeling behind it, no subconcious association.

 

 

Think about it:

![](https://img.shields.io/badge/code%20style-Quality%20JS-brightgreen.svg?style=flat-square)

– doesn't that feel reassuring?

tomekwi:
By the way, the NPM package *quality* is free :)

Whether we just want to put the readme there, or a zero-conf tool like <http://npm.im/standard>.

goatslacker:
![code style - Airbnb](https://img.shields.io/badge/code%20style-Airbnb-ff5a5f.svg)

nkbt:
@tomekwi, agree, using Airbnb in a shield tells almost nothing how "much standard" it is. And it apparently is the most standard code style according to different code style surveys.

[...]

PR #455, opened 2015-08-05
Mention eslint-config-airbnb

tomekwi:
Actually, if we want to make the thing more convenient, we could go as far as saying:

    Before using our `.eslintrc` install these dependencies:

    ```sh
    npm install --save-dev eslint-config-airbnb eslint babel-eslint eslint-plugin-react
    ```

I want to leave the decision up to you though.

### DIFF
--- a/linters/README.md
+++ b/linters/README.md
@@ -2,6 +2,7 @@
 
 Our `.eslintrc` requires the following NPM packages:
 
+- `eslint-config-airbnb`
 - `eslint`
 - `babel-eslint`
 - `eslint-plugin-react`


chicoxyzzy:
I agree about adding instruction on how to install, use, and update style guide using npm. But adding `eslint-config-airbnb` as a dependancy of `.eslintrc` looks wrong (or it should be added somewhere that .eslint should contain  `{ "extends": "airbnb" }`)

hshoff:
Love that this PR opened up. The Airbnb team was just discussing this yesterday. We're in the process of making the `Getting Started` process much easier.

We started by talking about the name/branding of the package (related to https://github.com/airbnb/javascript/issues/263) and @ljharb brought up the wonderful idea of using scoped npm packages. So the one-liner command would become:

```bash
npm install @airbnb/javascript --save-dev
```

Hopefully scoped packages catches on so if your team (Super Cool Co.) forks the guide and publishes your own package it would be:

```bash
npm install @supercoolco/javascript --save-dev
```

Which I think is a nice benefit of the scoped package route and encourages teams to use our guide as a starting point for their team's style guide.

`@airbnb/javascript` will be the one-step install package. Current plan is to have this package be a fork of https://github.com/feross/standard which will change the defaults to `@airbnb/eslint-config`. `standard` already has really great tooling for getting up and going.

Of course for those that don't want the batteries included approach with `@airbnb/javascript` and `eslint`, we'll offer and maintain:

- `@airbnb/jshint-config`
- `@airbnb/eslint-config`
- [jscs preset?](https://github.com/jscs-dev/node-jscs/blob/master/presets/airbnb.json) need to talk with the jscs team and see if they prefer maintaining the preset or if we can move it to `@airbnb/jscs-config`.

To @chicoxyzzy's point, once we have the above in place we'll add a getting started/how to use section to the top of the guide.

How does this sound to everyone? We're still very much in the proposal stage of all this.

tomekwi:
Wow! Great news! :smiley: 

I spent quite a lot of time updating my toolset last week, then merging any changes to your `.eslintrc` – which were actually [quite frequent](https://github.com/airbnb/javascript/commits/master/linters/.eslintrc) at that time. It’d be lovely to have all that in one tool, and have the power of npm and clarity of semver at hand!

chicoxyzzy:
@tomekwi you can already use semver by doing `npm i eslint-config-airbnb --save-dev` and use https://github.com/airbnb/javascript/blob/master/linters/.eslintrc with `npm shrinkwrap --dev` to be sure your project will work after deployment.

@hshoff sounds very reasonable!

tomekwi:
> you can already use semver

Yeah, I know. With ESLint config, since Friday last week. Before that I had to update by hand.

At the moment JSCS config follows its own independent versioning. Not to mention JSHint – I don’t know of a versionable way to configure it.

hzoo:
@tomekwi @hshoff You can define your own presets in JSCS now - I'm going to ask if we can move the airbnb preset to here instead (so it's easier to maintain/version). So there could be a `jscs-preset-airbnb` npm package or something. I'd rather move it here although it seems like there needs to be seperate es5/es6 presets now?

tomekwi:
@hzoo @hshoff sounds great!

ariporad:
@hshoff: The standard bit really isn't that hard. They have this thing called standard-engine, it's designed for exactly this. Basically, you copy and paste three files in, change one line from 'standard' to 'airbnb', then create a README and publish. I'll happily do it. I still think (very strongly), that `mostly-reasonable` would be a better name, (or better yet, `mostly-harmless`), but it's yours to do what you want with it.

tomekwi:
@ariporad :+1: :+1:

tomekwi:
Oh @ariporad I’ve just read https://github.com/airbnb/javascript/issues/263#issuecomment-146388087 – it looks like the discussion is now closed.

hzoo:
@hshoff I made https://github.com/jscs-dev/node-jscs/issues/1811 to track moving org presets to their own repos. So like you said making a new repo like `@airbnb/jscs-config.` Let us now if you need help with this

PR #458, opened 2015-08-06
use 120 max-len

levithomason:
# Brief
The max-len rule of 80 columns is causing us to write less performant and harder to read code for an arbitrary goal.  This PR shows our code will be faster, easier to read, and easier to write by adopting a more modern wider max column length.

## Why 80 columns?
First, where did this standard come from and why was it used?

#### Punchcards
Yes, columns of 80 characters were instituted for IBM [punch cards](http://programmers.stackexchange.com/questions/148677/why-is-80-characters-the-standard-limit-for-code-width).  Somewhat arbitrarily as because they were wanted to reuse currency carriers [from 1890](http://programmers.stackexchange.com/questions/148677/why-is-80-characters-the-standard-limit-for-code-width/148678#comment-281684).

#### Old Tech
Typewriters, early text editors, and early terminals also used this *traditional* width as seen [here](https://en.wikipedia.org/wiki/Characters_per_line).  All modern text editors and terminals now support variable column widths (and softwrap if you prefer).  Even so, the StackOverflow 2015 survey [shows ~82%](http://stackoverflow.com/research/developer-survey-2015#tech-editor) of developers do not use a terminal as their text editor.
>Side note, JS is the [most popular](http://stackoverflow.com/research/developer-survey-2015#tech-lang) tech :D Woooo!

#### Readability
It is widely regarded that columns ~40-66-ish are most readable.  So, "narrower columns are more readable than wider columns". Science has shown this to be false.  On screen columns narrower than 95cpl actually decrease reading [speed and comprehension](http://psychology.wichita.edu/surl/usabilitynews/72/pdf/Usability%20News%2072%20-%20Shaikh.pdf).  Summary [here](http://viget.com/inspire/the-line-length-misconception).

## Why *not* 80 columns
Here are some pragmatic reasons why we should *not* be using this antiquated arbitrary standard.

#### Variables & Closures
In order to fit code into narrower columns, variables that would *not otherwise be initialized* are often initialized just to break down column count.  This is unjustified as it is not for readability, performance, or organization.

Declaring and initializing variables costs cycles and memory.  More importantly, it encourages [closure abuse](https://developers.google.com/speed/articles/optimizing-javascript#avoiding-pitfalls-with-closures) because the developer is likely to move code in a nested location up to an outer closure to gain the extra indentation, allowing it to fit on a single line.

#### Concatenating strings
Strings are often forced to be concatenated when arbitrarily limiting column width.  This causes extra code, extra operations, and decreased readability.  Concatenating strings is a known [performance issue](https://developers.google.com/speed/articles/optimizing-javascript#working-with-strings).  Using a `[].join('')` is faster but also unnecessarily slow and verbose when the entire issue can be circumvented by allowing a string to be defined in a single line.

#### Problem Solving Energy
The developer is no longer just thinking "how do I write elegant code?" but also now adds the constraint "how I do refactor my elegant code to fit inside 80 columns?"  These are extra brain cycles that could be put toward solving application engineering problems instead of arbitrarily fitting their code into 80 columns.

#### Code Reasoning
Breaking code down into such small segments can turn an easily turn a readable 2/3 LOC into a very contrived set of variables and string concatenations that is difficult to parse.  In my experience, this outcome is more common than an improved outcome when limiting lines to 80 columns.

It is also harder to retain your train of thought about the system when you have to stop and reason about the superfluous variables and concatenation for an individual method before moving on.

#### Comprehension
The developer regularly has to ask "what words can I cut out of this to make it fit".  The result is often a hasty move to hack and slash a less helpful error message, a more cryptic method name, further abbreviated variables, etc.

#### Diversion
Really stop and think about how many times a day you fiddle with a line, 2, 3, 4 times trying various refactors because you are 1-4 characters over the limit.  These changes were not prompted to improve user experience, developer efficiency, nor system performance but to meet an arbitrary goal with no tangible benefits.  Seems tiny, but this stuff adds up.  Like how recaptcha accidentally wasted a [>a quarter billion](http://www.cnet.com/news/recaptcha-reusing-your-wasted-time-online/) hours of human energy a year.

## Why 120 columns?
Breaking things down can be good.  Breaking them down too much can be bad.  The PR suggests 120 columns (another arbitrary standard) still requires the developer to break down code, but with far less negative impact to code performance, developer performance, and readability.

# Conclusion
The 'standard' 80 column width is arbitrary and antiquated.  There is no measurable gain to the user, the developer, nor the system by limiting columns.  Further, it is shown above that this limit actually decrease code performance, developer performance, and readability.  A wider arbitrary limit reduces these negative impacts while maintaining the sometimes helpful rule to break code down.

:facepunch:

### DIFF
--- a/linters/.eslintrc
+++ b/linters/.eslintrc
@@ -121,6 +121,9 @@
 /**
  * Style
  */
+    "max-len": [2, 120, 2, {         // http://eslint.org/docs/rules/max-len
+      ignoreUrls: true
+    }],
     "indent": [2, 2],                // http://eslint.org/docs/rules/
     "brace-style": [2,               // http://eslint.org/docs/rules/brace-style
       "1tbs", {
--- a/react/README.md
+++ b/react/README.md
@@ -1,4 +1,4 @@
-# Airbnb React/JSX Style Guide
+# TechnologyAdvice React/JSX Style Guide
 
 *A mostly reasonable approach to React and JSX*
 


levithomason:
wrong repo :P

hshoff:
:clap: :clap: Great write-up!

levithomason:
heh, thanks.  religious battles ya know...

PR #463, opened 2015-08-08
Simplify ESLint config package.

nkt:
This patch moves `.eslintrc` contents into `index.js` directly.
This allows us to drop external dependencies and using javascript
comments instead of not valid JSON.


nkbt:
:+1:

justjake:
LGTM

justjake:
i was mostly worried about preserving the file location inside our package for whatever reason

nkt:
@justjake I've created my own eslint config preset. I split rules into different files, [`/rules/possible-errors.js`](https://github.com/nkt/eslint-config-nkt/blob/master/src/rules/possible-errors.js) for example.
What do you think about this approach?

justjake:
i really like that approach. Maybe we should start by importing the work Walmart Labs has done for us? They've got a thing here: https://github.com/walmartlabs/eslint-config-defaults/blob/master/rules/eslint/best-practices/airbnb.js, https://github.com/walmartlabs/eslint-config-defaults/blob/master/rules with a bunch of our rules split up by category already

nkbt:
Perfect.

PR #457, opened 2015-08-06
Add missing `state`

nthtran:
There's `defaultProps` but there's no `state` for React.

PR #464, opened 2015-08-08
Add state to sort-comp rule before render

lauritzsh:
When using class properties you can just write

```javascript
class App extends Component {
  state = {counter: 0}

  render() { /* ... */ }
}
```

`state` should come before `render`


lauritzsh:
Didn't see https://github.com/airbnb/javascript/pull/457 so closed it.

Issue #466, opened 2015-08-10
How should I name my jQuery object variables if the referenced lookup has a hyphen in the classname or id?


chrisngobanh:
How should I name my jQuery object variables if the referenced lookup has a hyphen in the classname or id?

i.e. `const $??? = $('foo-btn');`

My take (camelCase): `const $fooBtn = $('foo-btn');`

Refer to [Section 25](https://github.com/airbnb/javascript#jquery)

chrisngobanh:
I created pull request #467, which addresses this by adding a case in [Section 25.1](https://github.com/chrisngobanh/javascript#25.1).

PR #467, opened 2015-08-10
Updated 25.1 to include a case with a lookup that has a hyphen in the classname

chrisngobanh:
I updated [Section 25.1](https://github.com/chrisngobanh/javascript#25.1) to include a case with a lookup that has a hyphen in the classname.
The naming convention that I have chosen for this is camelCase.

I felt that this is necessary because it isn't obvious what to do in this case.

Addresses #466

### DIFF
--- a/README.md
+++ b/README.md
@@ -1874,6 +1874,9 @@
 
     // good
     const $sidebar = $('.sidebar');
+
+    // good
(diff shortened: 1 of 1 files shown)


goatslacker:
Sure thing 👍🏼

PR #471, opened 2015-08-11
Add ComparaOnline to 'In The Wild' List

KimDongHwan:
We are trying to keep some order here at [ComparaOnline](http://www.comparaonline.cl/) :)


hshoff:
:+1:

PR #472, opened 2015-08-12
Implicit return + single expression

### LINKED ISSUE #438: Implicit return – one line vs. one expression

tomekwi:
Fixes #438.

I’ve upgraded the note about argument parens to a new rule – it reads better this way.

### DIFF
--- a/README.md
+++ b/README.md
@@ -597,9 +597,7 @@
     });
 
     // good
-    [1, 2, 3].map((x) => {
-      return x * x;
-    });
+    [1, 2, 3].map(x => x * x);
     ```
 
   - [8.2](#8.2) <a name='8.2'></a> If the function body consists of a single expression, feel free to omit the braces and use the implicit return. Otherwise use a `return` statement.
--- a/README.md
+++ b/README.md
@@ -602,8 +602,6 @@
 
   - [8.2](#8.2) <a name='8.2'></a> If the function body consists of a single expression, feel free to omit the braces and use the implicit return. Otherwise use a `return` statement.
 
-    In case the expression spans over multiple lines, wrap it in parentheses for better readability.
-
   > Why? Syntactic sugar. It reads well when multiple functions are chained together.
 
   > Why not? If you plan on returning an object.
@@ -611,12 +609,6 @@
     ```javascript
     // good
     [1, 2, 3].map(number => `A string containing the ${number}.`);
-
-    // good
-    [1, 2, 3].map(number => (
-      `As time went by, the string containing the ${number} became much ` +
-      'longer. So we needed to break it over multiple lines.'
-    ));
 
     // bad
     [1, 2, 3].map(number => {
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -597,12 +597,12 @@
...
-    [1, 2, 3].map((x) => {
-      return x * x;
-    });
+    [1, 2, 3].map(x => x * x);
     ```
 
-  - [8.2](#8.2) <a name='8.2'></a> If the function body fits on one line and there is only a single argument, feel free to omit the braces and parentheses, and use the implicit return. Otherwise, add the parentheses, braces, and use a `return` statement.
+  - [8.2](#8.2) <a name='8.2'></a> If the function body consists of a single expression, feel free to omit the braces and use the implicit return. Otherwise use a `return` statement.
+
+    In case the expression spans over multiple lines, wrap it in parentheses for better readability.
```
hshoff: Let's make this `8.3` and have the good bad examples nested up underneath.

README.md
```diff
@@ -610,12 +610,37 @@
...
+    });
+
+    // good
+    [1, 2, 3].map(number => {
+      let nextNumber = number + 1;
+      return `A string containing the ${nextNumber}.`;
+    });
+    ```
+
+  - [8.3](#8.3) <a name='8.3'></a> If your function only takes a single argument, feel free to omit the parentheses.
```
hshoff: and then let's bump this to 8.4

README.md
```diff
@@ -597,12 +597,12 @@
     });
 
     // good
-    [1, 2, 3].map((x) => {
-      return x * x;
-    });
+    [1, 2, 3].map(x => x * x);
```
hshoff: Let's keep the original example here...or let's update the example function to be more than one-line.

We try to introduce one-rule by example at a time prevent confusion. In this example, we have the arrow function, the fits on one line, and the only one argument rules being applied.

hshoff:
Couple of comments, love the added clarity. Thanks @tomekwi!

README.md
tomekwi: Good point.

>  let's update the example function to be more than one-line

I guess that’d be better – then all our examples will follow all rules :)

tomekwi:
Thanks for the comments @hshoff :+1: .

I can come back to this tomorrow morning, European time. Until then!

tomekwi:
:checkered_flag: 

Sorry for scattering the changes over so many commits. I can rebase if you don’t mind losing your line comments @hshoff.

hshoff:
:trophy:

README.md
```diff
@@ -593,16 +593,18 @@
...
     // good
     [1, 2, 3].map((x) => {
-      return x * x;
+      const y = x + 1;
+      return x * y;
     });
     ```
 
-  - [8.2](#8.2) <a name='8.2'></a> If the function body fits on one line and there is only a single argument, feel free to omit the braces and parentheses, and use the implicit return. Otherwise, add the parentheses, braces, and use a `return` statement.
+  - [8.2](#8.2) <a name='8.2'></a> If the function body consists of a single expression, feel free to omit the braces and use the implicit return. Otherwise use a `return` statement.
```
goatslacker: This change doesn't hold the original intent of the text where parens are ok to omit only when you're using implicit return, otherwise they're required. cc @ljharb

ljharb: Let's please revert this text then, and also ensure the eslint plugin enforces it.

PR #474, opened 2015-08-15
[React] Fixed Incorrect Class Declaration in Section 10 

chrisngobanh:
In the bad example for Section 10, the class that was defined was using half ES5 syntax and half ES6 syntax. That means there was a 100% of it producing a syntax error. I fixed it.

### DIFF
--- a/react/README.md
+++ b/react/README.md
@@ -220,7 +220,7 @@
   - Do not use underscore prefix for internal methods of a React component.
     ```javascript
     // bad
-    React.createClass({
+    class extends React.Component {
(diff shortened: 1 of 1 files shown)


ljharb:
There's no syntax error here. In ES6, concise methods work just fine in object literals too. The difference is that in "class" constructs, no comma is used between method definitions - but since this example has just one method, that doesn't apply.

chrisngobanh:
@ljharb Ah, right! That explains why the same "error" was being repeated earlier.

PR #475, opened 2015-08-18
Recommend exporting `propTypes` as a named export

ljharb:
By exporting `propTypes` as a named export *as well* as attaching them to the "class", they can be explicitly imported by tests and other components in a way that aligns with the future ES6 module dependency graph, rather than as an arbitrary property that isn't statically verifiable. Doing both is great imo.

Thoughts?

### DIFF
--- a/react/README.md
+++ b/react/README.md
@@ -262,7 +262,7 @@
   ```javascript
   import React, { Component, PropTypes } from 'react';
   
-  const propTypes = {
+  export const propTypes = {
     id: PropTypes.number.isRequired,
     url: PropTypes.string.isRequired,
     text: PropTypes.string,


nkbt:
Exporting prop types gives you consistent types in the app and higher order types. It is a really cool thing (we already use this for a while). But we keep `PropTypes.js` file with all custom props, instead of pulling props from particular modules.

ljharb:
That's certainly another valid alternative - I'm just proposing amending our current approach (attaching propTypes as a property on the default export "class") :-)

justjake:
k

goatslacker:
I think we should revert this.

The main issue I have is that when we switch to the ultimate form

```js
class X extends Component {
  static propTypes = {
    ...
  }
}
```

You won't be able to import propTypes. I don't want to get in the habit of importing propTypes now.

If you need the propTypes you can import the component and pull them from there I think that should be the blessed way to do things.

ljharb:
@lelandrichardson and I were just discussing this, and came to a similar conclusion - namely, that instead of sharing propTypes, we should be sharing Shapes, and *not* exporting propTypes.

We're thinking we'll chat about it in the arch meeting tomorrow, and if everyone's cool, one of us will make a PR to explain that philosophy more in-depth.

PR #491, opened 2015-08-28
[react] inline export syntax

gilbox:
b more 1337

### DIFF
--- a/react/README.md
+++ b/react/README.md
@@ -85,10 +85,8 @@
     });
 
     // good
-    class ReservationCard extends React.Component {
+    export default class ReservationCard extends React.Component {
     }
- 
-    export default ReservationCard;
     ```
 
 ## Alignment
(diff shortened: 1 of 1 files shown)


justjake:
LGTM, but this may be controversial with the original guide authors? Some people complained about having exports other than at the very bottom of the file.

ljharb:
In this specific case, I like this change, because 100% of the code below the default export is something that will be moved into inline `static` properties at some point. In all other cases, I'd want the exports to be the final thing in any file. However, I'm not sure if the inconsistency is clearly explained.

gilbox:
If the goal is to place all exports at the bottom, then [at least one](https://github.com/airbnb/javascript/tree/master/react#ordering) code example needs to be updated.

goatslacker:
yeah sure

ljharb:
(I opened https://github.com/airbnb/javascript/pull/492 to fix that code example)

justjake:
I'm merging this because I agree and discussion stopped.

PR #492, opened 2015-08-28
do not export propTypes

ljharb:
We decided to stick with the decision to not export `propTypes`, and leave them as only a static property on the default export.

/cc @goatslacker @gilbox


goatslacker:
:+1:

PR #494, opened 2015-08-30
Default parameters go last.

johnmanong:
Do you think it's worthwhile to explicitly call out the order when only some parameters have a default? 

[The compiled output is pretty different](https://babeljs.io/repl/#?experimental=true&evaluate=true&loose=false&spec=false&code=%2F%2F%20bad%0Afunction%20handleThings(opts%20%3D%20%7B%7D%2C%20name)%20%7B%0A%20%20%2F%2F%20...%0A%7D%0A%0A%2F%2F%20good%0Afunction%20handleThings(name%2C%20opts%20%3D%20%7B%7D)%20%7B%0A%20%2F%2F%20...%0A%7D). Being new-ish to es6, this seems pretty important to note, but maybe too much of a NSD.

### DIFF
--- a/README.md
+++ b/README.md
@@ -556,21 +556,7 @@ Other Style Guides
     }
     ```
 
-  - [7.8](#7.8) <a name='7.8'></a> Always put default parameters last.
-
-    ```javascript
-    // bad
-    function handleThings(opts = {}, name) {
-      // ...
-    }
-
-    // good
-    function handleThings(name, opts = {}) {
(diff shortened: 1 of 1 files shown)


goatslacker:
Can you flip these? Leave 7.8 as-is and add 7.9 as the new rule.

johnmanong:
@goatslacker done.

goatslacker:
Thank you!

PR #505, opened 2015-09-13
Added id-length rule to eslintrc

amsardesai:
This helps make sure variable names are longer than 1 character.

Single letter variable names are almost guaranteed to be completely non descriptive, this makes sure our variable names are longer than that.

Documentation on the id-length rule: http://eslint.org/docs/rules/id-length.html

### DIFF
--- a/packages/eslint-config-airbnb/base/index.js
+++ b/packages/eslint-config-airbnb/base/index.js
@@ -128,6 +128,10 @@ module.exports = {
     'quotes': [
       2, 'single', 'avoid-escape'    // http://eslint.org/docs/rules/quotes
     ],
+    'id-length': [2, {
+      'min': 2,
+      'properties': 'never'
+    }],
     'camelcase': [2, {               // http://eslint.org/docs/rules/camelcase
       'properties': 'never'
     }],
--- a/packages/eslint-config-airbnb/base/index.js
+++ b/packages/eslint-config-airbnb/base/index.js
@@ -128,7 +128,7 @@ module.exports = {
     'quotes': [
       2, 'single', 'avoid-escape'    // http://eslint.org/docs/rules/quotes
     ],
-    'id-length': [2, {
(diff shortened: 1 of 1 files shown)


justjake:
please add the id-length documentation URL to your PR as a comment, following the style of the rest of the eslintrc.

amsardesai:
Done!

toxicFork:
@amsardesai What variable do you recommend using for "for loops"?

amsardesai:
@toxicFork I generally try to avoid using for loops if I could use `map`, `forEach`, or `filter` instead. If I have to use for loops, I think just calling the variable name `index` seems like the best way to do it.

```js
for (var index = 0; index < someLength; index++) {
    ...
}
```

What do you think?

macebedo:
For the "for loop"... I've always use "ctr". These loops involves counter. ctr for short. Simple enough for me.

gausie:
What are people's opinions on adding an exception for "_", typically used by lodash and underscore?

tomekwi:
@justjake @hshoff I also got bitten by it. There are other valid uses:
* `h` in [*hyperscript*](http://npm.im/hyperscript), [*virtual-dom*](http://npm.im/virtual-dom) and [*@cycle/core*](http://npm.im/@cycle/core).
* `m` in [*mithril*](http://npm.im/mithril).
* `$`, we all know where. Brr!

toxicFork:
Also anything to do with geometry and math, e.g. THREE.Vector3.{x, y, z}
and many others. It will remain disabled in all my configs.

tomekwi:
@toxicFork Yeah, true.

Generally I’m no fan of one-letter variables – but how else can you name `x`?

justjake:
Turns out no one likes this rule. Will turn it back off eventually

tomekwi:
:+1: :+1:

gausie:
Thanks!

PR #506, opened 2015-09-14
Remove JSX from eslint-config-airbnb/base

ariporad:
Hi,

JSX was present as an `ecmaFeature` in base, so I moved it to the react-specific settings. Also, because I would have had to add another loop for merging `ecmaFeatures`, I decided to just use lodash to merge base and react. All the tests (still) pass.


justjake:
I'd prefer to not have any dependencies -- just do the merge manually; it's fine that way IMO.

PR #507, opened 2015-09-14
Allow unused function parameters.

### LINKED ISSUE #426: Why does the .eslintrc prohibit unused function parameters?
Hi,

I'm coming at this from a node point of view, so sorry if it's weird for client side JS.

Why do you prohibit unused function parameters in the .eslintrc? In node, (and AFAIK in the browser), ignoring one or more of your parameters for now isn't that big a deal (for example, if a callback gets err, info, res, ignoring info is fine.)

Thanks!
Ari

ariporad:
Hi,

I accordance with #426, here's a PR which allows unused function parameters. While I was doing it, I noticed #506, which this depends on (actually, it includes all those commits as well).

(Testing GitHub:) This Fixes #426 

Thanks!

### DIFF
--- a/packages/eslint-config-airbnb/base/index.js
+++ b/packages/eslint-config-airbnb/base/index.js
@@ -19,8 +19,7 @@ module.exports = {
     'objectLiteralShorthandProperties': true,
     'spread': true,
     'superInFunctions': true,
-    'templateStrings': true,
-    'jsx': true
+    'templateStrings': true
   },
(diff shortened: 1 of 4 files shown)


justjake:
as I said in the other PR, i would prefer not to add dependencies. I'll merge this once you drop lodash.*

ariporad:
I'm going to close this and open a new one, I've messed it up somehow.

PR #508, opened 2015-09-15
Added clarifying statement to React README

Ryan-McBride:
Added a clarifying statement to explain why the underscore prefix isn't used in React.

### DIFF
--- a/react/README.md
+++ b/react/README.md
@@ -217,7 +217,7 @@
     ```
 
 ## Methods
-  - Do not use underscore prefix for internal methods of a React component.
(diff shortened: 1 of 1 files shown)


justjake:
They're not private by default. It's straightforward to call them from a parent component using refs.

ljharb:
Closing this per https://github.com/airbnb/javascript/pull/508#issuecomment-140566824. In addition, underscore-prefixed methods shouldn't be used *ever* in any context, since they provide a false sense of privacy and encourage breaking changes in a non-major release.

Issue #512, opened 2015-09-16
Hyper Oslo using your style guide


sindrenm:
Hey guys,

The devs here at Hyper in Oslo, Norway love your style guide. Even though we didn't fork it (like normal people), we have a document based off of it here:

https://github.com/hyperoslo/javascript-playbook/blob/master/style.md

So thank you guys so much for putting this together, and if you feel like adding us to that list of yours then you're more than welcome!

goatslacker:
Create a PR adding it to the list.

sindrenm:
Thanks. Have a look ta #516. :smiley:

PR #517, opened 2015-09-17
Remove issues from how to add to “in the wild”

sindrenm:
https://github.com/airbnb/javascript/issues/512#issuecomment-140903596 suggests that creating an issue in order to get added to the “In the Wild” list is not sufficient, so I propose removing that from the instructions. :smiley:


justjake:
indeed

PR #518, opened 2015-09-18
Update eslint

ariporad:
Update eslint to v1.4.x

Fixes #515.

PR #514, opened 2015-09-16
deprecated react/jsx-quotes

brad-decker:
replaced this rule with the correct rule. Resolves #513

PR #519, opened 2015-09-19
Fix jsx quotes

axyz:
N.B. this pull should be merged after #518 as jsx-quotes is a 1.4.x feature

Now jsx-quotes is included on vanilla eslint and must be used instead of `react/jsx-quotes` that is now deprecated.

Also deprecated rules causes flycheck on emacs to break.


axyz:
seems to be an already addressed problem #514

Issue #523, opened 2015-09-25
Possible typo at #21.3


DannyNemer:
At item [21.3](https://github.com/airbnb/javascript#21.3), the guide reads:
> Use `parseInt` for Numbers and always with a radix for type casting.

However, the example displays:
```javascript
// good
const val = Number(inputValue);

// good
const val = parseInt(inputValue, 10);
```
**Question**: Is the first of the above examples supposed to be `// bad` (not `// good`), or should the guide read "Use `parseInt` **or `Number`** for Numbers..."?

Thank you.

ljharb:
`parseInt` is only meant to take a `String`, and will also discard non-number characters - ie, `parseInt('3a', 10)` will be `3`, but `Number('3a')` will be `NaN` - so they're slightly different there. `Number` is also the proper way to coerce non-string values to numbers.

My guess is that the wording should be changed to encourage proper usage of both.

DannyNemer:
Thank you for your response. I agree and opened a pull request to address this: #524.

PR #524, opened 2015-09-25
Clarify proper type casting for Numbers

DannyNemer:
Clarifies when to use `Number` vs. `parseInt` for Number type casting, encouraging the proper usage of both.

Fixes: #523

Previously, at item [21.3](https://github.com/airbnb/javascript#21.3), the guide read:
> Use `parseInt` for Numbers and always with a radix for type casting.

However, the example displays:
```javascript
// good
const val = Number(inputValue);

// good
const val = parseInt(inputValue, 10);
```
**Issue**: It is unclear whether `Number` is acceptable because the guide only directly states to use `parseInt` despite the example for `Number`. Indeed, `Number` and `parseInt` are [not interchangeable](https://github.com/airbnb/javascript/issues/523#issuecomment-143229271) and should both be used properply.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1640,7 +1640,7 @@ Other Style Guides
     const totalScore = String(this.reviewScore);
     ```
 
-  - [21.3](#21.3) <a name='21.3'></a> Use `parseInt` for Numbers and always with a radix for type casting.
+  - [21.3](#21.3) <a name='21.3'></a> Numbers: Use `Number` for type casting and `parseInt` always with a radix for parsing strings.
 
     ```javascript
     const inputValue = '4';


README.md
```diff
@@ -1640,7 +1640,7 @@ Other Style Guides
     const totalScore = String(this.reviewScore);
     ```
 
-  - [21.3](#21.3) <a name='21.3'></a> Use `parseInt` for Numbers and always with a radix for type casting.
+  - [21.3](#21.3) <a name='21.3'></a> Numbers: Use `Number` for type casting and `parseInt` with a radix for parsing.
```
ljharb: could this say `parsing strings`, and also continue to say `always with a radix`?

DannyNemer: Done!

ljharb:
👍 this LGTM

Issue #496, opened 2015-09-02
ES5 settings for eslint


honzajavorek:
There seem to be ES5 rules for `eslint` created by someone else: https://github.com/1hella/eslint-config-airbnb-es5 Wouldn't it be worth to merge these two projects and provide ES5 configuration out of the box the same way as the one for ES6 is?

ljharb:
@honzajavorek https://www.npmjs.com/package/eslint-config-airbnb-base - `"extends": "airbnb/base"`. If this doesn't suffice, please file specific issues and I'll be happy to address them.

@1hella: if you'd like to merge your project into ours, please file issues/PRs as needed.

gaboesquivel:
@ljharb  I'm trying to setup ES5 for an angular 1.x client in our project, we use ES6 on the backend. 
however it doesn't seem to work out-of-the-box, specifically I'm seeing prefer arrow functions errors.
It would be nice if I could extend ES5 explicitly. I had to use `eslint-config-airbnb-es5` since I didn't want to do rule overwrites.

ljharb:
We recommend using Babel and writing ES6. That said, I'd accept a PR that exposed an `es5` entry point to obviate the need for `eslint-config-airbnb-es5`.

gaboesquivel:
got it, thanks

shreyas-a:
I'm trying to use airbnb/legacy, but it fails to find the errors while airbnb/base works.

brownieboy:
airbnb/base flagged my ES5 code with `no-var` errors, i.e. I have to use `let` and `const` instead of `var`.   But `let` and `const` are ES6 not ES5!!

So I've gone with eslint-config-airbnb-es5 too.

ljharb:
@brownieboy please file a new issue, and i can update the legacy entry point to disable `no-var`.

Issue #366, opened 2015-06-18
eslint shareable configs / split


orlin:
Because the react plugin / jsx rules aren't optional with this style-guide, I found out there is an *official* way for [sharing eslint configs](http://eslint.org/docs/developer-guide/shareable-configs).  The "standard" style, with [standard-react](https://github.com/feross/eslint-config-standard-react), makes a good example of splitting `eslint` rules.  Not all js is front-end, and using react...

justjake:
Ok, sounds good. I will try to move React styleguide options from `eslint-config-airbnb` to `eslint-config-airbnb-react` next week. I'm also planning to do an audit of the `.eslintrc` to make sure it conforms with both the letter and the spirit of the styleguide.

@goatslacker @hshoff should we have 3 separate ESlintrcs? `eslint-config-airbnb`, `eslint-config-airbnb-es6`, `eslint-config-airbnb-react`? I would prefer not to split out the ES6 rules since those are considered part of the core style guide, but I'm up for discussion on the matter. I would prefer to only maintain two, `eslint-config-airbnb` and `eslint-config-airbnb-react`.

goatslacker:
+1 on just two. We can have a third es5 only if people really want that but since ES2015 is a final draft and its actual JS now I vote we move forward with no distinction between the two.

justjake:
:+1: ok

ljharb:
The spec being finalized is not the same as being free to use ES6 sans transpilation - ES5 was finalized in 2009 and it wasn't safe to use getters/setters for many many years. I'd prefer have an ES5 one as well for those who don't wish to use babel, and wish to support browsers.

justjake:
:+1: ok

baer:
Hey guys, we (walmartlabs) have been grappling with updating our ESLint config as well. Both making sure that it works in multiple environments (ES5, ES6, React, Node, Browser, Test etc.) and we decided to take the same approach and split our config out into an ES5 base + extensions. It's been working out well so far.

We also had issues with protecting CI from new rules that get turned on when we update ESLint and maintaining the giant YAML files so we created [eslint-config-defaults](https://github.com/walmartlabs/eslint-config-defaults). It starts with a reset where everything is off (this will be default in the upcoming ESLint 1.x anyway) then builds up config with composable rule sets using ESLint's [sharable config](http://eslint.org/docs/developer-guide/shareable-configs). 

[...]

PR #526, opened 2015-09-25
Eslint config defaults

### LINKED ISSUE #522: Switch to rules from eslint-config-defaults
@justjake: Do you still plan to proceed with https://github.com/airbnb/javascript/issues/454#issuecomment-134410088? If so, are you looking for a PR to do this?

taion:
Closes #522.

@justjake: I can see why you didn't want to do this. This was a lot more time-consuming than I had anticipated.

It's split up into 3 commits:

1. Bring in the rules from `eslint-config-defaults` and plug them in.
2. Account for the mismatches against what's currently in 0.0.9.
3. Update for changes in latest versions of ESLint and eslint-plugin-react.


packages/eslint-config-airbnb/index.js
```diff
@@ -1,13 +1,7 @@
...
-// manually merge in React rules
-eslintrc.plugins = reactRules.plugins;
-Object.keys(reactRules.rules).forEach(function assignRule(ruleId) {
-  eslintrc.rules[ruleId] = reactRules.rules[ruleId];
-});
-
-module.exports = eslintrc;
+module.exports = {
+  'extends': [
+    // Need to qualify these here for ESLint to resolve them properly.
```
justjake: can you not just reference them relatively? I think that issue was why I was constructing a rules object instead of just extending 'base' previously. Can eslint resolve these when running `npm run lint` without special linking magicks?

justjake:
SO GOOD! I'm very excited for this; will make it easy to fix things like #496 in the next few PRs.

@hshoff you should look this over and make sure you're ok with this.
@baer how does this satisfy #366 for you? Is the structure similar enough? Or do you want index.js files in folders?

hshoff:
Lovely! Love open source. Thanks @taion! :beers:

packages/eslint-config-airbnb/index.js
taion: `npm run lint` didn't work when referencing them relatively. It interpreted them as the names of `eslint-config-*` packages for some reason.

It's possible that doing something like https://github.com/walmartlabs/eslint-config-defaults/blob/master/package.json#L34 involving making `main` not be `index.js` could work, but I didn't try it, because this approach seemed tolerable.

ljharb: I believe all `extends` in `eslint` have to be named starting with "eslint-config", or a scoped package with the name ending in "/eslint-config".

taion: @ljharb I have no clue - the relative imports work here: https://github.com/airbnb/javascript/pull/526/files#diff-ca246d3b4ba146243647f872c0e7e8d1R3, but they didn't work on `index.js`. ¯\\\_(ツ)_/¯

baer:
This is super awesome guys! Nice work!

PR #534, opened 2015-10-01
Add reasoning for using double quotes in JSX

cooperka:
Add reasoning for using double quotes in JSX, since it's the opposite of the recommendation for [JS strings](https://github.com/airbnb/javascript#strings). There may be other reasons as well, but the one added here seems the most compelling to me.

### DIFF
--- a/react/README.md
+++ b/react/README.md
@@ -117,6 +117,9 @@
 
 ## Quotes
   - Always use double quotes (`"`) for JSX attributes, but single quotes for all other JS.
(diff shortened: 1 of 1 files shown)


ljharb:
There's also that HTML attributes typically (conventionally) use double quotes, not single quotes, but that may not be as compelling a reason to be front-and-center.

cooperka:
Good point @ljharb, added.

PR #537, opened 2015-10-04
Add npmignore to eslint-config-airbnb

amsardesai:
It only makes sense now that there's a `test` directory.


ljharb:
Thanks, but I'd prefer not to ignore this. Test files should always be published to `npm`, so that `npm explore foo && npm install foo && npm test` always works.

PR #542, opened 2015-10-12
Allow '==' to check for null and undefined in one go.

ludovicofischer:
Using `==` against `null` is a common pattern (see [here](https://github.com/chenglou/react-motion/blob/3aa46cbbc2c136cbbf535538c58b2161978eef6f/src/stripStyle.js) for example). Since ESLint caters to it, should it not be allowed? Relevant ESLint documentation: http://eslint.org/docs/rules/eqeqeq

### DIFF
--- a/packages/eslint-config-airbnb/rules/best-practices.js
+++ b/packages/eslint-config-airbnb/rules/best-practices.js
@@ -17,7 +17,7 @@ module.exports = {
     // enforces consistent newlines before or after dots
     'dot-location': 0,
     // require the use of === and !==
-    'eqeqeq': 2,
+    'eqeqeq': [2, 'allow-null'],
     // make sure for-in loops have an if statement
     'guard-for-in': 2,
     // disallow the use of alert, confirm, and prompt


ljharb:
@ludovicofischer are you still interested in landing this PR? if so, please rebase it on top of latest master. I'd be happy to land it.

ludovicofischer:
Hi, thanks for the heads up.

ljharb:
@ludovicofischer I ended up committing it because I'm hoping to release soon :-) thanks for the contribution!

spikebrehm:
Thanks @ljharb!

mhofman:
Any reason not to go all the way and allow `smart` equals? basically same as `allow-null` with relaxation when the type is known, e.g. literals comparison or more importantly `typeof`

ljharb:
@mhofman yes - that setting presents a refactoring hazard when one is changing a literal to a variable. `== null` is special, and otherwise, `===` should be used without exception.

mhofman:
Well the lint rule should still catch it if the refactoring change didn't also change the `==` to `===` ;)

ljharb:
yes that is a fair point. but `== null` makes me uncomfortable enough; I don't want to train developers that `==` is ever OK.

PR #546, opened 2015-10-15
New feature for Classes and Objects definition

teamcreed:
It is very important to consider OOP, because one of the most important aspects writing maintainable code is being able to notice the recurring themes in that code and optimize them.

### DIFF
--- a/README.md
+++ b/README.md
@@ -38,6 +38,7 @@ Other Style Guides
   1. [Accessors](#accessors)
   1. [Events](#events)
   1. [jQuery](#jquery)
+  1. [OOP](#oop)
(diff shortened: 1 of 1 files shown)


goatslacker:
https://github.com/airbnb/javascript#constructors

goatslacker:
We should probably rename that to Classes.

teamcreed:
Yes, you can rename it 

are pattern designs considered in this guide?

ljharb:
Pattern designs that we use at Airbnb are definitely worth including.

In this case, a constructor function that isn't an ES6 "class" contradicts https://github.com/airbnb/javascript#constructors, so I think this needs to be closed as-is.

Issue #483, opened 2015-08-22
[ESLint] Why are padded blocks not permitted?


chrisngobanh:
I don't quite understand why the ESLint config file prohibits padded blocks. What would happen if you needed to make a comment after a block opens: 

```
// ESLint Error: "Block must not be padded by blank lines. (padded-blocks)"
$document.ready(() => {

  // Initialize/Render React
  React.render(React.createElement(Index), $react.get(0));
  ...
}

// Violates 17.2 of the style guide: "Put an empty line before the [single-line] comment."
$document.ready(() => {
  // Initialize/Render React
  React.render(React.createElement(Index), $react.get(0));
  ...
}
```

In addition, I don't see anywhere in the style guide where padded blocks are prohibited.

justjake:
Yeah, the eslint config is out of sync with our actual styleguide in a number of places. I am working on a test suite to bring them together here:

https://github.com/airbnb/javascript/pull/480

justjake:
However in this case I think we should amend the styleguide to say that a comment on the first line of a block does not need a preceding line.

chrisngobanh:
I don't see anywhere in the style guide that mentions padded blocks, so maybe we should make a change in the styleguide to address it as well.

chrisngobanh:
#547 addresses this issue by adding [a section to the style guide that prohibits padded blocks](https://github.com/airbnb/javascript#18.8).

jbcpollak:
I know this question is closed, but I am wondering if there is a justification for _why_ padded blocks aren't allowed? I noticed both AirBnB and Google's styles disallow it, so we are following the same convention, but we are wondering _why_.

ljharb:
Whitespace is beneficial when it adds readability, but not when it creates visual noise. With syntax highlighting, lines padding blocks add zero benefit, but increase the vertical length of code, and reduce the amount you can see at once.

jbcpollak:
@ljharb - good enough for me, thanks! Would be awesome to see that in the style guide (I often wish style guides had more justifications) but that is very helpful.

dietergeerts:
@ljharb , that's not true for all cases. In several cases, adding a blank line add a HUGE readability improvement to the code, and it outweighs the fact that the code gets longer by 1 single line. For example in a test file:

```
import { withJsonDefault } from './with-json-default';

describe('withJsonDefault', () => {

  it(`must use the serializer of the given schema directly,
    as defaults are only meant to be used on deserialization,
[...]

PR #547, opened 2015-10-16
Prohibit Padded Blocks and Allow Comments in the First Line of a Block

chrisngobanh:
In #483, @justjake says that we should amend the style guide to allow single line comments in the first line of a block. I updated 17.2 with this change.

In addition, our ESLint config file prohibits padded blocks, so I decided to make a section 18.8 to address that.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1248,7 +1248,7 @@ Other Style Guides
     }
     ```
 
-  - [17.2](#17.2) <a name='17.2'></a> Use `//` for single line comments. Place single line comments on a newline above the subject of the comment. Put an empty line before the comment.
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -1511,6 +1519,34 @@ Other Style Guides
     return arr;
     ```
 
+  - [18.8](#18.8) <a name='18.8'></a> Do not pad your blocks with blank lines.
+
+    ```javascript
+    // bad
+    function bar() {
```
justjake: It would be nice to show one of these examples as an if...else

justjake:
Nice! Thanks for the PR

README.md
chrisngobanh: Sure thing! I just made my change; let me know what you think!

PR #554, opened 2015-10-23
eslint: Add missing `space-before-keywords` rule

adrienverge:
All examples in README.md seem to agree on avoiding missing whitespaces before keywords such as:

```js
if (cond) {
}else {       // no space before 'else'
}

try {
}catch (e) {  // no space before 'catch'
}
```

This patchs adds the `space-before-keywords` rule as an error (as is `space-after-keywords` already).

### DIFF
--- a/packages/eslint-config-airbnb/rules/style.js
+++ b/packages/eslint-config-airbnb/rules/style.js
@@ -88,8 +88,10 @@ module.exports = {
     'semi': [2, 'always'],
     // sort variables within the same declaration block
     'sort-vars': 0,
+    // require a space before certain keywords
+    'space-before-keywords': [2, 'always'],
     // require a space after certain keywords
-    'space-after-keywords': 2,
+    'space-after-keywords': [2, 'always'],
     // require or disallow space before blocks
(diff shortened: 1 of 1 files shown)


ljharb:
LGTM :+1: 

(altho it might be nice to specify `[2, "always"]` on both keyword spacing rules to make it clear what behavior we want, even though it's the default)

adrienverge:
@ljharb Good idea, done! Also added "always" to `space-after-keywords` to be clear.

ljharb:
Thanks! I don't see the diff yet; perhaps it's not yet committed and/or pushed?

adrienverge:
Yeah, sorry, didn't pushed to the right repo :)

ljharb:
Thanks, this looks great! I'll defer merging this to @justjake.

justjake:
:+1:

taion:
Isn't the `'always'` redundant here, given that it's the default?

ljharb:
`s/redundant/explicit`. Just because it's the default doesn't mean we shouldn't explicitly specify it.

PR #556, opened 2015-10-23
Changed syntax highlighting from JavaScript to es6

mjkaufer:
Also added a code block around a stray `eval()`

### DIFF
--- a/README.md
+++ b/README.md
@@ -60,7 +60,7 @@ Other Style Guides
     + `null`
     + `undefined`
 
-    ```javascript
+    ```es6
     const foo = 1;
     let bar = foo;
 
@@ -74,7 +74,7 @@ Other Style Guides
     + `array`
     + `function`
 
-    ```javascript
+    ```es6
     const foo = [1, 2];
     const bar = foo;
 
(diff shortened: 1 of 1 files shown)


justjake:
Dang, does Github's Javascript highlighter not work on these examples? I would prefer to keep `javascript` if it works fine because this code will eventually be normal javascript, but if this makes it pretty, then I'm in.

ljharb:
It doesn't appear to make a difference to me:
```es6
class Foo { }
const { a } = { a: b => c };
```

```javascript
class Foo { }
const { a } = { a: b => c };
```

hshoff:
This should be fixed from a while ago: https://github.com/github/linguist/issues/2398

README.md
```diff
@@ -447,7 +447,7 @@ Other Style Guides
       return `How are you, ${name}?`;
     }
     ```
-  - [6.5](#6.5) <a name='6.5'></a> Never use eval() on a string, it opens too many vulnerabilities.
+  - [6.5](#6.5) <a name='6.5'></a> Never use `eval()` on a string, it opens too many vulnerabilities.
```
ljharb: this is the only change in this PR that i think should stay

hshoff: Yup! fixed here: https://github.com/airbnb/javascript/commit/f130c0ffa26ae24f2e18bac33bc5cf7f62a2d938

PR #557, opened 2015-10-25
J query new feature about some rules of jQuery plugins

teamcreed:


README.md
```diff
@@ -1983,6 +1983,36 @@ Other Style Guides
     // good
     $sidebar.find('ul').hide();
     ```
+    
+  - [25.5](#25.5) <a name='25.5'></a> jQuery plugins have few concrete rules, which is one of the reasons for the incredible diversity in how they are implemented across the community. At the most basic level, we can write a plugin simply by adding a new function property to jQuery’s `jQuery.fn` object, as follows:
+    
+    ```javascript
+    $.fn.myPluginName = function () {
```
ljharb: imo best practice is to not create or use jQuery plugins, since they involve modifying a global - I'd prefer a pure function that takes a jQuery object and operates on it.

justjake:
We try not to write jQuery plugins at Airbnb. I think it's much better to share code using regular modules, and then `import myFunction from 'myModule'`

PR #563, opened 2015-11-02
ESLint: excempt '$' and '_' from 'id-length' rule

MhSaadany:
Because '$' and '_' and commonly used as identifiers for the jQuery
and lodash libraries respectively. This commit adds both identifiers to
the 'id-length' rule exception list.


justjake:
Closed by https://github.com/airbnb/javascript/pull/569

Issue #567, opened 2015-11-05
id-length exceptions


jacobp100:
* `$` and `_` should be allowed, as these are frequent for libraries
* `x`, `y`, `z` can't really be made longer
*  `e` is frequently used for the event object

May have forgotten a few

PR #510, opened 2015-09-16
Added max-len rule

screendriver:
Closes #509

PR #569, opened 2015-11-05
[eslint] remove id length rule (too much nutrage)

justjake:
Will close: #567 #563, #510 

It seems like everyone is nutraged about one letter variable names. I know I am. Let's just turn this rule off and forget we ever thought it was a good idea. If you find yourself having a problem with one-letter variable names on your team, you can easily re-enable this rule.

@ljharb @hshoff


ljharb:
Agreed. We could still set "min" to `0` and set a "max", but this is a good change.

PR #570, opened 2015-11-06
update `eslint`

### LINKED ISSUE #565: Why is babel-eslint required?
ESLint supports all of ES6 plus the experimental object rest/spread that React uses. Is it possible to remove babel-eslint as a dependency? (It would make linting faster.)

ljharb:
This updates `eslint`, `eslint-plugin-react`, and `tape` to the latest version. It also removes the `babel-eslint` dependency, which fixes #565.

### DIFF
--- a/packages/eslint-config-airbnb/base.js
+++ b/packages/eslint-config-airbnb/base.js
@@ -3,6 +3,5 @@ module.exports = {
     'eslint-config-airbnb/legacy',
     'eslint-config-airbnb/rules/es6',
   ],
-  'parser': 'babel-eslint',
   'rules': {}
 };
--- a/packages/eslint-config-airbnb/package.json
+++ b/packages/eslint-config-airbnb/package.json
@@ -26,7 +26,6 @@
   },
   "homepage": "https://github.com/airbnb/javascript",
   "devDependencies": {
-    "babel-eslint": "4.1.3",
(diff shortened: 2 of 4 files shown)


packages/eslint-config-airbnb/.eslintrc
```diff
@@ -1,8 +1,32 @@
...
+  "plugins": [
+  ],
+
   "rules": {
     // disable requiring trailing commas because it might be nice to revert to
     // being JSON at some point, and I don't want to make big changes now.
     "comma-dangle": 0
+  },
+
+  "ecmaFeatures": {
```
justjake: is this needed? won't `"extends": "airbnb"` take care of this

ljharb: this *is* `eslint-config-airbnb`, so it can't really extend itself

justjake: no this is what .eslintrc to extend itself. it should work!

justjake:
Should we specify eslint as a peerDep given that this change needs eslint >= 1? or should we just bump the major version? hm

justjake:
anyways, this LGTM

ljharb:
I'm just planning on bumping the major version - but yeah a peerDep probably makes sense too.

PR #574, opened 2015-11-09
Updated wording and slight consistancy changes

dsgriffin:


README.md
```diff
@@ -232,7 +232,7 @@ Other Style Guides
   <a name="es6-object-concise"></a>
   - [3.6](#3.6) <a name='3.6'></a> Use property value shorthand.
 
-  > Why? It is shorter to write and descriptive.
+  > Why? It is shorter to write and autological.
```
ljharb: This is a confusing term - I don't see this as better. Can you elaborate?

README.md
```diff
@@ -250,7 +250,7 @@ Other Style Guides
 
   - [3.7](#3.7) <a name='3.7'></a> Group your shorthand properties at the beginning of your object declaration.
 
-  > Why? It's easier to tell which properties are using the shorthand.
+  > Why? It is easier to tell which properties are using the shorthand.
```
ljharb: Contractions are conversational and idiomatic, we'd prefer to keep them in.

README.md
dsgriffin: Autological = describes itself. Could use self-descriptive instead, I just don't see 'descriptive' as the right word here.

README.md
dsgriffin: It's because some places are using "Why? It's.." while other places are using "Why? It is.." and I thought making it consistent in one way would be best.

PR #575, opened 2015-11-10
Allow multiple stateless components in a single file

SimenB:
IDK if you want it, but we had linting errors after upgrading eslint-plugin-react.

This change allows multiple stateless components in a single file.

Requires 3.80 of eslint-plugin-react
https://github.com/yannickcr/eslint-plugin-react/blob/master/CHANGELOG.md#380---2015-11-07

### DIFF
--- a/packages/eslint-config-airbnb/rules/react.js
+++ b/packages/eslint-config-airbnb/rules/react.js
@@ -91,7 +91,7 @@ module.exports = {
     'react/no-is-mounted': 2,
     // Prevent multiple component definition per file
     // https://github.com/yannickcr/eslint-plugin-react/blob/master/docs/rules/no-multi-comp.md
-    'react/no-multi-comp': [2, {'ignoreStateless': false}],
+    'react/no-multi-comp': [2, {'ignoreStateless': true}],
     // Prevent usage of setState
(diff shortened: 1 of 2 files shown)


mkscrg:
:+1:

ljharb:
This relates to #612.

What is your rationale for not putting each component in its own module?

SimenB:
Mostly it's because the stateless functions makes it really easy to have lots of really tiny components, that might not be reusable. Having them in their own files seems overkill when they are only used in a single place. As we use CSS Modules as well, that'd mean an extra stylus file in addition, for something that might be 3 lines of code.

ljharb:
That seems reasonable to me. Let's get some more consensus (which might take a week or so, due to holidays) before proceeding.

goatslacker:
:+1: but we don't use stateless components yet so :neutral_face:

ljharb:
Sounds like this is OK to merge, it just won't do us any good at Airbnb just yet :-)

@SimenB can you rebase this freshly on top of latest master, and also add a note to this effect in the react styleguide?

SimenB:
Will do later today!

SimenB:
@ljharb Updated. I don't know what text you actually want though. Is this fine?

ljharb:
This looks great! perhaps a link inline to the eslint rule docs?

SimenB:
Like so?

PR #578, opened 2015-11-11
[eslint] Enforce rule 18.5

adrienverge:
Rule [18.5][1] states that files shall end with a single newline
character. Until now, this was not checked because there was no such
option in eslint.

Since version 1.8.0, eslint provides the ability to do that. See [pull
request on eslint][2] for details on implementation and usage. Let's use
it in airbnb eslint plugin now!

[1]: https://github.com/airbnb/javascript#18.5
[2]: https://github.com/eslint/eslint/pull/4266


ljharb:
LGTM :+1:

PR #579, opened 2015-11-12
Bring back note about braces

tomekwi:
https://github.com/airbnb/javascript/pull/472#discussion_r44334308

### DIFF
--- a/README.md
+++ b/README.md
@@ -743,7 +743,7 @@ Other Style Guides
     ```
 
 
-  - [8.4](#8.4) <a name='8.4'></a> If your function takes a single argument and doesn’t use braces, omit the parentheses.
+  - [8.4](#8.4) <a name='8.4'></a> If your function takes a single argument and doesn’t use braces, omit the parentheses. Otherwise, always include parentheses around arguments.
 
     > Why? Less visual clutter.
 
--- a/README.md
+++ b/README.md
@@ -699,7 +699,7 @@ Other Style Guides
     });
     ```
 
-  - [8.2](#8.2) <a name='8.2'></a> If the function body consists of a single expression, feel free to omit the braces and use the implicit return. Otherwise add the parentheses, braces, and use a `return` statement.
+  - [8.2](#8.2) <a name='8.2'></a> If the function body consists of a single expression, omit the braces and use the implicit return. Otherwise, keep the braces and use a `return` statement.
 
     > Why? Syntactic sugar. It reads well when multiple functions are chained together.
 
@@ -743,18 +743,30 @@ Other Style Guides
     ```
 
 
-  - [8.4](#8.4) <a name='8.4'></a> If your function only takes a single argument, feel free to omit the parentheses.
+  - [8.4](#8.4) <a name='8.4'></a> If your function takes a single argument and consists of a single expression, omit the parentheses.
 
     > Why? Less visual clutter.
 
     eslint rules: [`arrow-parens`](http://eslint.org/docs/rules/arrow-parens.html).
 
     ```js
+    // bad
+    [1, 2, 3].map((x) => x * x);
+
     // good
     [1, 2, 3].map(x => x * x);
 
+    // bad
+    [1, 2, 3].reduce(x => {
+      const y = x + 1;
+      return x * y;
+    });
+
     // good
-    [1, 2, 3].reduce((y, x) => x + y);
+    [1, 2, 3].reduce((x) => {
+      const y = x + 1;
+      return x * y;
+    });
     ```
 
 **[⬆ back to top](#table-of-contents)**
--- a/README.md
+++ b/README.md
@@ -743,7 +743,7 @@ Other Style Guides
     ```
 
 
-  - [8.4](#8.4) <a name='8.4'></a> If your function takes a single argument and consists of a single expression, omit the parentheses.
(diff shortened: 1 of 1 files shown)


tomekwi:
@goatslacker @ljhar, sorry guys!

README.md
```diff
@@ -622,7 +622,7 @@ Other Style Guides
     });
     ```
 
-  - [8.2](#8.2) <a name='8.2'></a> If the function body consists of a single expression, feel free to omit the braces and use the implicit return. Otherwise use a `return` statement.
+  - [8.2](#8.2) <a name='8.2'></a> If the function body consists of a single expression, feel free to omit the braces and use the implicit return. Otherwise add the parentheses, braces, and use a `return` statement.
```
ljharb: @goatslacker i think you also wanted this to be explicit about whether the parens around a single arrow function argument were required, prohibited, or either, and when - if the intention is that you must always use parens around arrow function args that have a body with braces, i think that should be explicitly stated.

tomekwi: > explicit about whether the parens around a single arrow function argument were required

I guess that’s stated in [8.4](https://github.com/airbnb/javascript/tree/60480c05ea6c2d0f172302f845c39f5f62accb8d#8.4). The rule originally came from you guys – I just [bumped the number](https://github.com/airbnb/javascript/pull/472#discussion_r36888047).

> and […] that you must always use parens around arrow function args that have a body with braces

Yeah, I noticed that intention in the original text – and I did my best to make it explicit through “add *the parentheses*, braces, and use a `return` statement”. Originally it was also formulated this way.

ljharb: That says "feel free to omit" - what I'm looking for here is a requirement - either with one arg, parebs must be there or they must not be, in each scenario (implicit return and explicit return).

tomekwi: 8.2 is about braces. 8.4 is about parentheses.

How about saying “This rule does not apply to functions with a multiline body wrapped in braces or parentheses. If your function spans over multiple lines, always wrap its arguments in parentheses.” in 8.2 then?

Then we can drop “feel free” and “add the parentheses” from this rule – and we’re done!

ljharb: If @goatslacker agrees, then that sounds good.

goatslacker: What we've agreed to:

* If you're using braces then the parentheses are always required.
* If it fits in a single line then feel free to omit anything (parentheses, braces, return).

goatslacker: We should also update the lint rule to reflect this.

ljharb: :thumb:

tomekwi:
:checkered_flag:

I’ve written down what you suggested. The rules have actually got simpler!

Sorry for the multiple commits. Let me know if I should rebase.

README.md
```diff
@@ -664,16 +664,34 @@ Other Style Guides
     ```
 
 
-  - [8.4](#8.4) <a name='8.4'></a> If your function only takes a single argument, feel free to omit the parentheses.
+  - [8.4](#8.4) <a name='8.4'></a> If your function takes a single argument and doesn’t use braces, omit the parentheses.
```
ljharb: let's add "Otherwise, always include parentheses around arguments." to communicate that it's not up to the individual developer.

ljharb:
@tomekwi One comment, plus a rebase, and I'd love to merge this!

Are there any eslint rules we could include/enable/link to that would enforce this?

tomekwi:
> One comment, plus a rebase

Done sir!

ljharb:
A followup PR that ensures the eslint rule matches this updated text would be awesome

tomekwi:
@ljharb I’ll happily do it – but we have an aweful lot of work this week! Is it OK if it waits until next week?

ljharb:
Of course!

tomekwi:
Hi @ljharb, I’ve looked through ESLint rules and found nothing that would need an update. Two relevant rules are [`arrow-parens`](http://eslint.org/docs/rules/arrow-parens) and [`arrow-body-style`](http://eslint.org/docs/rules/arrow-body-style). They are already configured and compatible with your decisions here.

Sadly, it doesn’t seem like ESLint can specify arrow function syntax that strongly.

ljharb:
Sounds like time for a PR into eslint itself! :-D

tomekwi:
Yeah, but I’m afraid it’s a task for someone who’s bitten his way through the source ESLint or has time to do it. You’d be waiting for ages if I were to PR that!

goatslacker:
What's missing from eslint exactly?

tomekwi:
```js
// bad
[1, 2, 3].map(number => 'As time went by, the string containing the ' +
  `${number} became much longer. So we needed to break it over multiple ` +
  'lines.'
);

// good
[1, 2, 3].map(number => (
  `As time went by, the string containing the ${number} became much ` +
  'longer. So we needed to break it over multiple lines.'
));
```

```js
// bad
[1, 2, 3].map((x) => x * x);

// good
[1, 2, 3].map(x => x * x);

// good
[1, 2, 3].map(number => (
  `A long string with the ${number}. It’s so long that we’ve broken it ` +
  'over multiple lines!'
));

// bad
[1, 2, 3].map(x => {
  const y = x + 1;
  return x * y;
});

// good
[1, 2, 3].map((x) => {
  const y = x + 1;
  return x * y;
});
```
(`arrow-parens` doesn’t cover the latter.)

PR #581, opened 2015-11-13
add missing rules from eslint-plugin-react (enforcing where necessary )

tikotzky:
Adds all missing rules from `eslint-plugin-react`.

Enforces [`react/jsx-closing-bracket-location`](https://github.com/yannickcr/eslint-plugin-react/blob/master/docs/rules/jsx-closing-bracket-location.md) and [`react/jsx-indent-props`](https://github.com/yannickcr/eslint-plugin-react/blob/master/docs/rules/jsx-indent-props.md) to match react style guide.

### DIFF
--- a/packages/eslint-config-airbnb/rules/react.js
+++ b/packages/eslint-config-airbnb/rules/react.js
@@ -5,10 +5,15 @@ module.exports = {
   'ecmaFeatures': {
     'jsx': true
   },
+  // View link below for react rules documentation
+  // https://github.com/yannickcr/eslint-plugin-react#list-of-supported-rules
   'rules': {
     // Prevent missing displayName in a React component definition
     // https://github.com/yannickcr/eslint-plugin-react/blob/master/docs/rules/display-name.md
     'react/display-name': 0,
+    // Forbid certain propTypes (any, array, object)
+    // https://github.com/yannickcr/eslint-plugin-react/blob/master/docs/rules/forbid-prop-types.md
+    'react/forbid-prop-types': [0, {'forbid': ['any', 'array', 'object']}],
     // Enforce boolean attributes notation in JSX
     // https://github.com/yannickcr/eslint-plugin-react/blob/master/docs/rules/jsx-boolean-value.md
     'react/jsx-boolean-value': [2, 'never'],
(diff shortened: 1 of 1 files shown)


packages/eslint-config-airbnb/rules/react.js
```diff
@@ -8,12 +8,22 @@ module.exports = {
   'rules': {
     // Prevent missing displayName in a React component definition
     'react/display-name': 0,
+    // Forbid certain propTypes (any, array, object)
+    'react/forbid-prop-types': 0,
```
ljharb: does this take any configuration? even if disabled, it's useful documentation to include the default options imo

packages/eslint-config-airbnb/rules/react.js
```diff
@@ -8,12 +8,22 @@ module.exports = {
...
     // Enforce boolean attributes notation in JSX
     'react/jsx-boolean-value': 2,
+    // Validate closing bracket location in JSX
+    'react/jsx-closing-bracket-location': 2,
```
ljharb: Can you link to documentation showing what this rule warns on and allows?

packages/eslint-config-airbnb/rules/react.js
```diff
@@ -8,12 +8,22 @@ module.exports = {
...
     // Enforce or disallow spaces inside of curly braces in JSX attributes
     'react/jsx-curly-spacing': 0,
+    // Validate props indentation in JSX
+    'react/jsx-indent-props': [2, 2],
```
ljharb: Can you link to documentation showing what this rule warns on and allows?

packages/eslint-config-airbnb/rules/react.js
tikotzky: Sorry, I linked in the issue description. I guess I should have made it clearer.
 [`react/jsx-closing-bracket-location`](https://github.com/yannickcr/eslint-plugin-react/blob/master/docs/rules/jsx-closing-bracket-location.md)

packages/eslint-config-airbnb/rules/react.js
tikotzky: yes it does take a config...
I'll add the default config to all that have a config

tikotzky: just to clarify would you want it to look like this?
```
'react/forbid-prop-types': [0, {forbid: ['any', 'array', 'object']}]
```

tikotzky: or would you rather me add a link pointing to the docs by each config, like this.
```
    // Forbid certain propTypes (any, array, object)
    // https://github.com/yannickcr/eslint-plugin-react/blob/master/docs/rules/forbid-prop-types.md
    'react/forbid-prop-types': 0,
```

ljharb: I think both would be ideal - the link if people want to click through, but the default config for minimal diffs if we want to override the defaults.

packages/eslint-config-airbnb/rules/react.js
ljharb: Thanks - please use a configuration of `{ "nonEmpty": "tag-aligned", "selfClosing": "tag-aligned" }` such that in a multiline component declaration, `/>` should always be on its own line, aligned with the opening tag.

packages/eslint-config-airbnb/rules/react.js
tikotzky: Existing disabled rules don't have the default config. 
Is that something you would want added everywhere?

One issue I see with that is that if the default config were to change we would out of date.

packages/eslint-config-airbnb/rules/react.js
tikotzky: tag-aligned would be the default. I guess we should add the config if we're adding a config to all rules.

packages/eslint-config-airbnb/rules/react.js
ljharb: Ideally yes - that's exactly the problem. I don't really care what the defaults are (now, or later) - I only care what style we want to encourage.

tikotzky: But what about rules that are disabled? What's the reason to include the default config there?

If someone is going to override they'll probably be looking at the docs anyway to see which config they would want to select when enabling.

ljharb: Because we may want to enable the rules later without having to look up the configuration from the docs :-)

tikotzky: I'll start by making sure each rule that is enabled specifies a config (if it has one)

I'll update the PR late tomorrow.

tikotzky:
@ljharb I added configs for all active rules.

I also added a link to the rule docs above all the rules, I started adding by each rule but it was looking pretty messy.

I'm gonna continue adding configs to the other disabled rules.
Please let me know if you have any feedback, Thanks.

packages/eslint-config-airbnb/rules/react.js
```diff
@@ -5,25 +5,46 @@ module.exports = {
...
+    'react/jsx-no-bind': 0,
     // Prevent duplicate props in JSX
-    'react/jsx-no-duplicate-props': 0,
+    'react/jsx-no-duplicate-props': [0, {'ignoreCase': false}],
+    // Prevent usage of unwrapped JSX strings
+    'react/jsx-no-literals': 0,
     // Disallow undeclared variables in JSX
     'react/jsx-no-undef': 2,
-    // Enforce quote style for JSX attributes
-    'react/jsx-quotes': 0,
```
ljharb: Why remove this rule? I think this is something we eventually want to enforce.

ljharb: oops, nevermind, i see your commit message

ljharb:
Thanks! I'd like a few more contributors to take a look before merging it, but this LGTM.

tikotzky:
ok great, let me know if you want me to change anything or squash.

ljharb:
@tikotzky could you rebase, and also add links to the rule configs in comments above the rules? Thanks!

ljharb:
@tikotzky oops, i just merged another react-related PR and there's conflicts again :-/ once more?

tikotzky:
ok rebased. now to add the links...

tikotzky:
I added links to all rules that were missing.

It seems like there has been some more rules added since this PR was originally created i'll go ahead and add those rules too.

tikotzky:
@ljharb should be good to go now.
Let me know if you want me to squash or make any other changes.

ljharb:
Sure, let's squash this down to one or two atomic commits, and let's merge it!

tikotzky:
ok squashed. should be good to go :smile:

PR #582, opened 2015-11-16
Use require.resolve to allow nested extend

jkrems:
By default eslint always resolves files in `extends` relative to the top-level project. Using `require.resolve` converts the references to absolute paths. Example:

```
node_modules/
  eslint-config-groupon/
    node_modules/
      eslint-config-airbnb/
        index.js # refers to 'eslint-config-airbnb/base'
```

By default eslint will look for `eslint-config-airbnb/base` in the top-level node_modules directory instead of one level down. By using `require.resolve` the "normal" npm resolution rules apply.

We use this approach in `eslint-config-groupon`. This change would allow us to get rid of [the uglier parts](https://github.com/groupon/javascript/blob/master/linters/eslint-config-groupon/legacy.js#L28-L48) in that repo.

### DIFF
--- a/packages/eslint-config-airbnb/base.js
+++ b/packages/eslint-config-airbnb/base.js
@@ -2,6 +2,6 @@ module.exports = {
   'extends': [
     'eslint-config-airbnb/legacy',
     'eslint-config-airbnb/rules/es6',
-  ],
+  ].map(require.resolve),
   'rules': {}
 };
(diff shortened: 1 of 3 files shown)


ljharb:
This seems like something eslint should be handling - `require.resolve` is something that generally shouldn't be done outside of bundling tools.

jkrems:
Reading the [existing discussion](https://github.com/eslint/eslint/issues/3458) doesn't make it sound like they want to address this anytime soon. But I understand why you'd be uncomfortable with this change.

justjake:
This seems legit to me -- do linting & tests still pass after this change?

jkrems:
The tests of this repo itself pass (e.g. running `npm test` in `packages/eslint-config-airbnb`).

joshhunt:
Keen to get this in :) This issue is preventing our internal config from upgrading to latest config.

PR #584, opened 2015-11-17
Avoid the word "mutate" while explaining `const`

jonathanp:
Since a common misconception about the `const` keyword in ES2015 is that it prohibits mutation of values when in fact it only prohibits mutation of references I think it’s clearer to avoid this word in this context completely. When we talk about mutability and immutability in JS we almost always mean _values_ and not _references_. So although the wording in this section is correct I think it should be re-phrased to avoid confusion. :)


ljharb:
LGTM

iamnirav:
nice

PR #188, opened 2014-08-18
Add section about Short Circuit Evaluation

vitorlopez:
Add section about Short Circuit Evaluation

PR #590, opened 2015-11-24
Rule: Do not use || for defaulting

adrienverge:
Defaulting omitted variables using `||` is like using `==` instead of `===`. If a function's argument is falsy (`0`, `false`, `null`, `''`...), it will be defaulted while it shouldn't be.

Example:

``` javascript
// bad
function f(arg, optionalArg, opts = {}) {
    const message = optionalArg || 'not set';  // what if arg === ''?
    const rate = opts.rate || 0.1;             // what if arg === 0.0?
}

// bad
const useUnicode = ENV['USE_UTF8'] || true;    // what if arg === false?

// good
function f(arg, optionalArg = 'not set', opts = {}) {
    const rate = opts.rate !== undefined ? opts.rate : 0.1;
}

// good
const useUnicode = ENV['USE_UTF8'] !== undefined : ENV['USE_UTF8'] || true;
```

More reading: http://www.codereadability.com/javascript-default-parameters-with-or-operator/

### DIFF
--- a/README.md
+++ b/README.md
@@ -537,7 +537,7 @@ Other Style Guides
     function handleThings(opts) {
       // No! We shouldn't mutate function arguments.
       // Double bad: if opts is falsy it'll be set to an object which may
-      // be what you want but it can introduce subtle bugs.
+      // be what you want but it can introduce subtle bugs (see rule 7.11).
       opts = opts || {};
       // ...
     }
@@ -596,6 +596,29 @@ Other Style Guides
 
   // still bad
   var subtract = Function('a', 'b', 'return a - b');
+  ```
+
+- [7.11](#7.11) <a name='7.11'></a> Avoid defaulting variables using `||`.
+
+  > Why? Because `||` does not check for strict equality (i.e. `===` vs. `==`). If a passed argument is falsy (`0`, `false`, `null`, `''`...), it will be defaulted while it shouldn't be.
+
+  ```javascript
+  // bad
+  function f(arg, optionalArg, opts = {}) {
+      const message = optionalArg || 'not set';  // what if arg === ''?
+      const rate = opts.rate || 0.1;             // what if arg === 0.0?
+  }
+
+  // bad
+  const useUnicode = ENV['USE_UTF8'] || true;    // what if arg === false?
+
+  // good
+  function f(arg, optionalArg = 'not set', opts = {}) {
+      const rate = opts.rate !== undefined ? opts.rate : 0.1;
+  }
+
+  // good
+  const useUnicode = ENV['USE_UTF8'] !== undefined : ENV['USE_UTF8'] || true;
   ```
 
 **[⬆ back to top](#table-of-contents)**


ljharb:
It's good to avoid the footgun of a falsy non-undefined value when using `||` - however, in the case of your example, where "port" needs to be a positive integer, `opts.port || 80` is actually more accurate than your "good" example, although nowhere near as precise as it should be.

Separately, is there an eslint rule that would enforce this?

goatslacker:
Does [this rule](https://github.com/airbnb/javascript#7.7) not handle this case?

ljharb:
@goatslacker that talks about reusing function argument names, which deopts in v8 and hurts readability, but this PR seems to be talking about use of `||` for defaulting any value, even to a new var.

adrienverge:
Thanks for feedback @ljharb @goatslacker.

> in the case of your example, where "port" needs to be a positive integer, opts.port || 80 is actually more accurate

That true, `port` was not a good a example. I've updated it.

> Does this rule not handle this case?

Rule 7.7 mentions that using `||` is a bad practice, but is about the default parameter syntax. The rule I want to add applies to any var in general (including `opts` arg: `opts.optionalKey`).

README.md
```diff
@@ -598,6 +598,23 @@ Other Style Guides
...
   ```
 
+- [7.11](#7.11) <a name='7.11'></a> Avoid defaulting variables using `||`.
+
+  > Why? Because `||` does not check for strict equality (i.e. `===` vs. `==`). If a passed argument is falsy (`0`, `false`, `null`, `''`...), it will be defaulted while it shouldn't be.
+
+  ```javascript
+  // bad
+  function f(arg, optionalArg, opts = {}) {
+      const message = optionalArg : 'not set';  // what if arg === ''?
```
goatslacker: this is a syntax error

README.md
```diff
@@ -598,6 +598,23 @@ Other Style Guides
...
+      const rate = opts.rate || 0.1;            // what if arg === 0.0?
+  }
+
+  // good
+  function f(arg, optionalArg = 'not set', opts = {}) {
+      const rate = opts.rate !== undefined ? opts.rate : 0.1;
```
ljharb: a "rate" that is numeric has the same problem as "port" before

README.md
adrienverge: My bad, sorry. I'm correcting it.

README.md
adrienverge: What do you mean?

Your point was that a `port` with value `0` is usually considered invalid, right? With `rate`, we can accept `0.0` or even `-2.34`. But maybe `opts.percent || 100.0` would be even better?

ljharb: i meant i could pass in `true` and it would accept it. but it's not that important

adrienverge: Oh, I see!

This is, in my opinion, an orthogonal problem. Checking input validity is up to the project error-handling guidelines (checking before calling *vs.* checking the arguments for example). But preserving input data (i.e. do not transform `0` into `'default'`) should be ensured in all cases.

adrienverge:
> Separately, is there an eslint rule that would enforce this?

I did not see any rule like that. It would be hard to enforce, since `||` is legitimate in certains cases (boolean operations).

ljharb:
It seems like an eslint rule could look for assignments where the right hand side contained a `||`? There very well may not be an existing rule though.

justjake:
@ljharb then we have to tell if the values involved in the right-hand side are nullable or something -- it could be possible to have this sort of check with Flow or some other null-aware type system but I don't think you could write a heuristic in ESLint that did anything but cause sadness

ljharb:
@justjake the idea would be to always disallow `||` in assignments, not just nullable values.

adrienverge:
> the idea would be to always disallow `||` in assignments, not just nullable values.

Actually what I proposed was less strict: disallow `||` for defaulting, i.e. do not use `||` for non-boolean values.

> eg `const shouldRun = isOnline() || (isOffline() && isIsomporphic())` should be allowed

I agree.

ljharb:
@adrienverge that means there's no way we could enforce a lint rule then, because there's no way to tell at compile time that `isOffline`, for example, is a boolean.

adrienverge:
@ljharb Asolutely. But in my opinion, not being able to enforce it does not mean we shouldn't recommend it.

jonmarkprice:
#188 generalizes this to include `&&`.

odykyi (approved):

PR #592, opened 2015-11-26
Enable Rest Params!

ariporad:
For some reason this is disabled. This enables it.

Thanks!


ljharb:
It's strange that this hasn't come up previously - LGTM

PR #599, opened 2015-12-02
Update README.md

camazorro:
Add Bisk Education to the list of organizations that are using this javascript guide


justjake:
This also renames the styleguide to Bisk

camazorro:
Yeah my bad

PR #602, opened 2015-12-02
Adding jeopardybot to styleguide users.

kesne:
This adds arguably the most important consumer of the glorious AirB&B Styleguide ever.


ljharb:
What is "merge"?

philihp:
So this list now includes products as well as organizations?

ljharb:
that's a fair point, but i think we can make an exception for this case

philihp:
I, for one, welcome our new Trebek overlords.

jimrhoskins:
Resolution is simple, incorporate JeopardyBot LLC.

PR #605, opened 2015-12-03
Require a space between `function` and `(`.

ljharb:
The guide was inconsistent in its use of spacing, and the most common convention both inside Airbnb and in community JS is to require this space. This both adds a section on it, and enables the linting rule.

### DIFF
--- a/README.md
+++ b/README.md
@@ -598,6 +598,20 @@ Other Style Guides
   var subtract = Function('a', 'b', 'return a - b');
   ```
 
+- [7.11](#7.11) <a name="7.11"></a> Spacing in a function signature.
+
+  > Why? Consistency is good, and you shouldn’t have to add or remove a space when adding or removing a name.
+  ```javascript
+  // bad
+  const f = function(){};
+  const g = function (){};
+  const h = function() {};
+
+  // good
+  const x = function () {};
+  const y = function a() {};
+  ```
+
 **[⬆ back to top](#table-of-contents)**
 
 ## Arrow Functions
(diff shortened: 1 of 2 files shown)


goatslacker:
Yes please :+1:

wyattdanger:
:+1:

jimrhoskins:
So much :+1:

README.md
```diff
@@ -598,6 +598,20 @@ Other Style Guides
...
+  > Why? Consistency is good, and you shouldn’t have to add or remove a space when adding or removing a name.
+  ```javascript
+  // bad
+  const f = function(){};
+  const g = function (){};
+  const h = function() {};
+
+  // good
+  const x = function () {};
+  const y = function a() {};
```
jimrhoskins: According to the guide (https://github.com/eslint/eslint/blob/master/docs/rules/space-before-function-paren.md) this should be `const y = function a () {}`

jimrhoskins:
I like this rule, but for a reason different than the explanation submitted. It reserves parentheses touching the word before them strictly for function invocations. Then other parenthesis uses are spaced (e.g. if/while/for and func definitions). It makes for a simple rule to remember "put a space before ( unless you are invoking a method".

To that end, this would also apply to method definitions and named functions. See the "always" examples for the rule https://github.com/eslint/eslint/blob/master/docs/rules/space-before-function-paren.md

README.md
ljharb: ah, i think i want `{"anonymous": "always", "named": "never"}` as a config then.

ljharb:
Sorry about that - I've fixed the eslint rules to match what I typed in the guide itself. There must never be a space between a function's name and the `(`, and there must always be precisely one space after `function`.

README.md
tleunen: Anyway there's another rule that says anonymous function should be named (or use arrow functions)

jimrhoskins:
Ah, darn. I like "always", but understand not everyone agrees :disappointed:

justjake:
rip

PR #606, opened 2015-12-03
Fix switch statement indentation rule

ljharb:
I abhor switch statements, but their indentation should be:

```js
switch (foo) {
  case 'bar':
    break;
}
```

not what the plugin requires by default:
```js
switch (foo) {
case: 'bar':
  break;
}
```


goatslacker:
:+1:

PR #613, opened 2015-12-07
Don't enforce spaces inside curly braces for nested objects/arrays

freesteph:
I couldn't find any examples around the project but I have a feeling this syntax is pretty standard across the web.

```javascript
var obj = { "foo": { "baz": 1, "bar": 2 }};
// or
var obj = { "foo": [ "baz", "bar" ]};
```
instead of 
```javascript
var obj = { "foo": { "baz": 1, "bar": 2 } };
var obj = { "foo": [ "baz", "bar" ] };
```
http://eslint.org/docs/rules/object-curly-spacing.html#exceptions

### DIFF
--- a/packages/eslint-config-airbnb/rules/style.js
+++ b/packages/eslint-config-airbnb/rules/style.js
@@ -67,8 +67,8 @@ module.exports = {
     'no-underscore-dangle': 0,
     // disallow the use of Boolean literals in conditional expressions
     'no-unneeded-ternary': 0,
-    // require padding inside curly braces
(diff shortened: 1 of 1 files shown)


ljharb:
That would end up with asymmetric spacing inside the outer curly braces, which could also be a refactoring annoyance if one added a key at the end whose value wasn't an object or an array. The current setting is what we use and would like to enforce.

Thanks for your contribution!

freesteph:
My bad, thanks for the feedback!

PR #615, opened 2015-12-09
Fix syntax highlighting in section 15

winterbe:
This is a fix for broken code blocks in section 15:

bildschirmfoto 2015-12-09 um 10 53 25


ljharb:
LGTM - thanks!

PR #619, opened 2015-12-14
Minor cleanup for the React styleguide

kesne:
This fixes some minor things that I thought were missing from the styleguide.

If people are unhappy about enforcing names for event handlers or additional render methods, I'm happy to discuss/remove.

---

- Obey our own styleguide for modules.
- Added missing rule for `true` props.
- New rule on event handler binding.

### DIFF
--- a/react/README.md
+++ b/react/README.md
@@ -323,51 +323,6 @@
     }
     ```
 
-  - Prefix the name of event handlers with `on`.
-    ```javascript
-    // bad
-    class extends React.Component {
-      handleClickDiv() {
-        // do stuff
-      }
-
-      // other stuff
-    }
-
-    // good
-    class extends React.Component {
-      onClickDiv() {
-        // do stuff
-      }
-
-      // other stuff
-    }
-    ```
-
-  - Prefix the name of additional render methods with `render`.
-
-    ```javascript
-    // bad
-    class extends React.Component {
-      createNavigation() {
-        // render stuff
-      }
-      render() {
-        {this.createNavigation()}
-      }
-    }
-
-    // good
-    class extends React.Component {
-      renderNavigation() {
-        // render stuff
-      }
-      render() {
-        {this.renderNavigation()}
-      }
-    }
-    ```
-
 ## Ordering
 
   - Ordering for `class extends React.Component`:
--- a/react/README.md
+++ b/react/README.md
@@ -56,10 +56,10 @@
 
     ```javascript
     // bad
-    const reservationCard = require('./ReservationCard');
+    import reservationCard from './ReservationCard';
 
     // good
-    const ReservationCard = require('./ReservationCard');
+    import ReservationCard from './ReservationCard';
 
     // bad
     const ReservationItem = <ReservationCard />;
@@ -72,13 +72,13 @@
 
     ```javascript
     // bad
-    const Footer = require('./Footer/Footer.jsx')
+    import Footer from './Footer/Footer.jsx';
 
     // bad
-    const Footer = require('./Footer/index.jsx')
+    import Footer from './Footer/index.jsx';
 
     // good
-    const Footer = require('./Footer')
+    import Footer from './Footer';
     ```
 
 ## Declaration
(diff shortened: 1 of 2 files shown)


react/README.md
```diff
@@ -69,13 +69,13 @@
     **Component Naming**: Use the filename as the component name. For example, `ReservationCard.jsx` should have a reference name of `ReservationCard`. However, for root components of a directory, use `index.jsx` as the filename and use the directory name as the component name:
     ```javascript
     // bad
-    const Footer = require('./Footer/Footer.jsx')
+    import Footer from './Footer/Footer.jsx';
```
ljharb: can we omit the extension in all our examples? file extensions should ideally never be specified in a `require`

react/README.md
```diff
@@ -237,16 +253,52 @@
     ```
 
 ## Methods
+  - Bind event handlers for the render method in the constructor.
+
+  > Why? A bind call in a prop will create a brand new function on every single render.
```
ljharb: what do you mean by "in a prop"? this example is "in the render path"

react/README.md
```diff
@@ -237,16 +253,52 @@
...
+      render() {
+        return <div onClick={this.onClickDiv} />
+      }
+    }
+    ```
+
   - Do not use underscore prefix for internal methods of a React component.
     ```javascript
     // bad
-    React.createClass({
```
ljharb: this syntax is intentionally included in the "bad" section - ie, it's not just illustrating the underscore, it's also reinforcing that one should use "class" syntax

react/README.md
```diff
@@ -255,7 +307,52 @@
...
+        // render stuff
+      }
+      render() {
+        {this.createNavigation()}
+      }
+    }
+
+    // good
+    class extends React.Component {
+      renderNavigation(){
```
ljharb: "good" requires a space between every ")" and "{", in all situations

kesne: Good catch!

react/README.md
kesne: Ah, okay. I thought that would be covered in the "use classes" section but I'm happy to leave to reinforce.

react/README.md
kesne: Whoops, will fix.

react/README.md
koistya: @ljharb :+1:

koistya:
```jsx
<button onClick={handleClick}>Click me</button>
           │          │
           │          └─ (verb) component's action
           └─ (noun) component's property
```

In many cases the even handlers (actions) should be named so that it's easy to understand what a particular function is doing. For example:

```jsx
<button onClick={pokePlayer}>Poke Player</button>
           │          │
           │          └─ (verb) a function definition performing a specific action
           └─ (noun) a property holding a reference to the component's action(s)
```
           
In both these examples using `on` prefix for action names (as this PR suggests) doesn't make any sence to me.

ljharb:
That is a pretty convincing counterpoint.

kesne:
The one issue with that is it makes the `sort-comp` rule pretty useless, assuming those methods exist on the component.

ljharb:
That is true. However, I'm not convinced we want to retain that rule (obv it'll need further discussion)

kesne:
I'm happy to drop that from this PR and leave it for another conversation.

ljharb:
@kesne yeah let's split the naming section into a separate PR, and keep the rule fix and the rest of the guide cleanups?

Charpeni:
I think we should use arrow functions instead of `.bind(this)`, it's more ES6.

```javascript
// bad
  class extends React.Component {
    constructor(props) {
      super(props);

      this.onClickDiv = this.onClickDiv.bind(this);
    }

    onClickDiv() {
      // do stuff
    }

    render() {
      return <div onClick={this.onClickDiv} />
    }
  }

// good
  class extends React.Component {
    constructor(props) {
      super(props);
    }

    onClickDiv = () => {
      // do stuff
    }

    render() {
      return <div onClick={this.onClickDiv} />
    }
  }

ljharb:
@Charpeni That would require the class properties proposal, which we don't yet use at Airbnb because it's too far from being finalized.

goatslacker:
> it's more ES6.

It's actually not even ES7 :sob:

Charpeni:
@lijharb I did not noticed that class properties was only a proposal, thank you for the clarification.

ljharb:
LGTM pending one final rebase. (This is a breaking change, fwiw)

kesne:
@ljharb Looks rebased to me?

ljharb:
do another fetch. it's currently on top of 4adf9303c2de35d189d359ce50f0d764b0e8d2ec, master is at 8815205471efaf071bebfaa7eced95c11aea1095

PR #618, opened 2015-12-12
[react] Including missing defaults to the react eslint

kesne:
We were missing including the defaults in a couple react rules for the eslint configuration. This adds them back.

PR #622, opened 2015-12-15
Remove deprecated react/jsx-quotes

billyjanitsch:
This rule is deprecated, but was reintroduced in #618. Running eslint with this config now prints:

```
The react/jsx-quotes rule is deprecated. Please use the jsx-quotes rule instead.
```

The correct rule is already enforced [here](https://github.com/airbnb/javascript/blob/master/packages/eslint-config-airbnb/rules/style.js#L29).

### DIFF
--- a/packages/eslint-config-airbnb/rules/react.js
+++ b/packages/eslint-config-airbnb/rules/react.js
@@ -27,9 +27,6 @@ module.exports = {
     // Disallow undeclared variables in JSX
     // https://github.com/yannickcr/eslint-plugin-react/blob/master/docs/rules/jsx-no-undef.md
     'react/jsx-no-undef': 2,
-    // Enforce quote style for JSX attributes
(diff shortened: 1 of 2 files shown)


ljharb:
oops, good call, totally forgot about that. thanks!

kesne:
That should probably exist in the `react` file, no?

ljharb:
it's not from the react plugin, it's in core eslint - i suppose it could be moved there.

kesne:
Yeah, but the way I see it the jsx rules don't make sense to apply outside of our `react` ruleset.

ljharb:
that's a fair point. a PR to move it over would be just fine, and wouldn't impact the module at all

manmadareddy:
abcdedd - Ysignature 

    On Tuesday, December 15, 2015 1:38 PM, Jordan Harband <notifications@github.com> wrote:
 

 that's a fair point. a PR to move it over would be just fine, and wouldn't impact the module at all—

PR #631, opened 2015-12-21
Method chaining with prototype

Fakerr:
Example provided in 9.3  looks like we can not use method chaining with prototype. It's actually possible, so we should put the two different ways.(oh, and I added a function constructor).

### DIFF
--- a/README.md
+++ b/README.md
@@ -807,18 +807,21 @@ Other Style Guides
 
     ```javascript
     // bad
+    function Jedi() {}
+
     Jedi.prototype.jump = function () {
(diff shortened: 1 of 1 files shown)


ljharb:
What you've done is modified the "bad" version to show proper chaining, when the intention is that the bad version  does *not* chain.

In addition, our style guide recommends never using the `.prototype.foo` approach now that "class" exists, so we wouldn't want the former to be in an example of a "good" pattern.

Thanks for your suggestion! Feel free to fork the style guide if you'd like to follow different practices than we do.

PR #632, opened 2015-12-22
[eslint config] [breaking] enable `quote-props` rule.

ljharb:
This is the rule we follow at Airbnb, but the styleguide and linter plugin haven't been enforcing it.

### DIFF
--- a/packages/eslint-config-airbnb/rules/style.js
+++ b/packages/eslint-config-airbnb/rules/style.js
@@ -85,7 +85,8 @@ module.exports = {
     // enforce padding within blocks
     'padded-blocks': [2, 'never'],
     // require quotes around object literal property names
-    'quote-props': 0,
+    // http://eslint.org/docs/rules/quote-props.html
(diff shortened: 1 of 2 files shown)


packages/eslint-config-airbnb/rules/style.js
```diff
@@ -85,7 +85,8 @@ module.exports = {
     // enforce padding within blocks
     'padded-blocks': [2, 'never'],
     // require quotes around object literal property names
-    'quote-props': 0,
+    // http://eslint.org/docs/rules/quote-props.html
+    'quote-props': [2, 'as-needed', { 'keywords': true, 'unnecessary': true, 'numbers': false }],
```
w-site: what about 'new' keyword ?

ljharb: Since we recommend using babel, which takes care of quoting reserved property names in legacy engines, this is not a concern.

PR #634, opened 2015-12-22
Sort react/prefer-es6-class alphabetically

lencioni:
All of the other rules in this file are in alphabetical order, but this
one was added by c98990c0f out of order at the end. Keeping these in
alphabetical order will help developers find the rules that they are
looking for.


ljharb:
Thanks!

PR #636, opened 2015-12-23
React readme cleanup

lencioni:
While I was poking around in this file today, I noticed a few very minor things that could use some attention. These commits mostly just adjust some of the formatting--some for the benefit of those reading the rendered markdown, and others for the benefit of those looking at the markdown source. There's likely still more that could be done, but I figured I'd throw this up here since I think it makes a few small improvements.


ljharb:
This all seems good to me

PR #637, opened 2015-12-23
Permit strict mode

evenfrost:
Permit strict mode for legacy configuration as this mode is not intended to be used with Babel.


ljharb:
If it's removed from `legacy` it'd need to be re-added to https://github.com/airbnb/javascript/blob/master/packages/eslint-config-airbnb/index.js#L2-L5 to prevent it from being removed from the default export.

Issue #509, opened 2015-09-16
No max-len ESLint rule?


screendriver:
Is there a reason why there is no [max-len](http://eslint.org/docs/rules/max-len) rule present in your [rules](https://github.com/airbnb/javascript/blob/master/packages/eslint-config-airbnb/base/index.js)?

justjake:
Nope! Submit a PR and I'll merge it.

screendriver:
With pleasure :sunglasses:

pitaj:
Should this be closed since #639 defined it?

ljharb:
Thanks, closed with #639.

PR #639, opened 2015-12-24
Define a max line length of 100 characters

chrisngobanh:
Taking over #510 
Addresses #509 

This PR adds section 18.12, which defines a limit for the length of a line to be **100 characters (including whitespace)**. This number was agreed upon in #510 and is also the line length defined in the Ruby style guide.

I enabled ESLint rule [max-len](http://eslint.org/docs/rules/max-len.html) to enforce this, as well as updated `.jshintrc`.

I also updated section 6.2 to recognize the newly defined line length.

Something to discuss: I put this new rule in the Whitespace section (18), but should it have its own section? 
The Ruby style guide has it in its section, separate from the Whitespace section.

### DIFF
--- a/packages/eslint-config-airbnb/rules/style.js
+++ b/packages/eslint-config-airbnb/rules/style.js
@@ -34,6 +34,8 @@ module.exports = {
     'lines-around-comment': 0,
     // disallow mixed 'LF' and 'CRLF' as linebreaks
     'linebreak-style': 0,
+    // specify the maximum length of a line in your program
+    "max-len": [2, 100, 4],
     // specify the maximum depth callbacks can be nested
     'max-nested-callbacks': 0,
     // require a capital letter for constructors
--- a/packages/eslint-config-airbnb/rules/style.js
+++ b/packages/eslint-config-airbnb/rules/style.js
@@ -35,7 +35,17 @@ module.exports = {
     // disallow mixed 'LF' and 'CRLF' as linebreaks
     'linebreak-style': 0,
     // specify the maximum length of a line in your program
-    "max-len": [2, 100, 4],
+    // https://github.com/eslint/eslint/blob/master/docs/rules/max-len.md
+    'max-len': [2, {
+      'code': 100,
+      'comments': 100,
+      'commentLength': 100,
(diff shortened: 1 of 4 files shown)


packages/eslint-config-airbnb/rules/style.js
```diff
@@ -34,6 +34,8 @@ module.exports = {
     'lines-around-comment': 0,
     // disallow mixed 'LF' and 'CRLF' as linebreaks
     'linebreak-style': 0,
+    // specify the maximum length of a line in your program
```
ljharb: can you add https://github.com/eslint/eslint/blob/master/docs/rules/max-len.md in a comment?

packages/eslint-config-airbnb/rules/style.js
```diff
@@ -34,6 +34,8 @@ module.exports = {
...
+    "max-len": [2, 100, 4],
```
ljharb: re the third option: since we use 2-space indentation, we should probably count this as `2` (even though we don't use tabs).

Also, can we use single quotes, and the object form, and include all defaults? For example:
```
'max-len': [2, {
  "code": 100,
  "comments": 100,
  "commentLength": 100,
  "tabWidth": 2,
  "ignoreUrls": false,
  "ignorePattern": null,
  "ignoreTrailingComments": false,
  "ignoreComments": false
}],
```

ljharb:
This LGTM, pending comments. I'd be fine with it either in its own section, or in the whitespace section.

chrisngobanh:
@ljharb done! I also rebased

gergoerdosi:
This rule definition results in an error:

```
Configuration for rule "max-len" is invalid:
Value "[object Object]" is the wrong type.
```

The current definition is valid only in eslint@2.x.x, but the package has a dependency on eslint@1.10.x which requires a different format.

ljharb:
Oops! 2.x isn't released yet, so we should fix that. Luckily, it hasn't yet been released in `eslint-config-airbnb`. @gergoerdosi, would you mind submitting a PR to fix that? It'd be ideal to also preserve the object form in a comment noting that it should be used with 2.x.

Issue #641, opened 2015-12-29
Thoughts on "Never mutate parameters"


AlicanC:
1- This [rule](https://github.com/airbnb/javascript#7.12)'s "Do not reassign parameters" part should be split into another rule, because reassigning a parameter isn't actually mutation.

Plus it sounds like `no-param-reassign` rule also covers mutation, but it doesn't. It would be nice to easily see what gets covered by ESLint and what doesn't.

---

2- Why shouldn't we reassign parameters?

> Overwriting parameters can lead to unexpected behavior, especially when accessing the arguments object.

Accessing `arguments` is already discouraged. Should we really care about breaking something we aren't supposed to use? What are other unexpected behaviors can this cause?

sandyjoshi:
2- In javascript objects are passed by reference, so when you pass object to any function and modify that object inside that function, it means the reference of object which is passed into a function will get affected. I think it is main "unexpected behaviors".

AlicanC:
Reassigning parameters doesn't do that. Mutating parameters do that.

```javascript
const myObj = {
  adele: 'hello',
};

function myFunc(myParam) {
  myParam = {};

  myParam.adele = 'rolling in the deep';
}

myFunc(myObj);

console.log(myObj); // { adele: 'hello' }
```

ljharb:
Reassigning parameters deoptimizes in many engines, specifically v8 - it's a horrible idea to do it. Variables are free, and creating new ones rather than reusing old ones makes code much clearer.

(In addition, nothing in JS is passed by reference, everything is passed by value, where objects are a form of "reference value" - [this](http://www.jon-carlos.com/2013/is-javascript-call-by-value-or-call-by-reference/) is a good read on the subject)

A PR would be welcome if there's verbage that could be cleared up.

carpeliam:
Not sure if this is the place for additional thoughts, but I'm a little unsure as to how I should handle something like `array.forEach((obj) => { obj.prop = true; });`. Other than that, I'm a fan.

ljharb:
@carpeliam by copying instead of mutating: `array.map(obj => Object.assign({}, obj, { prop: true }))`

noahleigh:
@ljharb I'm not sure that solution works in the case of setting the style attributes of a list of DOM nodes:
```javascript
nodeArray.forEach(node => (node.style.display = 'none'));
```
What do you do when you _have_ to mutate parameters in-place?

ljharb:
[...]

PR #643, opened 2015-12-29
Make reassigning a separate rule.

AlicanC:
Made the change discussed in #641.

### DIFF
--- a/README.md
+++ b/README.md
@@ -631,7 +631,23 @@ Other Style Guides
 
 - [7.12](#7.12) <a name="7.12"></a> Never mutate parameters.
 
-  > Why? Overwriting parameters can lead to unexpected behavior, especially when accessing the `arguments` object. Manipulating objects passed in as parameters can cause unwanted variable side effects in the original caller.
+  > Why? Manipulating objects passed in as parameters can cause unwanted variable side effects in the original caller.
+
+  ```javascript
+  // bad
+  function f1(obj) {
+    obj.key = 1;
+  };
+
+  // good
+  function f2(obj) {
+    const key = Object.prototype.hasOwnProperty.call(obj, 'key') ? obj.key : 1;
+  };
+  ```
+
+- [7.13](#7.13) <a name="7.13"></a> Never reassign parameters.
+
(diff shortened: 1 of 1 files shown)


ljharb:
I'm not sure why GH is unable to show the diff here :-/ However, it looks good on my machine.

AlicanC:
I have said that `no-param-reassign` has nothing to do with mutation, but it looks like it does:
https://github.com/eslint/eslint/blob/master/docs/rules/no-param-reassign.md

Setting `props` to `true` should also prevent parameter mutation and it's enabled in Airbnb config:
https://github.com/airbnb/javascript/blob/master/packages/eslint-config-airbnb/rules/best-practices.js#L79

The weird thing is, ESLint doesn't warn me when I do this:
```javascript
function f(obj) {
  obj.key = 1;
}

f({ key: 1 });
```

Am I misreading the rule? What is wrong here?

ljharb:
hm, it should give you an error when doing that. that might be an `eslint` bug?

AlicanC:
Hmm, the change was actually made in https://github.com/airbnb/javascript/commit/2589c67b0c392a118aba3e3938856ddfead26baa which is yet to be released. It should work when it's released.

---

I think putting these under separate rules is still a good idea. I could add the same `eslint rules:` indication to "Never mutate parameters." since it will also covered by that rule on the next release.

If having two Airbnb rules pointing to the same ESLint rule would be weird, I could merge the rules again under "Never reassign or mutate parameters." and keep the "It can also cause optimization issues, especially in V8." part.

ljharb:
I like having a separate section for `no-param-reassign`. Multiple rules in one guide section is fine, but ideally each rule is only in one guide section.

AlicanC:
So we merge this one and we are done?
https://github.com/AlicanC/javascript/commit/a5030475badd272e0ca3d3562fa18d8f14194ad8

It looks like this PR won't update itself since it's merged. I'm supposed to create another PR right?

ljharb:
yes, you'd need to make a new PR once the original is merged, on a new branch.

PR #644, opened 2015-12-29
update package

rajansubedi32:
homepage to Homepage

PR #645, opened 2015-12-29
update package

rajansubedi32:
homepage to Homepage


ljharb:
@rajansubedi32 this is a duplicate of #644, which was closed for the same reason: "as with all fields in package.json, this field must be lowercased for npm to understand it properly."

It's great that you're learning to use Github! I'd recommend creating your own repository to play with, rather than creating noise for an existing project.

PR #649, opened 2015-12-29
Add ESLint link for "Never mutate parameters." rule.

AlicanC:
Continuing #643.


goatslacker:
Are we enforcing these in our own eslint config? We should write a test script that checks our rules vs eslint config.

PR #651, opened 2015-12-31
Fix max-len rule definition

gergoerdosi:
The current definition is valid only in eslint@2.x.x, but this package has a dependency on eslint@1.10.3 which doesn't have all those options yet.


ljharb:
per https://github.com/airbnb/javascript/pull/639#issuecomment-168099321

ljharb:
LGTM, thanks

PR #656, opened 2016-01-04
assign variables: give a reason for the assignment

eddiemonge:
I know the examples are contrived but they should at least be best practices in themselves. The variable assignment wasn't necessary. The function could have passed the getter result directly. 

https://github.com/airbnb/javascript#13.4

This wasn't necessary:
```
const name = getName();
this.setFirstName(name);
```
since it could have been written like this:
```
this.setFirstName(getName());
```

This change gives a reason to have it there.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1102,45 +1102,36 @@ Other Style Guides
     > Why? `let` and `const` are block scoped and not function scoped.
 
     ```javascript
-    // good
-    function () {
-      test();
-      console.log('doing stuff..');
-
-      //..other stuff..
-
+    // bad - unnecessary function call
+    function checkName(hasName) {
       const name = getName();
 
+      if (hasName === 'test') {
+        return false;
+      }
+
       if (name === 'test') {
+        this.setName('');
         return false;
       }
 
       return name;
     }
 
-    // bad - unnecessary function call
-    function (hasName) {
-      const name = getName();
-
-      if (!hasName) {
-        return false;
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -1136,6 +1139,9 @@ Other Style Guides
       }
 
       const name = getName();
+      if (name.indexOf(' ') > -1) {
+        name = name.split(' ')[ 0 ];
```
ljharb: this isn't good, since it's reassigning a `const` - it will throw.

eddiemonge: oops you are right. Should be `lets`

ljharb: (also, there shouldn't be spaces inside brackets: `[0]`)

eddiemonge:
removed the first example as it seemed like a duplicate

ljharb:
LGTM, just needs a fresh rebase :-) thanks!

eddiemonge:
rebased. not sure why I needed to but did it anyway.

ljharb:
Thanks! If it's on top of latest master I can merge it in locally without a merge commit.

eddiemonge:
If you are doing it locally then its trivial to also do the rebase there as well. Not saying there isn't a reason for the process but it makes it harder to for people trying to help. How often should I be rebasing, every time there is a new commit to master, every day, etc,? A submitter should only rebase if there are conflicts or to squash imo.

PR #658, opened 2016-01-06
Fixed typo in URL

foaly-nr1:


ljharb:
Good catch, this should have been done as part of #421

PR #661, opened 2016-01-06
Use Array.from to convert arguments to array

silvenon:
You recommended that technique before. Or is that hurting the point? Is `prototype.slice.call` intentionally used to make the bad example worse?

### DIFF
--- a/README.md
+++ b/README.md
@@ -558,7 +558,7 @@ Other Style Guides
     ```javascript
     // bad
     function concatenateAll() {
-      const args = Array.prototype.slice.call(arguments);
+      const args = Array.from(arguments);
       return args.join('');
     }
 


geekjuice:
Might be a tangential argument, but [`Array#from`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/from) doesn't seem to have very good support yet, so using `Array::slice` is safer. With that in mind, I would also argue [4.4](https://github.com/airbnb/javascript#4.4) should be updated to `[...foo]`.

ljharb:
The style guide is written assuming that the [es5-shim](https://github.com/es-shims/es5-shim) and [es6-shim](https://github.com/paulmillr/es6-shim) are being used, so availability of the API isn't a problem.

However, this particular example is meant to convey how an ES5 pattern can be improved with an ES6 pattern - `Array.from`, I think, would lessen that impact.

geekjuice:
Ah, didn't realize those shims were used. Carry on :rocket:

silvenon:
Ok :+1:

PR #662, opened 2016-01-06
Use even shorter syntax for exporting from modules

silvenon:


silvenon:
Didn't read below :sweat_smile:

PR #664, opened 2016-01-07
Ignore URLs in max-len

silvenon:
How do you get around this situation?

```js
// https://github.com/webpack/css-loader/tree/1298d2b38c4770dbf853ff1eed632fe239881cc2#css-modules
const foo = 'bar';
```

I could use a URL shortener, but that would remove the information from the URL.


ljharb:
This seems legit. LGTM. I'll merge if other contributors agree.

kesne:
I'm not a real contributor but this looks great to me.

silvenon:
Awesome :sparkles:

PR #667, opened 2016-01-08
Disallow declaration of global variables

foray1010:


packages/eslint-config-airbnb/rules/variables.js
```diff
@@ -11,7 +11,7 @@ module.exports = {
     // disallow shadowing of names such as arguments
     'no-shadow-restricted-names': 2,
     // disallow declaration of variables already declared in the outer scope
```
ljharb: would you mind adding a new comment line linking to the rule definition (http://eslint.org/docs/rules/no-shadow.html)?

packages/eslint-config-airbnb/rules/variables.js
```diff
@@ -11,7 +11,7 @@ module.exports = {
...
-    'no-shadow': 2,
+    'no-shadow': [2, {'builtinGlobals': true}],
```
ljharb: please flesh out all the config options here - ie, `hoist`, and `allow`, with their defaults.

ljharb:
This might also be nice to reference somewhere in the actual guide. LGTM overall, pending 2 comments.

foray1010:
Updated according to the comments

foray1010:
enabling this will disable the use of a lot of variables when env `browser` is true, such as `name` and `length`
See https://github.com/sindresorhus/globals/blob/master/globals.json#L164

My team disable `browser` because most of our projects are backend
but for airbnb, I think you guys may have different preference
Feel free to close this PR

ljharb:
I don't want the JS builtins shadowed, but that's a good point - given that it includes all environment builtins, perhaps this would be a bit too restrictive.

PR #668, opened 2016-01-08
Remove whitespace before function call parens

sahat:
Having a space here is a bit distracting for 2 reasons:

1. This rule is about using soft tabs over tabs
2.  `// good` goes against violation of [18.3](https://github.com/airbnb/javascript#18.3) that says "*Place no space before the argument list in function calls and declarations.*"

### DIFF
--- a/README.md
+++ b/README.md
@@ -1457,17 +1457,17 @@ Other Style Guides
 
     ```javascript
     // bad
-    function () {
+    function() {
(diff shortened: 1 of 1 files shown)


ljharb:
There must **always** be a space after the `function` keyword, which is addressed by [7.11](https://github.com/airbnb/javascript#7.11).

I've tweaked 18.3 to make this clear.

Issue #671, opened 2016-01-11
Add section on nested/multiline ternaries


ljharb:
Specifically, prohibiting nested ternaries (to match https://github.com/airbnb/javascript/blob/master/packages/eslint-config-airbnb/rules/style.js#L63-L64) and perhaps discouraging multiline ternaries (not sure if there's a rule for it).

bgvianyc:
opened a PR: https://github.com/airbnb/javascript/pull/673

ljharb:
Done in #673.

PR #673, opened 2016-01-11
Add section on nested/multiline ternaries

bgvianyc:
As per this issue: https://github.com/airbnb/javascript/issues/671

### DIFF
--- a/README.md
+++ b/README.md
@@ -1246,6 +1246,29 @@ Other Style Guides
 
   - [15.4](#15.4) <a name='15.4'></a> For more information see [Truth Equality and JavaScript](http://javascriptweblog.wordpress.com/2011/02/07/truth-equality-and-javascript/#more-2108) by Angus Croll.
 
+  - [15.5](#15.5) <a name='15.5'></a> Ternaries should not be nested and generally be single line expressions.
+ 
+    eslint rules: [`no-nested-ternary`](http://eslint.org/docs/rules/no-nested-ternary.html).
+    
+    ```javascript
+    //bad
+    const foo = maybe1 > maybe2
+      ? "bar"
+      : value1 > value2 ? "baz" : null;
+     
+    //better
+    const maybeNull = value1 > value2 ? 'baz' : null;
+    
+    const foo = maybe1 > maybe2
+      ? 'bar'
+      : maybeNull;
+    
+    //best
+    const maybeNull = value1 > value2 ? 'baz' : null;
+    
+    const foo = maybe1 > maybe2 ? 'bar' : maybeNull;
+    ```
+    
 **[⬆ back to top](#table-of-contents)**
 
 


ljharb:
This could use a "bad" example of a nested ternary - also, would you ming rebasing these down to perhaps a single commit?

README.md
```diff
@@ -1289,6 +1289,27 @@ Other Style Guides
...
+  - [15.5](#15.5) <a name='15.5'></a> Ternaries should not be nested and generally be single line expressions.
+ 
+    ```javascript
+    //bad
+    const foo = maybe1 > maybe2
+      ? "bar"
+      : value1 > value2 ? "baz" : null;
+     
+    //better
+    const maybeNull = value1 > value2 ? "baz" : null;
```
ljharb: per our "quotes" rule, all the non-bad ones should use single quotes

README.md
```diff
@@ -1289,6 +1289,27 @@ Other Style Guides
 
   - [15.4](#15.4) <a name='15.4'></a> For more information see [Truth Equality and JavaScript](http://javascriptweblog.wordpress.com/2011/02/07/truth-equality-and-javascript/#more-2108) by Angus Croll.
 
+  - [15.5](#15.5) <a name='15.5'></a> Ternaries should not be nested and generally be single line expressions.
```
kesne: The wording here is a little weak. Are there cases when ternaries should not be single-line expressions? If so, what are they?

ljharb: The only case that's been brought up is when you want to use `const` but have a multi-part condition that you don't want to evaluate unless you have to - it gets really ugly using a `let` and an `if` block, or breaking up the condition.

That said, I still prefer only ever using single-line ternaries.

bgvianyc: I think the general rule for a multi-line ternary is if it would be a long single-line

README.md
bgvianyc: done

bgvianyc:
rebased

bgvianyc:
In this node style guide they encourage multi-line ternaries, although no explanation is given why.

https://github.com/felixge/node-style-guide#use-multi-line-ternary-operator

ljharb:
Can we make this link to the relevant eslint rule(s)?

bgvianyc:
linked to one rule, did you have others in mind?

ljharb:
I think that's it :-) can you rebase this down to 1 commit on latest master? I'll merge it ASAP

README.md
DerFlatulator: I don't think there's anything wrong with multi-line ternary statements. 
```js
const someValue = this.hasSomeThing() 
    ? this.someThing(param)
    : this.otherThing(param);
```
Is clearer than
```js
const someValue = this.hasSomeThing() ? this.someThing(param) : this.otherThing(param);
```

ljharb: That's definitely subjective - I find the latter far clearer.

bgvianyc:
rebased

README.md
bgvianyc: Since it's subjective, should the section just be called `Ternaries should not be nested.` ?

DerFlatulator: How about something like:
> - Ternaries should not be nested.
> - Single line ternaries should be preferred, unless the resulting line is _long_.
>
>     ```js
>     // good 
>     const val = _.isFunction(x.foo) ? x.foo() : x.backupFoo();
>     // good
>     const name = 'nickname' in this
>         ? `${this.firstName} "${this.nickname}" ${this.surname}`
>         : `${this.firstName} ${this.surname}`;
>     ```

In my opinion "_long_" should be "> 80 characters", but this style guide suggests 100 elsewhere.

ljharb: But if the line is long, you can just use intermediate variables instead of a ternary. eg,
```js
const nameWithNick = `${this.firstName} "${this.nickname}" ${this.surname}`;
const nameWithoutNick = `${this.firstName} ${this.surname}`;
const name = this.nickname ? nameWithNick : nameWithoutNick;
```
It's still three lines, but the ternary is on a single line, and you have three statements to step through with a debugger instead of one.

DerFlatulator: That's a fair point regarding the debugger. Only down side is that you are no long lazy evaluating the branch. I wonder if engines optimize for this?

ljharb: They can't optimize it because `toString` can have side effects - but looking up and stringifying a few properties really isn't that big a deal. If it was, then you'd move those things to a function, and it'd still be a one-line ternary.

ljharb:
Seems like the discussion above about multiline ternaries has been resolved?

Please share any further thoughts on the subject; I'd like to merge this within the next day or two.

bgvianyc:
No additional thoughts here.

ljharb:
@bgvianyc Would you mind one last rebase on this? I think it's time to merge it.

bgvianyc:
How do you want it rebased? I only see one commit now.

ljharb:
@bgvianyc just stacked on top of latest master so i can fast-forward merge it :-)

bgvianyc:
@ljharb done

PR #674, opened 2016-01-12
Update "Comparison Operators & Equality" in README.md

pgeiss:
The guide should be more clear that [] is truthy because arrays are objects. As is, a reader could be confused about this because an empty string will be evaluated as false and it's possible (but not a good idea) to access a string using array-like syntax. It also doesn't help that this array-like syntax is in line with other languages like C where strings ARE arrays, further adding to potential confusion.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1250,9 +1250,9 @@ Other Style Guides
     + **Strings** evaluate to **false** if an empty string `''`, otherwise **true**
 
     ```javascript
-    if ([0]) {
+    if ([0] && []) {
       // true
-      // An array is an object, objects evaluate to true
+      // An array (even an empty one) is an object, objects will evaluate to true
     }
     ```
 


ljharb:
`[]` is truthy, it doesn't === true.

The only 'falsy' values in JavaScript are undefined, null, 0, -0, NaN, '', and false. The 'truthy' values are all other values. All objects are truthy. There is nothing both truthy and falsy, or neither truthy nor falsy.

It might be good to indicate this somewhere, but I do think that a basic level of understanding of JS is OK to assume.

pgeiss:
I suppose I should have used '==' in my request then. In any event, I didn't write '===' anywhere in the file, so I'll just edit the request.

ljharb:
`[] == true` is false also - it's `!![]` that's true.

pgeiss:
Really now. Wow, I assume way too much about the way JS logic works. Perhaps I should just write that it will evaluate as true (since that's what will happen in the end if you let the interpreter decide)?

ljharb:
The most precise way to describe it is with "truthy" and "falsy" imo.

pgeiss:
Does that look better?

ljharb:
Sure, this looks great! Are you perhaps able to rebase this down to a single commit?

pgeiss:
I think the rebase worked (it created 5820bcd which contains both commits with the proper title), but that empty commit c2ea6c2 snuck its way in when I did git push. Nothing I do locally can touch it even after pull, re-checkout, and even using its hash in the rebase program instead of HEAD~#. git log knows it exists, but git rebase seems to pretend it doesn't. Please tell me if it's a problem.

ljharb:
@pgeiss try `git rebase -i HEAD~4` and in the resulting vim prompt, change the last 3 "pick"s to "s" (for squash), and then `:wq` - and in the resulting vim prompt, edit the commit message. then force push.

pgeiss:
> git rebase -i HEAD~4
> \*vim*
> error: could not apply b79dfcb... Clearly show that [] === True
> git rebase --continue
> README.md: needs merge -- you must edit all merge conflicts and then mark them as resolved using git add

It's also probably worth noting that it gave me commits af81beb (v3.1.0) and 7f62252 ([guide] clarify function spacing in 18.3) when I did HEAD~4. I left those as picks and only s'd the ones that were mine. It did not offer me c2ea6c2 (git show HEAD shows c2ea6c2).

ljharb:
@pgeiss the easiest is to make a brand new branch, and force push to your `patch-1` branch - which will update this PR in-place. You definitely have to force push whenever you rebase.

pgeiss:
Got it! I guess all my issues were because I wasn't force pushing. Anyways, should be all set now.

pgeiss:
Rebased again after #675 and #676.

ljharb:
Sure, but this way I don't have to have a merge commit :-) Thanks!

PR #687, opened 2016-01-20
Fix syntax issue on createClass example

joaomilho:
Since `createClass` gets an object, you can't use the `method() {}` syntax, but `{method: function()}`.


ljharb:
This is incorrect - concise method syntax works identically on object literals.

Issue #685, opened 2016-01-17
Lint the readme


hzoo:
Don't know if there's already an issue to do this, but @btmills published https://github.com/eslint/eslint-plugin-markdown. So you should be able to lint the readme with the eslint config?

ljharb:
That would be excellent - the trick is that the readme, intentionally, has a bunch of "bad" examples.

If we could lint all the "good" ones, and even better, require the linter to **fail** on all the "bad" ones, that would be stellar.

hzoo:
I'l make a feature request in the repo.

joaomilho:
+1

kjesek:
Is this issue still available?

ljharb:
@kjesek if you mean, you want to try it, sure, that'd be great!

raunofreiberg:
Any progress @kjesek? Super interested in figuring this one out, let me know if I can help in any way ✌️

Garv978:
is this availabble ?

PR #688, opened 2016-01-20
Favor stateless functions over classes when there's no state

joaomilho:


react/README.md
```diff
@@ -27,25 +27,44 @@
...
+
+    // bad
+    class Listing extends React.Component {
+      render() {
+        return <div>{ this.props.hello }</div>;
+      }
+    }
+    
+    // good
+    function Listing({ hello }){
```
ljharb: there should always be a space between `)` and `{`

react/README.md
```diff
@@ -27,25 +27,44 @@
...
+      return <div>{ hello }</div>;
```
ljharb: jsx curly braces should not have spaces inside them (here or line 59)

ljharb:
Thanks, this looks great pending a few comments.

It would be ideal if there was a linter rule that could help enforce this, although it may not be possible.

react/README.md
joaomilho: solved.

react/README.md
joaomilho: solved.

react/README.md
```diff
@@ -5,7 +5,7 @@
 ## Table of Contents
 
   1. [Basic Rules](#basic-rules)
-  1. [Class vs `React.createClass`](#class-vs-reactcreateclass)
+  1. [Class vs `React.createClass` vs stateless](#class-vs-reactcreateclass-vs-stateless)
```
ljharb: it's unfortunate that this will break the link - however since the title on line 28 didn't change, are you sure this link works now?

react/README.md
```diff
@@ -27,25 +27,44 @@
...
 
   eslint rules: [`react/prefer-es6-class`](https://github.com/yannickcr/eslint-plugin-react/blob/master/docs/rules/prefer-es6-class.md).
 
     ```javascript
     // bad
     const Listing = React.createClass({
-      render() {
-        return <div />;
+      // ...
+      render: function() {
```
ljharb: per our other linting rules, this should be `render() {`, even in an object literal.

react/README.md
```diff
@@ -27,25 +27,44 @@
...
+
+    // bad
+    class Listing extends React.Component {
+      render() {
+        return <div>{this.props.hello}</div>;
+      }
+    }
+    
+    // good
+    function Listing ({ hello }) {
```
kesne: There should be no space after listing (i.e. this should be `function Listing({ hello }) {`)

ljharb: thanks, good catch

kesne: This is an example of why #685 would be awesome.

joaomilho: done

react/README.md
joaomilho: done

react/README.md
joaomilho: now it works :D

joaomilho: Yes

joaomilho:
It's all fixed.

PR #695, opened 2016-01-26
[rule links] reduce visual clutter

hshoff:
It's great that we added the eslint and jscs links for each rule, but I think there's an opportunity here to reduce the visual clutter a bit for easier reading and skimming. I'm proposing moving the links to a comma separated list at the end of the corresponding rule. I feel it's not too difficult to identify which links are eslint vs jscs, but open to discussion on bringing back the labels.

before:
screen shot 2016-01-25 at 9 49 01 pm

^ seeing all that space used for a label and a link makes me :cry: 

after:
screen shot 2016-01-25 at 10 01 19 pm

@ljharb

### DIFF
--- a/README.md
+++ b/README.md
@@ -87,11 +87,9 @@ Other Style Guides
 
 ## References
 
-  - [2.1](#2.1) <a name='2.1'></a> Use `const` for all of your references; avoid using `var`.
+  - [2.1](#2.1) <a name='2.1'></a> Use `const` for all of your references; avoid using `var`. [`prefer-const`](http://eslint.org/docs/rules/prefer-const.html), [`no-const-assign`](http://eslint.org/docs/rules/no-const-assign.html)
 
     > Why? This ensures that you can't reassign your references, which can lead to bugs and difficult to comprehend code.
-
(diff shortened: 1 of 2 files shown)


kesne:
I love this <3

ljharb:
I'm down for the line condensing, but I think labelling as "eslint" and "jscs" is critical, especially since there's 0 to N rules that could apply for each.

hshoff:
I battled with that too, but here's how I settled on removing the labels:

- `eslint-rule` vs `jscsRule`
- hovering over the link
- we can remove 37 instances of eslint and 36 instances of jscs
- what's the use case where a reader wishes they could get to the eslint rule faster?

ljharb:
hovering isn't a good affordance and doesn't work on mobile :-/

To me it's not about getting to it faster, as much as it's about clearly identifying, at a glance, that we have **eslint** and **jscs** rule coverage. The rule names are secondary (but necessary)

hshoff:
There came a point where I was thinking to myself, "I shouldn't have done this by hand"...but it was already too late.

ljharb:
LGTM

PR #698, opened 2016-01-26
Remove language interpretable as an ableist slur

quarterto:


ljharb:
While I'm pretty dubious that anyone would actually interpret this word as an ableist slur, let alone in this context, I'll always prefer to err on the side of caution. Thanks for the PR!

PR #700, opened 2016-01-27
Added names to invalid function declarations

LarryBattle:
Function declarations can’t be anonymous.


ljharb:
Arguably they're valid in expression position, but these changes certainly don't hurt.

PR #701, opened 2016-01-27
Edit sample code in 24.1 to follow rule 8.1

fredrikolovsson:
Replaced function expression with arrow function notation as per https://github.com/airbnb/javascript#arrow-functions


lencioni:
Good catch, and thanks for the PR @fredrikolovsson!

PR #703, opened 2016-01-28
Clarified ambiguity in the README

framerate:
I assumed by step 1 that the react plugin was inherited through use of `extend: airbnb` but it turns out not to be the case.


framerate:
User Error. Doh!

PR #706, opened 2016-01-29
tweaked syntax of existing comment

MissedMemo:


ljharb:
Thanks for the contribution! I prefer the way it's currently worded, and your change would erase the pop culture reference.

PR #711, opened 2016-02-02
Migrate to eslint@2.0.0-beta

johangirod:
- Replace removed rules in eslint@2.x with their counterparts (see [migration guide](http://eslint.org/docs/2.0.0/user-guide/migrating-to-2.0.0#removed-rules))
- Migrating ecmaOptions (see [migration guide](http://eslint.org/docs/2.0.0/user-guide/migrating-to-2.0.0\#language-options))
- Update eslint version in package.json to 2.0.0-beta

This is obviously breaking, so it might be for a future major version.

### DIFF
--- a/packages/eslint-config-airbnb/package.json
+++ b/packages/eslint-config-airbnb/package.json
@@ -39,13 +39,13 @@
   "homepage": "https://github.com/airbnb/javascript",
   "devDependencies": {
     "babel-tape-runner": "1.2.0",
-    "eslint": "^1.10.3",
+    "eslint": "^2.0.0-beta",
     "eslint-plugin-react": "^3.12.0",
     "react": "^0.13.3",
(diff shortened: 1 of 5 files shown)


ljharb:
Thanks! This will be useful to keep open for when eslint v2 is released, but updating to prerelease versions for anything but local testing is highly unadvised, so we won't be merging this until it's an actual release.

johangirod:
That's ok, I just needed it for personal uses, so I figured why not do a PR. I guess the v2 will be released soon.

Arcanemagus:
@johangirod Looks like the version of `eslint` you have specified needs to be updated to the released version.

hshoff:
Thanks for this @johangirod, we're going to jump the beta to the `2.0.0` release.

Will track progress here: https://github.com/airbnb/javascript/pull/730

PR #712, opened 2016-02-02
Avoid lexical declarations in case/default clauses

lencioni:
This commit adds guidance warning people to avoid lexical declarations
in case/default clauses and enables the corresponding ESLint rule.

  http://eslint.org/docs/rules/no-case-declarations.html

We didn't have a section on switch statements yet, so I thought about
starting one. It seemed like it would best fit between section 15
(Comparison Operators & Equality) and section 16 (Blocks), but I didn't
want to mess up all of the following numberings, since people probably
have references to them. I considered adding this near the end to
minimize this effect, but it really seemed to belong near these other
things. I landed on appending it to Section 15 (Comparison Operators &
Equality) and I think it sorta fits there since switch statements are a
related concept.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1245,6 +1245,47 @@ Other Style Guides
     ```
 
   - [15.4](#15.4) <a name='15.4'></a> For more information see [Truth Equality and JavaScript](http://javascriptweblog.wordpress.com/2011/02/07/truth-equality-and-javascript/#more-2108) by Angus Croll.
+  - [15.5](#15.5) <a name='15.5'></a> Use braces to create blocks in `case` and `default` clauses that contain lexical declarations (e.g. `let`, `const`, `function`, and `class`).
+
+  > Why? Lexical declarations are visible in the entire `switch` block but only get initialized when assigned, which only happens when its `case` is reached. This causes problems when multiple `case` clauses attempt to define the same thing.
+
+  eslint rules: [`no-case-declarations`](http://eslint.org/docs/rules/no-case-declarations.html).
+
+    ```javascript
+    // bad
+    switch (foo) {
+      case 1:
+        let x = 1;
+        break;
+      case 2:
+        const y = 2;
+        break;
+      case 3:
+        function f() {}
+        break;
+      default:
+        class C {}
+    }
+
+    // good
+    switch (foo) {
+      case 1: {
+        let x = 1;
+        break;
+      }
+      case 2: {
+        const y = 2;
+        break;
+      }
+      case 3: {
+        function f() {}
+        break;
+      }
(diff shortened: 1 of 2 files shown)


README.md
```diff
@@ -1237,6 +1237,48 @@ Other Style Guides
 
   - [15.4](#15.4) <a name='15.4'></a> For more information see [Truth Equality and JavaScript](http://javascriptweblog.wordpress.com/2011/02/07/truth-equality-and-javascript/#more-2108) by Angus Croll.
 
+  - [15.5](#15.5) <a name='16.1'></a> Avoid lexical declarations (e.g. `let`, `const`, `function`, and `class`) in `case` and `default` clauses.
```
kesne: Maybe clear up the wording so that it's clear that you should just create a block when you want to do this?

ljharb: also, is it 15.5 or 16.1?

lencioni: I pushed a new revision that addresses both of these comments.

README.md
```diff
@@ -1245,6 +1245,47 @@ Other Style Guides
...
+      }
+      case 2: {
+        const y = 2;
+        break;
+      }
+      case 3: {
+        function f() {}
+        break;
+      }
+      default: {
```
ljharb: should this "good" case include an example case that *doesn't* require a block?

lencioni: Seems reasonable, unless we don't have an opinion about whether it should have a block or not. I generally prefer simpler rules so I'd be in favor of always using blocks, but I don't feel strongly about it.

kesne: I don't think we force blocks if you're not doing anything that needs them.

ljharb: and in such scenarios, do we want all `case`s to have blocks for consistency? or just the ones that need it?

README.md
nivesh2: would be more suitable for 15.5 as it talks about **Comparison Operators** and we are preventing use of _lexical declarations in switch statements_.

ljharb: switch statements are relevant to comparison - my thinking was that Section 15 is the right place for constructs used for equality or comparison :-)

README.md
kesne: IMO, only blocks that need it. It matches the as-needed feel of [section 3.8](https://github.com/airbnb/javascript#3.8).

ljharb: so since there's no "as needed, consistent" setting, i think the example should indeed include a `case` that doesn't need and thus omits the block.

lencioni: Thanks for the feedback. I added an example in 4d74fe1f5ab7e2f67e4ad837387debbf276e3929.

PR #715, opened 2016-02-04
Sort static methods above constructor for React

lencioni:
I think it makes more sense to put static methods above the constructor
in classes. I would like to update the ESLint configuration to match
this, but it looks like the react/sort-comp rule does not support it
quite yet.

  https://github.com/yannickcr/eslint-plugin-react/issues/128


ljharb:
LGTM

PR #724, opened 2016-02-10
typo in rules style.js

brutalcrozt:
well this clearer for me to read.
hope can get some feedback :smile:

### DIFF
--- a/packages/eslint-config-airbnb/rules/style.js
+++ b/packages/eslint-config-airbnb/rules/style.js
@@ -6,7 +6,7 @@ module.exports = {
     'brace-style': [2, '1tbs', { 'allowSingleLine': true }],
     // require camel case names
     'camelcase': [2, { 'properties': 'never' }],
-    // enforce spacing before and after comma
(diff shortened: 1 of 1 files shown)


packages/eslint-config-airbnb/rules/style.js
```diff
@@ -6,7 +6,7 @@ module.exports = {
     'brace-style': [2, '1tbs', { 'allowSingleLine': true }],
     // require camel case names
     'camelcase': [2, { 'properties': 'never' }],
-    // enforce spacing before and after comma
```
ljharb: ah - i misunderstood your question in gitter. in fact, this is correct as-is, because "false" enforces that there is *no* space before the comma. it's "enforce spacing", not "enforce spaces".

brutalcrozt: never mind, my english just bad.
thanks for explaining that :smile:

ljharb: np, thanks for the contribution!

PR #729, opened 2016-02-12
no-empty-label was renamed to no-labels in ESLint v2

silvenon:
eslint/eslint#5129


silvenon:
Didn't see #711.

ljharb:
Thanks for the contribution! Currently we're still on v1.x.

I'll be updating to v2.0 at the same time as updating all these rules. Accepting this as-is would break everyone using the package, which currently requires 1.x and does not work with 2.x.

silvenon:
Yeah, I didn't do my research :sweat_smile:

ljharb:
Tracked in #730

PR #749, opened 2016-02-21
Small typo fix in 18.6 'bad'-example.

Keksike:
Function call 'class()' to 'classed()'.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1710,7 +1710,7 @@ Other Style Guides
         .updateCount();
 
     // bad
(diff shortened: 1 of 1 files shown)


ljharb:
Why is this a bad example?

Keksike:
I meant it's literally in the 'bad' examples of 18.6. (showing you how not to do something), not that using 'class()' is bad.

The 'bad' example uses `class()` and the 'good' example right under it uses `classed()`. I just fixed this to be consistent, and I believe the D3 function `classed()` is the actual function that is supposed to be called here.

ljharb:
Makes sense, thanks for the clarification!

PR #762, opened 2016-02-24
Enforce literal syntax for array creation

jahvi:
[As per your style guide](https://github.com/airbnb/javascript/blob/30c448d1df2c5a16bee4dd618ebdcf908b7475a3/README.md#4.1) you should create arrays using the literal syntax but the current value for the `no-array-constructor` rule is `0` which is disabled.

Changing this to `2` should show errors when arrays are created with constructors.


ljharb:
Thanks!

PR #767, opened 2016-03-01
remove statement on function expressions not being named

afeld:
Confirmed that the function names _do_ appear in Chrome 49.0.2, Safari 9.0.3, Firefox 45.0b6, and Node 5.6.0, using the following code:

```javascript
var cb = function() {
  throw new Error("foo");
};
setTimeout(cb, 1000);
```

screen shot 2016-03-01 at 12 03 53 pm


ljharb:
Thanks for your contribution!

Certainly in modern browsers (as per ES6) names for debugging purposes are inferred. However, it's useful to be able to grep for function names - and at Airbnb we support many more browsers than "the latest" - so since this advice remains true in browsers we support, it should stay in.

PR #772, opened 2016-03-02
Enable react/prefer-stateless-function rule

petersendidit:
Change to enable the [`react/prefer-stateless-function`](https://github.com/yannickcr/eslint-plugin-react/commit/0830226e87ad7fd28ca4ed1b45e838cbca78b2a9) once its released.


ljharb:
Thanks, but there's no need to submit this PR until it is actually released - I'm quite aware it's upcoming :-)

ljharb:
The plugin has been updated. https://github.com/yannickcr/eslint-plugin-react/releases/tag/v4.2.0

PR #780, opened 2016-03-08
[guide] Permanent links

ljharb:
- Preserve the original links, because cool URLs don’t change.

The idea will be, we'll never again add a numbered link - and we'll be able to freely reorder the sections, but we'll also never *change* the existing numbered links.

### DIFF
--- a/README.md
+++ b/README.md
@@ -52,7 +52,8 @@ Other Style Guides
 
 ## Types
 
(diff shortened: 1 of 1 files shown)


hshoff:
Yes! x5 million :trophy:

kesne:
:heart: This is great.

PR #782, opened 2016-03-09
Allow arrow functions in JSX props

gilbox:
Rationale: 

- Improves code readability. 
- Results in simpler solutions than the alternative.

### DIFF
--- a/packages/eslint-config-airbnb/rules/react.js
+++ b/packages/eslint-config-airbnb/rules/react.js
@@ -41,8 +41,8 @@ module.exports = {
     // Prevent usage of .bind() and arrow functions in JSX props
     // https://github.com/yannickcr/eslint-plugin-react/blob/master/docs/rules/jsx-no-bind.md
     'react/jsx-no-bind': [2, {
-      'ignoreRefs': false,
-      'allowArrowFunctions': false,
+      'ignoreRefs': true,
+      'allowArrowFunctions': true,
       'allowBind': false,
     }],
     // Prevent duplicate props in JSX
--- a/react/README.md
+++ b/react/README.md
@@ -274,6 +274,23 @@
 
 ## Methods
 
+  - Use arrow functions to close over local variables.
+
+    ```javascript
+    function ItemList(props) {
+      return (
+        <ul>
+          {props.items.map((item, index) => (
+            <Item
+              key={item.key}
+              onClick={() => doSomethingWith(item.name, index)}
+            />
+          ))}
+        </ul>
+      );
+    }
+    ```
+
   - Bind event handlers for the render method in the constructor. eslint: [`react/jsx-no-bind`](https://github.com/yannickcr/eslint-plugin-react/blob/master/docs/rules/jsx-no-bind.md)
 
   > Why? A bind call in the render path creates a brand new function on every single render.


kesne:
You should reference the eslint rule (which in this case is also `react/jsx-no-bind`).

tleunen:
:-1: 
You can always find a way to avoid this. For example passing the item and the index as props, with the onClick callback.

ljharb:
@tleunen what about at the bottom level, where you directly create a native DOM element?

tleunen:
What do you mean by that?

To me, it's way better to have
```jsx
function ItemList(props) {
  return (
    <ul>
      {props.items.map((item, index) => (
        <Item
          key={item.key}
          item={item}
          listIndex={index}
          onItemClick={doSomethingWith}
        />
      ))}
    </ul>
  );
}
```
And then Item can attach the `onClick` event where needed and call `props.doSomethingWith(props.item.name, props.listIndex)`

The downside is that it has to know the list index, but well.. I think it's better than creating a new bound function every time.

ljharb:
@tleunen ok so what does the implementation of `Item` look like? *somewhere* you are creating the `<a>` or `<input>` and that place needs to create a new function for the event handler.

tleunen:
Sure, but you only create it once

```jsx
class Item extends React.Component {
    constructor(props) {
        super(props);
        this.handleClick = this.handleClick.bind(this);
    }
    handleClick(e) {
        this.props.doSomethingWith(props.item.name, props.listIndex);
    }
    render() {
        return <button onClick={this.handleClick}>Click me</button>
    }
}

**Comment by ljharb** [comment]
Right, so the question is really whether `Item` should be an SFC with an arrow function in `onClick`, or whether it should be a `class`-based component like in your example.

**Comment by kesne** [comment]
To me, the arrow function seems a lot cleaner: 

```javascript
function Item({ doSomethingWith, item, listIndex }) {
  return <button onClick={() => doSomethingWith(item.name, listIndex)}>Click me</button>;
}
```

tleunen:
I know React can apply some performance optimizations on the SFC, but would that be better than having the class without recreating the function at every render?

ljharb:
First, as I have recently learned, the only advantage with the constructor-binding is that `.bind` is slow - creating new functions is not.
Second, performance should always be a secondary concern to code clarity and readability - something would have to be *very* slow to trump readability concerns, and I've been convinced that creating non-bound functions as event handlers (not as props, however, just as bottom-level event handlers) is not a performance hit due to the design of React.

Thus this is mostly a question of subjective readability, I think.

tleunen:
Make sense. So creating non-bound function is fine, but not otherwise.

lelandrichardson:
I am in favor of allowing arrow functions in render methods. We should continue disallowing `.bind(this)` though

packages/eslint-config-airbnb/rules/react.js
```diff
@@ -42,7 +42,7 @@ module.exports = {
     // https://github.com/yannickcr/eslint-plugin-react/blob/master/docs/rules/jsx-no-bind.md
     'react/jsx-no-bind': [2, {
       'ignoreRefs': false,
```
lelandrichardson: will we be able to do this, then: `ref={foo => this.foo = foo}`?

ljharb: only if this was set to `true`, as i understand it

lelandrichardson: if arrow fns were allows, seems like we should change that as well?

tleunen:
@lelandrichardson Would `this` be allowed in an arrow function?
I know Babel doesn't create a bound function, but how does the browser see the arrow function with the usage of `this` inside?

ljharb:
@tleunen the same as one without. it basically does a `var self = this` and creates a closure.

packages/eslint-config-airbnb/rules/react.js
kesne: I agree, arrows for refs are useful when you want to use inputs in a SFC.

gilbox:
Which is more readable?

*(A)* Foo passess the data down and Bar passes it back up again
```
function Foo({ data }) {
  return (
    <Bar onSomething={Actions.zoom} data={data} />
  )
}
```

*(B)* Use a closure
```
function Foo({ data }) {
  return (
    <Bar onSomething={() => Actions.zoom(data)} />
  )
}
```

The answer is *(B)* because you don't need to look at the code of `Bar` to understand what data is being passed into `Actions.zoom`. This is why we should usually favor using closures even for non-leaf nodes. Passing the same data down and then back up again is an anti-pattern imo.

ljharb:
I agree that B is more readable - however with the rules as-is on master, you'd model that as a stateful component.

PR #785, opened 2016-03-10
added no-dupe-class-members

aaqibhabib:
See http://eslint.org/docs/rules/no-dupe-class-members


ljharb:
Thanks - this isn't in the guide, so could we add a section about it that references the eslint rule?

PR #789, opened 2016-03-14
Add space to function signature

JeremyBernier:
At 7.11 you say that function signatures should have spacing, yet there's no space in this function signature. This is confusing.

### DIFF
--- a/README.md
+++ b/README.md
@@ -563,7 +563,7 @@ Other Style Guides
     };
 
     // good
-    function foo() {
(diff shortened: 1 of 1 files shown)


ljharb:
As you no doubt noticed, only after `function`, never after a function's name / before the parens :-)

lencioni:
@JeremyBernier can you suggest any changes that might make this clearer or less confusing?

PR #793, opened 2016-03-21
Fix "object destructuring for multiple return values" example

Arnavion:
... to use the same destructured properties in the good and bad code.


ljharb:
I'm not really sure it matters one way or another - one might argue that using different properties demonstrates the versatility of object destructuring over array destructuring. Sure tho, why not?

PR #813, opened 2016-04-06
Add some accessibility rules to the React style guide

lencioni:
I think there is a good opportunity to improve accessibility guidelines in this style guide. This PR takes the first steps in that direction by enabling a handful of rules from eslint-plugin-jsx-a11y and adding some text about these rules.

There are more rules from this plugin that I would like to enable, but I'm going to leave that for another time.

While I was at it, I switched the syntax highlighting on the codeblocks in the react style guide to highlight JSX better.

### DIFF
--- a/packages/eslint-config-airbnb/test/test-react-order.js
+++ b/packages/eslint-config-airbnb/test/test-react-order.js
@@ -28,9 +28,9 @@ ${body}
 }
 
 test('validate react prop order', t => {
-  t.test('make sure our eslintrc has React linting dependencies', t => {
+  t.test('make sure our eslintrc has React and JSX linting dependencies', t => {
     t.plan(1);
-    t.equal(reactRules.plugins[0], 'react', 'uses eslint-plugin-react');
+    t.deepEqual(reactRules.plugins, ['jsx-a11y', 'react']);
   });
 
   t.test('passes a good component', t => {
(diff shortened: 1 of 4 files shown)


packages/eslint-config-airbnb/test/test-react-order.js
```diff
@@ -28,9 +28,10 @@ ${body}
 }
 
 test('validate react prop order', t => {
-  t.test('make sure our eslintrc has React linting dependencies', t => {
-    t.plan(1);
-    t.equal(reactRules.plugins[0], 'react', 'uses eslint-plugin-react');
+  t.test('make sure our eslintrc has React and JSX linting dependencies', t => {
+    t.plan(2);
+    t.equal(reactRules.plugins[0], 'jsx-a11y', 'uses eslint-plugin-jsx-a11y');
```
ljharb: `t.deepEqual(reactRules.plugins, ['jsx-a11y', 'react'])`?

ljharb:
Awesome, other than my nonblocker comment, this LGTM. We've already queued up a few breaking changes since the last release, so this one can be bundled along with them.

packages/eslint-config-airbnb/test/test-react-order.js
lencioni: Done!

lencioni:
cc @evcohen

evcohen:
@lencioni thanks for all of your help so far and let me know what other rules we can prioritize building out next.

PR #816, opened 2016-04-08
Improve guideline for image alt text

lencioni:
After digging into this rule a little more with @evcohen, I believe that
it is okay for images to have an empty string for alt text. I expect the
next release of the linter rule to allow for this as well.

More context:

  https://github.com/evcohen/eslint-plugin-jsx-a11y/issues/6


ljharb:
Can we include a bump to `v0.6.2` for `eslint-plugin-a11y` in the same commit?

lencioni:
Done!

PR #824, opened 2016-04-13
Link to eslint.org instead of github where possible

lencioni:
In most cases we link to documentation on the eslint.org site instead of
on GitHub. I found a couple of stragglers and decided to make things a
little more consistent.

### DIFF
--- a/packages/eslint-config-airbnb/rules/es6.js
+++ b/packages/eslint-config-airbnb/rules/es6.js
@@ -18,7 +18,7 @@ module.exports = {
     // require parens in arrow function arguments
     'arrow-parens': 0,
     // require space before/after arrow function's arrow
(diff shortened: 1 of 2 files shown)


ljharb:
This is great for the sake of consistency. I wonder tho if it would be better to link to specific SHAs on github? the docs could get out of sync with the rules.

lencioni:
I don't think I quite follow your comment about linking to specific SHAs. Wouldn't that get out of sync as the rules are updated?

ljharb:
Yeah, that's true - it would definitely get out of sync as eslint was updated *in the package*, but the key is that it wouldn't get out of sync as eslint *itself* is updated.

lencioni:
thanks a lot

PR #825, opened 2016-04-13
Enable import/no-unresolved and import/export rules

lencioni:
These come from eslint-plugin-import. We've been using a couple of these rules for a
while at Airbnb so it seems like a good time to include them in this
configuration. We've found the no-unresolved rule in particular to be
very helpful when working across a large codebase.

  https://github.com/benmosher/eslint-plugin-import

Since the no-unresolved rule works on commonjs, I decided to include it
in the node configuration as well.

### DIFF
--- a/packages/eslint-config-airbnb/package.json
+++ b/packages/eslint-config-airbnb/package.json
@@ -45,6 +45,7 @@
     "babel-tape-runner": "^1.3.1",
     "eslint": "^2.7.0",
     "eslint-plugin-jsx-a11y": "^0.6.2",
+    "eslint-plugin-import": "^1.4.0",
     "eslint-plugin-react": "^4.3.0",
     "react": "^0.14.8",
     "tape": "^4.5.1",
@@ -53,6 +54,7 @@
   "peerDependencies": {
     "eslint": "^2.7.0",
     "eslint-plugin-jsx-a11y": "^0.6.2",
+    "eslint-plugin-import": "^1.4.0",
     "eslint-plugin-react": "^4.3.0"
   }
 }
--- a/packages/eslint-config-airbnb/rules/es6.js
+++ b/packages/eslint-config-airbnb/rules/es6.js
@@ -11,6 +11,9 @@ module.exports = {
       'objectLiteralDuplicateProperties': false
     }
   },
+  'plugins': [
+    'import'
+  ],
   'rules': {
     // enforces no braces where they can be omitted
     // http://eslint.org/docs/rules/arrow-body-style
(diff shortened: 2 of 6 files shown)


packages/eslint-config-airbnb/package.json
```diff
@@ -53,6 +54,7 @@
   "peerDependencies": {
     "eslint": "^2.7.0",
     "eslint-plugin-jsx-a11y": "^0.6.2",
+    "eslint-plugin-import": "^1.4.0",
```
ljharb: let's add this peer dep to the readme also

packages/eslint-config-airbnb/rules/es6.js
```diff
@@ -83,6 +86,19 @@ module.exports = {
...
+    // https://github.com/benmosher/eslint-plugin-import/blob/master/docs/rules/export.md
+    'import/export': 2,
+    // ensure imports point to files/modules that can be resolved
+    // https://github.com/benmosher/eslint-plugin-import/blob/master/docs/rules/no-unresolved.md
+    'import/no-unresolved': [2, { 'commonjs': true }]
+  },
+  'settings': {
+    'import/resolver': {
+      'node': {
+        'extensions': ['.js', '.jsx', '.json']
```
ljharb: `es6` probably shouldn't include `jsx` - and actually, probably `react` should be the file where `jsx` is added.

packages/eslint-config-airbnb/package.json
lencioni: Done!

packages/eslint-config-airbnb/rules/es6.js
lencioni: Done!

lencioni:
I found a couple more that I think we should enable. Stand by.

lencioni:
Alrighty, this is ready to go.

packages/eslint-config-airbnb/rules/es6.js
```diff
@@ -83,6 +86,22 @@ module.exports = {
     'template-curly-spacing': 2,
     // enforce spacing around the * in yield* expressions
     // http://eslint.org/docs/rules/yield-star-spacing
-    'yield-star-spacing': [2, 'after']
+    'yield-star-spacing': [2, 'after'],
+    // disallow invalid exports, e.g. multiple defaults
+    // https://github.com/benmosher/eslint-plugin-import/blob/master/docs/rules/export.md
+    'import/export': 2,
+    // prevents `import * from 'foo';`
```
ljharb: can we add the rule link to https://github.com/airbnb/javascript#modules--no-wildcard ?

lencioni: Good call. Done!

PR #826, opened 2016-04-14
[guide] Attempt to clarify filename conventions

### LINKED ISSUE #817: Filename and directory structure
I have a question about the naming convention, the styleguide specifies that if the file exports a single class, the name of the file should be the name of the class. (see https://github.com/airbnb/javascript#naming--filename-matches-export)

So something like this:
```js
import CheckBox from './CheckBox';
```

What about the directory structure? Is there any preference?
Because, these 2 structures are good based on the styleguide, but the import syntax will be different.
```
├── src
│   ├── CheckBox
│   │   ├── index.js
│   │   ├── CheckBox.js
```
```js
import CheckBox from './CheckBox';
```
```
├── src
│   ├── checkbox
│   │   ├── index.js
│   │   ├── CheckBox.js
```
```js
import CheckBox from './checkbox';
```

Does airbnb have a preference?
For internal component, I find it ok to have import with capital letters, but in an external library, it doesn't seem very common to have `import CheckBox from 'component-lib/CheckBox';`

ljharb:
Fixes #817.

### DIFF
--- a/README.md
+++ b/README.md
@@ -2337,24 +2337,39 @@ Other Style Guides
     ```
 
   <a name="naming--filename-matches-export"></a><a name="22.6"></a>
-  - [22.6](#naming--filename-matches-export) If your file exports a single class, your filename should be exactly the name of the class.
+  - [22.6](#naming--filename-matches-export) A base filename should exactly match the name of its default export.
 
     ```javascript
-    // file contents
+    // file 1 contents
     class CheckBox {
       // ...
     }
     export default CheckBox;
 
+    // file 2 contents
+    export default function fortyTwo() { return 42; }
+
+    // file 3 contents
+    export default function insideDirectory() {}
+
     // in some other file
     // bad
-    import CheckBox from './checkBox';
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -2337,24 +2337,39 @@ Other Style Guides
...
     export default CheckBox;
 
+    // file 2 contents
+    export default function fortyTwo() { return 42; }
+
+    // file 3 contents
+    export default function insideDirectory() {}
+
     // in some other file
     // bad
```
lencioni: It might be good to, as part of these `bad` comments, include couple words about how it is bad, since the distinction might now be pretty subtle.

ljharb: done!

ryerh:
@ljharb @lencioni 
How to choose a proper filename in the following scenario ?

case 1: no default export
```js
export function foo() {}
export function bar() {}
```

case 2: both default & non-default exports
```js
export function foo() {}
export default function bar() {}
```

ljharb:
In the former case, a meaningful name should be obvious; otherwise, the exports should be put into separate files. In the latter case, the default export determines the filename.

PR #829, opened 2016-04-16
Add `eslint-config-airbnb-base`

### LINKED ISSUE #751: npm@3 warn missing eslint-plugin-react a non react project
Seems like eslint-plugin-react was moved to peer dependency, and now I'm getting
```
├── UNMET PEER DEPENDENCY eslint-plugin-react@^4.0.0
```

I'm using eslint-config-airbnb/base, for a non react project.

How can I get rid of this warning _without_ installing eslint-plugin-react?
Is there any plan for splitting /base to a separate package?

ljharb:
Fixes #751.

This creates a new package, `eslint-config-airbnb-base`, which will be used by the main `eslint-config`, and will contain only the JS-related rules, and not the React/JSX ones.

Once this is ready to go in, I'll publish the base package, excise the temporary commit, and then update + merge the PR.


josmardias:
Is eslint-plugin-import really a peerDependency, instead of dependency?

josmardias:
Btw, this base package worked for me.

ljharb:
@josmardias that's a good point. Perhaps all of the plugins should become normal dependencies instead of peer deps (including `eslint-plugin-react`, etc)

tleunen:
If they are dependencies, does it mean we are not required to manually install them? Or because of the way eslint works, this requirement is still necessary?

ljharb:
@tleunen afaict, correct, you'd no longer have to manually install them - unless you were referencing the plugins in your own eslint configs.

tleunen:
I always thought it was devDep + peerDep because of this https://github.com/eslint/eslint/issues/2518

ljharb:
aha, you're right. https://github.com/eslint/eslint/issues/2518#issuecomment-102433136 specifically - so they need to remain peer deps, sadly.

josmardias:
Yeah, right, we can live with that. I was writing a long post about my toughts, but I think it rest the case :)

josmardias:
@ljharb 
Not sure if it's the same as you are saying but, what about this:

```
// eslint-config-airbnb-base
dependencies:
 - eslint-plugin-import
- ....
peerDependencies:
 - eslint

// eslint-config-airbnb
dependencies:
 - eslint-plugin-react
 - eslint-config-airbnb-base
- ...
peerDependencies:
 - eslint
```

So, we can have a standalone base package that can be used as is.

ljharb:
@josmardias if you read https://github.com/eslint/eslint/issues/2518#issuecomment-102433136, eslint itself requires that plugins be peerDependencies.

josmardias:
I get it :(

packages/eslint-config-airbnb-base/.eslintrc
```diff
@@ -0,0 +1,8 @@
+{
+  "extends": "./index.js",
+  "rules": {
+    // disable requiring trailing commas because it might be nice to revert to
+    // being JSON at some point, and I don't want to make big changes now.
```
lencioni: I'm curious about your thoughts around reverting to being JSON. Any reasons why?

packages/eslint-config-airbnb-base/README.md
```diff
@@ -0,0 +1,34 @@
...
+
+[![npm version](https://badge.fury.io/js/eslint-config-airbnb-base.svg)](http://badge.fury.io/js/eslint-config-airbnb-base)
+
+This package provides Airbnb's base JS .eslintrc as an extensible shared config.
+
+## Usage
+
+We export two ESLint configurations for your usage.
+
+### eslint-config-airbnb
```
lencioni: Should this be `eslint-config-airbnb-base`

packages/eslint-config-airbnb-base/README.md
```diff
@@ -0,0 +1,34 @@
...
+
+Our default export contains all of our ESLint rules, including ECMAScript 6+
+and React. It requires `eslint` and `eslint-plugin-import`.
+
+1. `npm install --save-dev eslint-config-airbnb-base eslint-plugin-import eslint`
+2. add `"extends": "airbnb-base"` to your .eslintrc
+
+### eslint-config-airbnb/legacy
```
lencioni: `eslint-config-airbnb-base/legacy`?

packages/eslint-config-airbnb-base/README.md
```diff
@@ -0,0 +1,34 @@
...
+
+Lints ES5 and below. Requires `eslint` and `eslint-plugin-import`.
+
+1. `npm install --save-dev eslint-config-airbnb eslint-plugin-import eslint`
```
lencioni: `eslint-config-airbnb-base`?

packages/eslint-config-airbnb-base/README.md
```diff
@@ -0,0 +1,34 @@
...
+2. add `"extends": "airbnb/legacy"` to your .eslintrc
+
+See [Airbnb's overarching ESLint config](https://npmjs.com/eslint-config-airbnb), [Airbnb's Javascript styleguide](https://github.com/airbnb/javascript) and the [ESlint config docs](http://eslint.org/docs/user-guide/configuring#extending-configuration-files) for more information.
```
lencioni: I think this URL might also need to be updated.

packages/eslint-config-airbnb-base/.eslintrc
ljharb: This is copied verbatim from the `eslint-config-airbnb` `.eslintrc`

packages/eslint-config-airbnb-base/README.md
ljharb: fixed

packages/eslint-config-airbnb-base/README.md
ljharb: fixed

packages/eslint-config-airbnb-base/package.json
```diff
@@ -0,0 +1,55 @@
+{
+  "name": "eslint-config-airbnb-base",
+  "version": "0.0.0",
```
lencioni: 1.0.0, but I assume you'll do that when you tag/publish

packages/eslint-config-airbnb-base/README.md
ljharb: It seems to work fine?

I will add the missing Oxford comma, though :-)

packages/eslint-config-airbnb-base/package.json
ljharb: ah good call. i intend to tag this commit, so i'll update the version number too.

packages/eslint-config-airbnb/package.json
```diff
@@ -41,9 +41,13 @@
     "url": "https://github.com/airbnb/javascript/issues"
   },
   "homepage": "https://github.com/airbnb/javascript",
+  "bundledDependencies": [
+    "eslint-config-airbnb-base"
+  ],
   "devDependencies": {
     "babel-tape-runner": "^1.3.1",
     "eslint": "^2.8.0",
+	"eslint-config-airbnb-base": "^1.0.0",
```
lencioni: I think you used hard tabs here when the rest of the file uses spaces

packages/eslint-config-airbnb/package.json
```diff
@@ -53,8 +57,8 @@
   },
   "peerDependencies": {
     "eslint": "^2.8.0",
+	"eslint-config-airbnb-base": "^1.0.0",
```
lencioni: same

packages/eslint-config-airbnb/package.json
ljharb: fixed

packages/eslint-config-airbnb-base/.eslintrc
lencioni: Gotcha. Still, any insight?

packages/eslint-config-airbnb-base/README.md
```diff
@@ -0,0 +1,33 @@
...
+
+1. `npm install --save-dev eslint-config-airbnb-base eslint-plugin-import eslint`
+2. add `"extends": "airbnb-base"` to your .eslintrc
+
+### eslint-config-airbnb-base/legacy
+
+Lints ES5 and below. Requires `eslint` and `eslint-plugin-import`.
+
+1. `npm install --save-dev eslint-config-airbnb eslint-plugin-import eslint`
+2. add `"extends": "airbnb/legacy"` to your .eslintrc
```
lencioni: I think you might still have a few things that need to be changed to `airbnb-base` in this file.

ljharb: thanks, fixed

packages/eslint-config-airbnb-base/.eslintrc
ljharb: I mean, it's pretty weird to have a non-standard eslintrc setup, and we're not doing much with most of the rule files being JS - ie, they could just be normal json. i dunno, you'd have to ask @justjake ¯\\\_(ツ)_/¯

packages/eslint-config-airbnb-base/README.md
```diff
@@ -0,0 +1,33 @@
...
+Our default export contains all of our ESLint rules, including ECMAScript 6+ and React. It requires `eslint` and `eslint-plugin-import`.
+
+1. `npm install --save-dev eslint-config-airbnb-base eslint-plugin-import eslint`
+2. add `"extends": "airbnb-base"` to your .eslintrc
+
+### eslint-config-airbnb-base/legacy
+
+Lints ES5 and below. Requires `eslint` and `eslint-plugin-import`.
+
+1. `npm install --save-dev eslint-config-airbnb eslint-plugin-import eslint`
```
lencioni: I think this one needs to be `eslint-config-airbnb-base` as well, right?

ljharb: lol thanks

packages/eslint-config-airbnb-base/README.md
```diff
@@ -0,0 +1,33 @@
+# eslint-config-airbnb-base
+
+[![npm version](https://badge.fury.io/js/eslint-config-airbnb-base.svg)](http://badge.fury.io/js/eslint-config-airbnb-base)
+
+This package provides Airbnb's base JS .eslintrc as an extensible shared config.
```
lencioni: Might it make sense to add a note here to point people somewhere else if they want React rules as well?

ljharb: I don't think that's really needed - they live in the same repo.

billyjanitsch:
@ljharb It's correct that the *plugins* need to be peer deps, but `eslint-config-airbnb-base` can be a regular dep of `eslint-config-airbnb` because it's just a config object.

As a user, I'd definitely prefer this option because the list of deps needed to be installed in parallel with `eslint-config-airbnb` is getting extremely unwieldy (`eslint-config-airbnb eslint-config-airbnb-base eslint-plugin-import eslint-plugin-react eslint-plugin-jsx-a11y eslint`)

ljharb:
that's a really good point. I'll make that change in `eslint-config-airbnb` - I definitely want to minimize the number of peer deps needed.

tleunen:
I wish we could do the same for plugins :/ Doesn't make sense a config cannot come with all its dependencies :/

packages/eslint-config-airbnb-base/index.js
```diff
@@ -0,0 +1,13 @@
...
+  extends: [
+    './legacy',
+    './rules/es6',
+  ].map(require.resolve),
+  parserOptions: {
+    ecmaVersion: 7,
+    sourceType: 'module',
+  },
+  rules: {
+    strict: 2,
```
evocateur: In the original [base.js](https://github.com/airbnb/javascript/pull/829/files#diff-ca246d3b4ba146243647f872c0e7e8d1), `rules` was empty and `parserOptions` did not exist (`ecmaVersion: 7` here makes no sense for a "base").

I would expect a project that extended `airbnb/base` converting to `airbnb-base` to be as simple as shuffling npm dependencies:

```bash
npm ls eslint-config-airbnb
# └── eslint-config-airbnb@7.0.0
cat .eslintrc.json
# { "extends": "airbnb/base" }
./node_modules/.bin/eslint .
# no errors

npm i -D eslint-config-airbnb-base
npm rm -D eslint-config-airbnb eslint-plugin-jsx-a11y eslint-plugin-react

echo '{ "extends": "airbnb-base" }' > .eslintrc.json
./node_modules/.bin/eslint .
# several dozen errors of type "strict" and "global-require"
```

I was heartened to see a base module was extracted without the unfortunate `peerDependencies` on things I wasn't ever going to use (I'm writing a native node v4+ `hapi` server application). I can of course overwrite the offending options, but it really shouldn't be necessary.

```
{
    "extends": "airbnb-base",
    "parserOptions": {
        "ecmaVersion": 6,
        "sourceType": "script"
    },
    "rules": {
        "global-require": 0
    }
}
```

I'm not sure where `global-require` popped up, since it wasn't complaining in `eslint-config-airbnb/base` before...

ljharb: I added a number of new rules, such that it's not simply the same as shuffling dependencies.

Why do you need ecmaVersion 6 instead of 7? The only difference is the exponentiation operator.

At any rate, our configs mirror what we use internally at Airbnb, and we transpile up to ES7.

ljharb: Specifically, this isn't "base", it's "Airbnb base". It matches what we use, minus React.

PR #837, opened 2016-04-19
Add arrow function returning object.

oskarcieslik:
Imo this looks better, than something like:

```javascript
[1, 2, 3].map((number, index) => {
  return {
    index: number
  }
});
```

What do you guys think? :) If you think that's bad style, maybe you should add this example with `//bad` comment?

### DIFF
--- a/README.md
+++ b/README.md
@@ -809,8 +809,6 @@ Other Style Guides
 
     > Why? Syntactic sugar. It reads well when multiple functions are chained together.
 
-    > Why not? If you plan on returning an object.
-
     ```javascript
     // bad
     [1, 2, 3].map(number => {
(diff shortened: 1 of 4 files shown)


README.md
```diff
@@ -826,6 +824,11 @@ Other Style Guides
       const nextNumber = number + 1;
       return `A string containing the ${nextNumber}.`;
     });
+
```
ljharb: please remove this trailing whitespace

ljharb:
Since our linter rules now prevent most developer errors in this scenario, I think this is OK.

Could you rebase it on top of the latest master?

oskarcieslik:
K, removed trailing whitespace & updated with master :)

ljharb:
@oskarcieslik thanks! this PR does contain a merge commit tho - could you rebase it on top of latest master, and force push, instead?

Issue #786, opened 2016-03-10
Stateless example seems wrong?


jcreamer898:
I noticed in the React guide it says to use regular functions because of the Function.name thing.

However, I noticed that when using Babel, it seems to output a function with the same name as the variable. So therefore wouldn't it in fact have a Function.name?

![](http://d.pr/i/1cvPJ+)

See what I mean?

![](http://d.pr/i/1jESi+)

Unless I missed something... :)

@ljharb I noticed you committed this bit. Thoughts?

ljharb:
@jcreamer898 babel is not spec compliant in this regard. We prefer to rely on standard behavior, not babel quirks.

jcreamer898:
Ahhhh. Ok great. Thanks for clarifying. Closing the issue!

taion:
@ljharb 

Can you clarify what the spec issue here is? I believe 12.14.4, 1.e.iii provides for function names for anonymous functions: http://www.ecma-international.org/ecma-262/6.0/#sec-assignment-operators-runtime-semantics-evaluation

2ality suggests the same in 1.1 of: http://www.2ality.com/2015/09/function-names-es6.html

ljharb:
@taion yes, but we do not recommend relying on function name inference.

PR #850, opened 2016-04-22
Clarify stateless component function naming

taion:
From https://github.com/airbnb/javascript/issues/786#issuecomment-213493620 – the previous comment is not correct.


taion:
Thanks. This caused a bit of confusion on our side.

PR #852, opened 2016-04-24
Fix README "4.5" bad reduce example

Oliboy50:
## Why
To be clearer about what is "bad" in this snippet compared to the "good" snippet.


ljharb:
Thanks!

PR #876, opened 2016-05-12
Prepare to enable import/prefer-default-export

lencioni:
This was recently added to eslint-plugin-export. It enforces that
modules that only have a single export use a default export instead of a
named export.

I believe that this will get our Travis build back to green.

### DIFF
--- a/packages/eslint-config-airbnb-base/rules/es6.js
+++ b/packages/eslint-config-airbnb-base/rules/es6.js
@@ -203,6 +203,11 @@ module.exports = {
       'newlines-between': 'never',
     }],
 
+    // Require modules with a single export to use a default export
+    // https://github.com/benmosher/eslint-plugin-import/blob/master/docs/rules/prefer-default-export.md
+    // TODO: enable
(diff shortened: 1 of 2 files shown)


packages/eslint-config-airbnb-base/rules/es6.js
```diff
@@ -156,6 +156,9 @@ module.exports = {
       'groups': ['builtin', 'external', 'internal', 'parent', 'sibling', 'index'],
       'newlines-between': 'never',
     }],
+    // Require modules with a single export to use a default export
+    // https://github.com/benmosher/eslint-plugin-import/blob/master/docs/rules/prefer-default-export.md
+    'import/prefer-default-export': 2,
```
ljharb: this makes this a breaking change - i'd prefer to cluster those all together. could we mark this with `// TODO: enable` and set it to `0` for now?

lencioni: Sure thing! Done.

PR #881, opened 2016-05-16
Re-order rules of "base"

akai:
Based on http://eslint.org/docs/rules/


ljharb:
I'm marking this as a breaking change just in case someone is relying on direct requires of the rule files.

packages/eslint-config-airbnb-base/rules/style.js
```diff
@@ -125,6 +144,19 @@ module.exports = {
     // disallow use of the Object constructor
     'no-new-object': 2,
 
+    // disallow use of unary operators, ++ and --
+    'no-plusplus': 0
```
ljharb: this is missing a comma after it

ljharb:
This LGTM - before I merge it, I want to cut a release of "base", and then I'll want this rebased, and then this can be merged :-)

akai:
Maybe we should migrate import's rules/settings to a separate config

ljharb:
That's probably a good idea actually, even though `import` is ES6. Let's do that in this PR - `rules/imports` or something?

packages/eslint-config-airbnb-base/rules/es6.js
```diff
@@ -10,9 +10,6 @@ module.exports = {
       'objectLiteralDuplicateProperties': false
```
akai: I can't find these options [here](http://eslint.org/docs/user-guide/configuring#specifying-parser-options). Is it useful?

ljharb: It's documented [here](http://eslint.org/docs/user-guide/migrating-to-2.0.0#language-options). Still useful :-)

PR #882, opened 2016-05-16
Enable imports/imports-first rule

lencioni:
Since imports are hoisted, keeping them all at the top prevents
surprising behavior.

### DIFF
--- a/packages/eslint-config-airbnb-base/rules/imports.js
+++ b/packages/eslint-config-airbnb-base/rules/imports.js
@@ -86,8 +86,7 @@ module.exports = {
 
     // disallow non-import statements appearing before import statements
     // https://github.com/benmosher/eslint-plugin-import/blob/master/docs/rules/imports-first.md
-    // TODO: enable?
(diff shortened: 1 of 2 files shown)


goatslacker:
LGTM

ljharb:
The import rules are now in a separate file - this will need a rebase.

lencioni:
Rebased!

packages/eslint-config-airbnb-base/rules/imports.js
```diff
@@ -86,7 +86,7 @@ module.exports = {
 
     // disallow non-import statements appearing before import statements
     // https://github.com/benmosher/eslint-plugin-import/blob/master/docs/rules/imports-first.md
-    // TODO: enable?
+    // TODO: enable
```
ljharb: actually since this is changing the guide, we should just go ahead and enable it now - i've got some breaking changes queued up anyway

lencioni: done!

PR #886, opened 2016-05-17
Fix deprecation warning of 'quotes' rule

mcdado:
According to [eslint docs](http://eslint.org/docs/rules/quotes) the previous 'avoid-escape' option is deprecated, replaced by an object that includes 'allowTemplateLiterals'.


ljharb:
The change to the object form of the option is great - however, the change to `allowTemplateLiterals` would allow ```var foo = `bar`;```, which according to our guide would need to be `var foo = 'bar';`

mcdado:
I started checking out eslint today, my bad for not checking the style guide ;)
I had closed the PR because I wasn't sure if I did something right :)

PR #893, opened 2016-05-25
Add DoSomething.org to the list of organizations in README

sergii-tkachenko:
Hi!

Here at [DoSomething](https://github.com/DoSomething) we appreciate your JS Style Guide a lot! We use it in our interface framework [Forge](http://forge.dosomething.org/) and in site [theme](https://github.com/DoSomething/phoenix/tree/dev/lib/themes/dosomething/paraneue_dosomething).

It's also the default choice for all new JS ES6 projects we start.

With best regards,
Sergii Tkachenko


sergii-tkachenko:
Thank you!

PR #894, opened 2016-05-25
Avoid using array index as `key` prop

ascott:
plz review @alvinsng @spikebrehm

### DIFF
--- a/react/README.md
+++ b/react/README.md
@@ -279,6 +279,26 @@
   <div />
   ```
 
+  - Avoid using an array index as `key` prop, prefer a unique ID. ([why?](https://medium.com/@robinpokorny/index-as-a-key-is-an-anti-pattern-e0349aece318))
+
+  ```jsx
+  // bad
+  {todos.map((todo, index) =>
+    <Todo
+      {...todo}
+      key={index}
+    />
+  )}
+
+  // good
+  {todos.map((todo) =>
+    <Todo
+      {...todo}
+      key={todo.id}
+    />
+  )}
+  ```
+
 ## Parentheses
 
   - Wrap JSX tags in parentheses when they span more than one line. eslint: [`react/wrap-multilines`](https://github.com/yannickcr/eslint-plugin-react/blob/master/docs/rules/wrap-multilines.md)


react/README.md
```diff
@@ -279,6 +279,22 @@
   <div />
   ```
 
+- Avoid using array index as `key` prop. ([why?](https://medium.com/@robinpokorny/index-as-a-key-is-an-anti-pattern-e0349aece318))
```
lencioni: What do you think about phrasing this more generally, in a way that makes it clear what kinds of keys are good to use?

react/README.md
```diff
@@ -279,6 +279,22 @@
...
+
+```jsx
```
spikebrehm: The alignment is off compared to the other items in the list. I think the way they keep it indented is to add a `> ` block before the code sample.

screen shot 2016-05-25 at 12 11 29 pm

react/README.md
```diff
@@ -279,6 +279,22 @@
...
+// bad
+{todos.map((todo, index) =>
+  <Todo {...todo}
```
spikebrehm: Let's have these examples conform to our prop formatting rules:

```jsx
// bad
{todos.map((todo, index => 
  <Todo
    {...todo}
    key={index}
  />
}
```

spikebrehm:
Thanks for this! Content LGTM, but there's some formatting suggestions.

react/README.md
lencioni: I don't think the indentation is related to the `>` blockquote. I think it is just the codeblock needs to be indented.

react/README.md
ljharb: also, "array indexes" or "array indices"?

ascott: yeah good point. what about:

`Avoid using array index as `key` prop, prefer a unique id.`

ljharb: `s/id/ID` and sounds great!

alvinsng: Use unique element ids and not array indexes or other counter indexes as React may display elements out of order. See this article for why.

alvinsng:
Thanks!

ascott:
fixed the formatting: 
screen shot 2016-05-25 at 12 23 54 pm

ljharb:
@ascott looks great! any chance you'd mind rebasing down to one commit? :-)

ascott:
will do @ljharb

goatslacker:
Can we get a lint rule for this? :)

ljharb:
There's https://github.com/yannickcr/eslint-plugin-react/blob/master/docs/rules/jsx-key.md but I don't know how reliably a lint rule would be that checked for "using the second param to a map callback" or "using something called i or index". Please do file an issue on `eslint-plugin-react` if you think one could still be written.

PR #895, opened 2016-05-25
allow wildcard to make redux consts easily available

alvinsng:
A pattern seen in redux examples like https://github.com/reactjs/redux/blob/master/examples/todomvc/actions/index.js#L1is to use wildcard when importing a file that contains a bunch of exported constants.

This was originally added by @goatslacker in https://github.com/airbnb/javascript/pull/276 before redux was adapted.

cc @kesne @ljharb

### DIFF
--- a/README.md
+++ b/README.md
@@ -1088,19 +1088,6 @@ Other Style Guides
     export default es6;
     ```
 
-  <a name="modules--no-wildcard"></a><a name="10.2"></a>
-  - [10.2](#modules--no-wildcard) Do not use wildcard imports.
-
-    > Why? This makes sure you have a single default export.
-
-    ```javascript
-    // bad
-    import * as AirbnbStyleGuide from './AirbnbStyleGuide';
-
-    // good
-    import AirbnbStyleGuide from './AirbnbStyleGuide';
-    ```
-
   <a name="modules--no-export-from-import"></a><a name="10.3"></a>
   - [10.3](#modules--no-export-from-import) And do not export directly from an import.
 


ljharb:
I'm not sure this should be added. It's still a very bad pattern overall, even if in the one case for redux, it *might* be acceptable.

Rather than removing this important section, we should leave this section intact, and add a section in the React styleguide instead that explicitly says it's only OK for Redux.

ljharb:
(even in Redux, I still think it's better to explicit import all items by name, rather than pulling in a bag)

kesne:
I personally agree that named imports are better. They can be statically analyzed by eslint to ensure that you're not using an ActionType that doesn't actually exist.

mereskin:
My 5ç:

I think `import * from '../constants'` can be useful to import constants under namespace (`types` in the [provided example](https://github.com/reactjs/redux/blob/master/examples/todomvc/actions/index.js#L1is)).

Maybe we can reword the rule along the lines of:
```
Prefer explicit imports to wildcard imports.
<example>

Only use wildcard imports for constants and only if you are going 
to use all (or most) of them.
```

ljharb:
To elaborate on why I think `import * as` is bad: when you're solely using it as a convenience mechanism to avoid specifying every named import manually - ie, when it's always trivially refactorable back to using explicit imports - then clearly it's fine.

The problem arises, when it's in object form, that people will start passing `Types`, say, around to things - leading to implicit dependencies on the constants that are hard to trace - or worse, when people use `Types[foo]` to do dynamic lookup of constant names, which harms greppability and readability.

If there's a way we can enforce via linter that `import * as` is only being used for dot notation lookups, and the object bag is never passed anywhere, then I'm totally fine with allowing this use case. However, absent that, I think the dangers far outweigh the minor convenience of not having to specify constant names twice in the file.

alvinsng:
@ljharb 

I think the problem is not just declaring it twice, it needs to be declared twice in many files. When you have more code than necessary, it makes the code more complex and harder to maintain.

I do agree there are drawbacks with folks abusing the wildcards to do crazy things but I don't think having an absolute ban on wildcards should be limiting us from the good use cases of it.

lencioni:
I'm not that familiar with Redux, but is there any reason you couldn't refactor your module to export an object with all of the constants in it instead of separate exports? If you are interested in reducing verbosity, this seems less verbose than what @mereskin proposed while maintaining great clarity. Example:

```js
-- constants/fooConstants.js

export {
  A: 1,
  B: 2,
};

-- constants/barConstants.js

export {
  A: 3,
  B: 4,
};

-- some/consumer.js

import fooConstants from 'constants/fooConstants';
import barConstants from 'constants/barConstants';

// fooConstants.A !== barCostants.A
```

lencioni:
> When you have more code than necessary, it makes the code more complex and harder to maintain.

It really depends on how you define "more code than necessary". I think that less code _can_ be less complex and easier to maintain, but oftentimes this may not actually be the case. Readability should be a primary goal when writing code, and in my experience, a little extra verbosity and explicitness for the sake of reducing unnecessary indirection can make code more readable which makes it easier to maintain.

alvinsng:
we totally could export it all as a single object. If that is cool with everyone then I'll do that than named exports for each const.

alvinsng:
It sounds like we'll go with exporting a consts file as a single object. Closing PR

PR #899, opened 2016-06-01
Fix react dev dep version

SimenB:
Doesn't really matter I suppose (as `15` is covered), but looks weird.


ljharb:
Thanks, good catch!

PR #900, opened 2016-06-01
fix sentence structure

petertdinh:


ljharb:
The existing structure isn't wrong - what makes this change better?

petertdinh:
You're right, I expected a list to normally follow a colon. Upon further [research](http://www.thepunctuationguide.com/colon.html), this isn't the case.

PR #901, opened 2016-06-02
Fix invalid code snippet in 23.2.

rkirsling:
§[23.2](https://github.com/airbnb/javascript#accessors--no-getters-setters) proscribes the use of JS [getters](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Functions/get)/[setters](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Functions/get), but the code snippet shows something entirely different.

The existing snippet suggests against the following, which doesn't actually use `get`/`set`:
```js
dragon.age();
dragon.age(25);
```
With `get`/`set`, of course, the usage standpoint would be indistinguishable from a plain-old property:
```js
dragon.age;
dragon.age = 25;
```

As such, this PR replaces the existing snippet with an example from the `class` definition standpoint, which demonstrates what was actually intended here.

### DIFF
--- a/README.md
+++ b/README.md
@@ -2502,16 +2502,26 @@ Other Style Guides
 
     ```javascript
     // bad
-    dragon.age();
-
+    class Dragon {
+      get age() {
+        // ...
+      }
+    
+      set age(value) {
+        // ...
+      }
+    }
+    
     // good
-    dragon.getAge();
-
-    // bad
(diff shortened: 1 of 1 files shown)


ljharb:
yeah this is much better, thanks!

PR #903, opened 2016-06-03
Style to avoid issues with events

pedrohenriquerls:


README.md
```diff
@@ -2661,6 +2661,23 @@ Other Style Guides
     $sidebar.find('ul').hide();
     ```
 
+  <a name="jquery--attach-events"></a><a name="25.5"></a>
+  - [25.5](#jquery--attach-events) Avoid mess up with the events queue.
```
ljharb: can you elaborate on this a bit more? Is the advice specifically to ignore "off"?

README.md
```diff
@@ -2661,6 +2661,23 @@ Other Style Guides
...
+
+    ```javascript
+    // bad
+    $('.sidebar').off('click').on('click', (e) => {
+     console.log('dont do this again!!');
+    });
+
+    // good
+    $('.sidebar').on('click', (e) => {
+     e.stopPropagation();
```
ljharb: generally speaking, stopping propagation is an impolite thing to the rest of the DOM, and should not be done lightly. Can you elaborate on this a bit?

README.md
```diff
@@ -2661,6 +2661,23 @@ Other Style Guides
...
+     console.log('well done dude!! :)');
```
ljharb: let's please avoid potentially gendered words like "dude"

pedrohenriquerls: ok, I will fix this

README.md
pedrohenriquerls: yes, the idea is avoid do this on event attach, I will improve the text

README.md
pedrohenriquerls: ok, I will do this thanks for the feedback :+1:

pedrohenriquerls:
@ljharb I improve the explanation about the new style, like you requested :)
[changes](https://github.com/airbnb/javascript/pull/903/files)

ljharb:
Thanks. I'm not sure this is a pattern that happens frequently enough to need commenting on. Certainly calling `off` with no arguments is a destructive action, but it both might be intended, and, I don't see it commonly occurring in jQuery codebases, now or throughout my career. In addition, Airbnb now uses React, so we don't tend to need jQuery for event handling any more, so this is even less likely to come up in the future.

Thanks for the contribution! I'm going to close this for now.

PR #910, opened 2016-06-11
add missing return to bad example 8.2

psimyn:
to make the bad/good examples do the same thing.

unless it was intentionally missing as part of being bad

### DIFF
--- a/README.md
+++ b/README.md
@@ -833,7 +833,7 @@ Other Style Guides
     // bad
     [1, 2, 3].map(number => {
       const nextNumber = number + 1;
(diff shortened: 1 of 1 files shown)


ljharb:
Yes, it was intentionally part of being bad - there's nothing bad about the example if we incorporate your fix.

psimyn:
I thought the example was 'no parens for implicit returns' (bad because it has none and is multiline)

ljharb:
The example deals with implicit returns in general, and the idea with this one is to point out the hazard of attempting to use implicit return inside curly braces.

PR #922, opened 2016-06-17
Remove parens around single argument in React example

staticshock:
Parens around a single argument aren't in line with the [javascript guide](https://github.com/airbnb/javascript/blob/0814be638f22b01a8e49efd77094be24055a3d48/README.md#arrows--one-arg-parens):

> If your function takes a single argument and doesn’t use braces, omit the parentheses.

### DIFF
--- a/react/README.md
+++ b/react/README.md
@@ -291,12 +291,12 @@
   )}
 
   // good
-  {todos.map((todo) =>
+  {todos.map(todo => (
     <Todo
       {...todo}
       key={todo.id}
     />
-  )}
+  ))}
   ```
 
 ## Parentheses


react/README.md
```diff
@@ -291,7 +291,7 @@
   )}
 
   // good
-  {todos.map((todo) =>
+  {todos.map(todo =>
```
ljharb: thanks! this example is actually also violating another rule, that multiline jsx must be wrapped with parens, ie `=> (` and `))}` - could you make that fix and amend this commit and force push?

ljharb:
Thanks! The test failures are due to an eslint bug, so I'll merge through it.

PR #925, opened 2016-06-20
fix defaults for import/no-extraneous-dependencies

billyjanitsch:
These should both [default to true](https://github.com/benmosher/eslint-plugin-import/blob/master/docs/rules/no-extraneous-dependencies.md). The rule is disabled so this doesn't really matter, but if you override it with `'import/no-extraneous-dependencies': 2`, now you'll get the expected defaults.

### DIFF
--- a/packages/eslint-config-airbnb-base/rules/imports.js
+++ b/packages/eslint-config-airbnb-base/rules/imports.js
@@ -60,8 +60,8 @@ module.exports = {
     // https://github.com/benmosher/eslint-plugin-import/blob/master/docs/rules/no-extraneous-dependencies.md
     // TODO: enable
     'import/no-extraneous-dependencies': [0, {
-      devDependencies: false,
-      optionalDependencies: false,
+      devDependencies: true,
+      optionalDependencies: true,
     }],
 
     // Forbid mutable exports


ljharb:
They are intentionally false, because you should not be importing anything except `dependencies` in your app. In your test directory, you'd probably want an `.eslintrc` file that enables `devDependencies`.

`optionalDependencies` and the requisite lazy loading are an antipattern anyways, so if you need that in your project, it's appropriate to override it in your own project.

I'll be enabling this rule soon as well.

billyjanitsch:
I see -- this wasn't a comment on the correctness of the rule; I just thought the convention in this repo was to copy the default options for disabled rules, so I assumed this was an omission. (I was surprised when I enabled this rule in my own repo with `2` and didn't get the default options.)

FWIW, the use case for importing devDependencies in an app folder is for apps which are bundled by webpack and shipped via the registry. Since dependencies are inlined at bundling time, you don't want the consumer to reinstall them. I agree that this shouldn't be the default option since it's limited to web apps.

ljharb:
@billyjanitsch devDependencies should never be shipped via the registry. They're for *developing the project*. Anything required at runtime - ie, shipped via the registry - belongs as a proper dependency, never a dev dep.

billyjanitsch:
In general, I agree, especially wrt packages in the public registry. I'm speaking specifically of the case where the package is meant to be consumed as a browser-ready bundle, which is common for internal packages. In that case, dependencies *must* be inlined, so they aren't required at runtime. Even if you wouldn't characterize these as "required only for developing the project" (ambiguous imo), pragmatically it's unnecessary for consumers to install them.

Anyway, that's all incidental to this PR. Thanks for clarifying the intent of disabled rules.

ljharb:
They must be inlined but they still are conceptually a dependency, not a dev dependency. In addition, a package with a browser-ready bundle should still be consumable via node, which would require the deps to be installed.

No problem, happy to clarify :-)

leeyeh:
>  In your test directory, you'd probably want an .eslintrc file that enables devDependencies.

Will be helpful if mentioned in the document.

ljharb:
@leeyeh that's a great point - I'd love to accept a PR improving the text.

kimmobrunfeldt:
How have you configured linting for files located in project root? As an example we have a gulpfile.js in the root of our frontend project and it is used to build our bundle.js which is the actual shippable end result. Gulp creates that bundle and it belongs to devDevepencies but the linter config makes that an error. One solution could be adding `.eslintrc` inside our ./src folder which has this rule enabled, and another .eslintrc in the project root which allows importing devDependencies. To my mind that sounds like the most correct approach as tests are not the only supporting code for the code base you might have.

ljharb:
@kimmobrunfeldt in that case, the risk of turning off important rules in the root of your repo isn't worth it. I'd recommend inline config comments there, in the Gulpfile only.

ljharb:
`eslint-config-airbnb` `v13.0.0` is now released.

PR #928, opened 2016-06-23
Minor typo: "coxcomb graph" v/s "coxcomb chart"

shreevatsa:
The "bad" and "good" diffs in section 19.2 ("Additional trailing comma ") had different text; making them the same to avoid confusion about what the difference is.


ljharb:
Thanks! I doubt anyone would ever be confused by that but we might as well make the text the same.

PR #929, opened 2016-06-24
Update README.md

gemscng:
fix good switch


ljharb:
Thanks for your contribution!

These brackets aren't necessary, however, because the case body doesn't contain any block-scoped variables. The example is specifically meant to indicate that the curly braces are OK to omit in this case.

PR #931, opened 2016-06-26
Change Batman to Superman

steveosneu:
Not sure if this is how to do this.  Please reconsider.  Superman is much better fit for this style guide.  I beg you merge.

### DIFF
--- a/README.md
+++ b/README.md
@@ -524,17 +524,17 @@ Other Style Guides
 
     ```javascript
     // bad
-    const errorMessage = 'This is a super long error that was thrown because of Batman. When you stop to think about how Batman had anything to do with this, you would get nowhere fast.';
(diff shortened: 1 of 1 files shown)


ljharb:
Superman is just being his normal Kryptonian self. Batman actually had to *work* to be a superhero.

Of course, both of them think the way to solve societal problems is to go around punching individual criminals.

In what way is Superman a better fit?

steveosneu:
http://youtu.be/uqiMY7nIBbI

Thanks.  Please watch this.  It convinced me.

ljharb:
http://youtu.be/UeEQAvkU5x0 is a pretty convincing counterpoint.

goatslacker:
This change is one that the JavaScript community deserves, but not the one it needs right now. So we'll close this. Because GitHub can take it. Because it's not our changeset.

hshoff:
```js
[Array(8).join('Superman' - 1),'Batman!'].join(' ');
```

PR #935, opened 2016-07-01
Update README.md

filso:
Previous description seems a bit confusing as it suggests always using const. I'm guessing that's a result of literal translation of document from "var" in ES5


ljharb:
This is intentional - **always** use `const`. Avoid variable reassignment. When you must do it (try not to), use `let`.

Issue #936, opened 2016-07-02
Allow peerDependency eslint@3.0.0


nutgaard:

ljharb:
This is blocked by:
 - https://github.com/benmosher/eslint-plugin-import/pull/404
 - the release of https://github.com/sarbbottam/eslint-find-rules/pull/119
 - https://github.com/yannickcr/eslint-plugin-react/pull/657

ljharb:
The base config is now published at v4.0.0. The main config is still blocked on a release of eslint-plugin-react (@yannickcr)

geminiyellow:
hi @ljharb ,

Does that mean that can be upgraded to support eslint3?

```
Merge pull request #657 from jokeyrhyme/bump-deps-eslint-3  …
ljharb committed on GitHub 11 hours ago
```

ljharb:
@geminiyellow no, because the update is not yet released (see the immediately preceding two comments).

roderik:
A release candidate of the react plugin was released? Maybe a beta release for this project?

ljharb:
I'm not a fan of rc's or alphas. I'm going to wait until the official release comes out - hopefully it will be soon, however.

nwhitmont:
I just tried to install the package https://www.npmjs.com/package/eslint-config-airbnb but it failed with this error FYI

merlinpatt:
v6.0.0 will be released in the coming hours according to the most recent comment on that thread

merlinpatt:
And it just got released!!!

ljharb:
v10.0.0 of [`eslint-config-airbnb`](https://www.npmjs.com/package/eslint-config-airbnb) is now published.

PR #939, opened 2016-07-05
Update for ESLint 3.0.0

lencioni:
ESLint 3.0.0 was recently released. I don't see any breaking changes
that affect our configuration, so it seems like we can support the
currently supported range in addition to 3.0.0+.

http://eslint.org/blog/2016/07/eslint-v3.0.0-released

I needed to fix a couple of minor things to get tests to pass.

### DIFF
--- a/packages/eslint-config-airbnb-base/package.json
+++ b/packages/eslint-config-airbnb-base/package.json
@@ -46,14 +46,14 @@
   "homepage": "https://github.com/airbnb/javascript",
   "devDependencies": {
     "babel-tape-runner": "^1.3.1",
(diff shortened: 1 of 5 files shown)


ljharb:
:x: this is captured in #936, and has peer dep blockages. In addition, I already have this change prepared locally.

lencioni:
Ah rats, I missed that entirely.

Are my first two commits still useful?

ljharb:
I'll definitely grab 2d387043b06fefed9789751a151324052f392230 - the other one I've already got. Thanks!

PR #942, opened 2016-07-07
Added link to Mozilla documentation on Function#bind

LukeL99:


ljharb:
Thanks, I'd hope that JS devs would be familiar with something that's been around since 2009, but links can't hurt!

LukeL99:
Just matching the style of Array#push in 4.2 which I would hope people would be even more familiar with. 😄  Thanks for the merge!

PR #943, opened 2016-07-08
Removed unnecessary text

khan4019:


ljharb:
I don't think whimsy is unnecessary.

PR #946, opened 2016-07-09
Changed ESLint rule configs to use 'off', 'warn', and 'error' instead of numbers for better readability

yangsu:
I wrote [a JsCodeShift script](https://gist.github.com/yangsu/f778e467ba1f2a75b6c27aa09bed8e73) to accomplish this.

### DIFF
--- a/packages/eslint-config-airbnb-base/index.js
+++ b/packages/eslint-config-airbnb-base/index.js
@@ -13,6 +13,6 @@ module.exports = {
     sourceType: 'module',
   },
   rules: {
-    strict: 2,
+    strict: 'error',
   }
 };
--- a/packages/eslint-config-airbnb-base/legacy.js
+++ b/packages/eslint-config-airbnb-base/legacy.js
@@ -16,6 +16,6 @@ module.exports = {
   ecmaFeatures: {},
   globals: {},
   rules: {
-    'comma-dangle': [2, 'never']
+    'comma-dangle': ['error', 'never']
   }
 };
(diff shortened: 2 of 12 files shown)


ljharb:
I prefer sticking with the historical 0/1/2 notation of jslint/jshint/eslint.

yangsu:
The main reason for me submitting this pull request is that the latest documentation for eslint rules all use the string enums rather than numbers.

ljharb:
Understood, and that's a fair point.

I'll leave this open for the time being so other contributors can weigh in (heads up, it'll likely need to be rebased and reran before merging, if it comes to that)

kesne:
I'm 👍🏻 to this change. It seems to me as though it makes the configuration more readable, which is what we should always strive for.

lencioni:
I also see this as a positive change, for the same reasons already stated.

yangsu:
I rebased and reran the codeshift script.

KooBiGitHub:
Đã gửi từ iPhone của tôi

Ngày 13-07-2016, vào lúc 11:58, Yang Su <notifications@github.com> viết:

hshoff:
Looks reasonable to me

yangsu:
Any other suggestions/thoughts before this can be merged? Otherwise I'll have to keep rebasing and re running the codeshift

ljharb:
At the very least, this needs to wait until after #936 is resolved, since I have those commits queued up and don't want the conflict.

I still do not think this is a good change, fwiw, but enough contributors have weighed in that I'm ok accepting it once the new release is out.

AnilRedshift:
+1 to this change: as more time goes forward, more code is just going to be using "off", "warn", and "error". Newer folks (like myself), then have to go back and look back to see what the old numbers mean.

ljharb:
@yangsu if you'd like to rebase this one last time, i'll merge it.

yangsu:
Done

PR #953, opened 2016-07-15
Capitalize all comments in eslint react.js

kennyt:


ljharb:
I don't think we need to be that pedantic about comments, but sure, why not.

KooBiGitHub:
Von meinem iPhone gesendet

PR #970, opened 2016-07-27
Update README.md

jglover:
Missing semicolon breaks `$PKG` reference

### DIFF
--- a/packages/eslint-config-airbnb/README.md
+++ b/packages/eslint-config-airbnb/README.md
@@ -12,7 +12,10 @@ We export three ESLint configurations for your usage.
 
 Our default export contains all of our ESLint rules, including ECMAScript 6+ and React. It requires `eslint`, `eslint-plugin-import`, `eslint-plugin-react`, and `eslint-plugin-jsx-a11y`.
 
-1. `PKG=eslint-config-airbnb; npm info "$PKG" peerDependencies --json | command sed 's/[\{\},]//g ; s/: /@/g' | xargs npm install --save-dev "$PKG"` (which produces and runs a command like `npm install --save-dev eslint-config-airbnb eslint@^2.9.0 eslint-plugin-jsx-a11y@^1.2.0 eslint-plugin-import@^1.7.0 eslint-plugin-react@^5.0.1` but with whatever the proper version numbers are)
+1. `(
+  export PKG=eslint-config-airbnb;
+  npm info "$PKG" peerDependencies --json | command sed 's/[\{\},]//g ; s/: /@/g' | xargs npm install --save-dev "$PKG"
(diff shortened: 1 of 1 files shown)


ljharb:
The idea is that the variable is set solely for the duration of the command - the semicolon wasn't necessary in my use of it.

lencioni:
Indeed, I typically don't need a semicolon for setting these environment variables either. @jglover what shell are you using?

jglover:
Omitting it on osx El Capitan (ZSH) led to it trying to use the project name in place of the value.

lencioni:
I also use zsh on OS X. I just tried this out and confirmed that it doesn't work without the semicolon but it does with the semicolon.

jglover:
Interesting. Looks like ZSH behaves strangely. I also expected it to work without *shrug*

ljharb:
`zsh` does all kinds of strange things so I'd prefer to leave it broken for the time being, and figure out how to rewrite the command so it doesn't pollute the shell session with a new environment variable, and to have the same command work on all shells.

Alternatively, we could leave it as-is and just document that it should be run in `bash`.

lencioni:
I ran this through ShellCheck and got the following errors:

https://github.com/koalaman/shellcheck/wiki/SC2097
https://github.com/koalaman/shellcheck/wiki/SC2098

So, I think we should use a subshell:

```sh
```

ljharb:
Great, that seems to work fine for me as well. Let's update it to use that.

jglover:
Makes more sense, seems like the safer option.

ljharb:
@jglover update in-place please :-) it would also be ideal to use git locally and rebase it down to one commit, rather than using the web UI. (Also, it still needs to update the `eslint-config-airbnb-base` README)

packages/eslint-config-airbnb/README.md
```diff
@@ -12,7 +12,10 @@ We export three ESLint configurations for your usage.
 
 Our default export contains all of our ESLint rules, including ECMAScript 6+ and React. It requires `eslint`, `eslint-plugin-import`, `eslint-plugin-react`, and `eslint-plugin-jsx-a11y`.
 
-1. `PKG=eslint-config-airbnb npm info "$PKG" peerDependencies --json | command sed 's/[\{\},]//g ; s/: /@/g' | xargs npm install --save-dev "$PKG"` (which produces and runs a command like `npm install --save-dev eslint-config-airbnb eslint@^2.9.0 eslint-plugin-jsx-a11y@^1.2.0 eslint-plugin-import@^1.7.0 eslint-plugin-react@^5.0.1` but with whatever the proper version numbers are)
+1. `(
```
ljharb: this should probably use three backticks with "bash", so that it's a proper multiline section.

jglover:
@ljharb will do :)

jglover:
Closing due to re-forking repo (can't edit source because original repo is gone). Will update with new PR link.

jglover:
Updated in #971

PR #971, opened 2016-07-27
Updated README for eslint-config-airbnb & eslint-config-airbnb-base

jglover:
Relates to #970 
Added subshell to usage commands and converted samples to the three backtick markdown syntax with language specifier for highlighting. 

This fixes an issue where ZSH (and potentially other shells) don't execute the commands properly because of a missing semicolon after variable declaration,

Had to change some of the wording because if you embed a code block under a numbered list item without text on the same line as the number, markdown gets upset and drops the list formatting.


ljharb:
Thanks!

PR #972, opened 2016-07-27
Improve React HOC example

lencioni:
A typo was spotted.

I also decided that `WrappedComponent` is clearer than `Component` here, so I
made the switch. I also realized that `WrappedComponent.name` might
still be undefined, so I added a fallback value of "Component".


pepakriz:
Nice example, thank you! :+1:

PR #975, opened 2016-07-30
Prepare to switch to eslint-plugin-react 6

akozhemiakin:
This pull request is applicable to eslint-plugin-react@^6.0.0-rc.0. So I think it will be a good idea
to keep it until switching to that version and update it along with new breaking changes in react plugin.

"pragma" parameter from "jsx-uses-react" and "react" parameter from "no-deprecated" must be removed because they were deprecated in eslint-plugin-react since version 5.* and removed in 6.0.0-rc.0. Also it is already specified in the shared section (settings.react.pragma) as it should be, so nothing to fix there.

"allow-in-func" parameter must be removed from "no-did-mount-set-state" and "no-did-update-set-state" because since 6.0.0-rc.0 it is the default behavior and "allow-in-func" parameter was removed.

"wrap-multilnes" was deprecated and "jsx-wrap-multilines" must be used instead.

### DIFF
--- a/packages/eslint-config-airbnb/rules/react.js
+++ b/packages/eslint-config-airbnb/rules/react.js
@@ -103,7 +103,7 @@ module.exports = {
 
     // Prevent React to be incorrectly marked as unused
     // https://github.com/yannickcr/eslint-plugin-react/blob/master/docs/rules/jsx-uses-react.md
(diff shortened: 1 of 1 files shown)


ljharb:
Thanks for the contribution, but I already have all the changes required to switch to eslint-plugin-react v6 locally. I do *not* want to switch to an rc version prior to v6 being released.

PR #984, opened 2016-08-02
Update eslint-config-airbnb README

jongold:
#936 changed some dependency versions


ljharb:
The intent of adding the automated instructions was so we wouldn't have to keep bumping version numbers in the readme. Unless we can make a prepublish script that updates this automatically, I don't think it's worth updating it manually.

lencioni:
Should we replace the numbers with `#`?

ljharb:
That's probably a good idea until we have an automated way to update it.

PR #989, opened 2016-08-03
[guide] Add eslint rules to function declaration and function constructor topics

felipethome:
I believe these ESlint rules are able to cover the 7.1 and 7.10 topics. `func-style: ["error", "declaration"]` forces function declarations and `no-new-func: "error"` forces the developer to not use the Function constructor.


ljharb:
Thanks!

PR #991, opened 2016-08-03
Remove rules about reserved words

### LINKED ISSUE #61: Allow reserved words for properties?
Under [Objects](https://github.com/airbnb/javascript#objects) it says you shouldn't use reserved words.  I'm not sure this is important, though avoiding them might make sense for readability.  At least using `klass` instead of `class` would seem wrong to me, as you are corrupting a word instead of finding a readable synonym (like `type`).

Right now you can use just about anything.  E.g., `test = {function: 1}; test.function === 1` –but additionally I think we can be confident this will continue to be the case.  IndexedDB uses the method `cursor.continue()` which isn't just a future-protected reserved word (like `class`) but a reserved word *right now*.  But IndexedDB represents current API design by people involved in the standardization process.

I think it's questionable from a readability point of view to use current keywords, but I don't see any reason at all to avoid [words reserved for possible future use](https://developer.mozilla.org/en-US/docs/JavaScript/Reference/Reserved_Words#Words_reserved_for_possible_future_use) (`class` and `private`, as used in the example, are both examples of that).

lencioni:
In 53b4173b we removed the ES5 guide which contains a lot of guidelines
that are no longer very relevant for us. Similarly, some folks have
raised the relevance of these rules about reserved words, given that we
are now living in an age where ES3 support has mostly waned and
transpilers such as Babel are widely adopted and pave over these issues.
This seems like a good opportunity to simplify.

Fixes #61

### DIFF
--- a/README.md
+++ b/README.md
@@ -153,43 +153,6 @@ Other Style Guides
     const item = {};
     ```
 
-  <a name="objects--reserved-words"></a><a name="3.2"></a>
-  - [3.2](#objects--reserved-words) If your code will be executed in browsers in script context, don't use [reserved words](http://es5.github.io/#x7.6.1) as keys. It won't work in IE8. [More info](https://github.com/airbnb/javascript/issues/61). It’s OK to use them in ES6 modules and server-side code. jscs: [`disallowIdentifierNames`](http://jscs.info/rule/disallowIdentifierNames)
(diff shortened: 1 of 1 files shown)


ljharb:
Agreed, thanks.

PR #994, opened 2016-08-04
Add template literal "bad" example to single quotes guideline

### LINKED ISSUE #992: About quotes in strings
https://github.com/airbnb/javascript#strings--quotes

I see there:
```
// bad
const name = "Capt. Janeway";

// good
const name = 'Capt. Janeway';
```
But below I see reversed qoutes:
https://github.com/airbnb/javascript#es6-template-literals
```
// good
function sayHi(name) {
  return `How are you, ${name}?`;
}
```
What about adding `` in example? I mean sth like that
```
// bad
const name = "Capt. Janeway";

// good
const name = 'Capt. Janeway';

// also good
const name = `Capt. Janeway`;
```

lencioni:
There has been some confusion around whether we should use single quotes
or template literals. To help avoid this confusion, I am adding a "bad"
example to the single quotes guideline. This rule is already enforced by
the quotes linter rule.

Generally, we want code to clearly express developer intention. Using
template literals communicates that you intend to use some of the
features that template literals offer (e.g. interpolation).

Fixes #992

### DIFF
--- a/README.md
+++ b/README.md
@@ -475,6 +475,9 @@ Other Style Guides
     // bad
     const name = "Capt. Janeway";
 
+    // bad - template literals should contain interpolation or newlines
+    const name = `Capt. Janeway`;
+
     // good
     const name = 'Capt. Janeway';
     ```


README.md
```diff
@@ -475,6 +475,9 @@ Other Style Guides
     // bad
     const name = "Capt. Janeway";
 
+    // bad
```
ljharb: should this elaborate, and say "because template literals should always contain an interpolation"?

lencioni: sure!

lencioni: Although, that's not strictly true. Sometimes you want to use them for newlines I think.

ljharb: that's true - "or newlines" would be a fine addition to my above suggestion

lencioni: done

lencioni: And I guess there's also tagged template literals, which might contain neither interpolation nor newlines.

ljharb: yeah, that's true

Issue #40, opened 2013-01-11
String concat speed


rjacoby:
In the [Strings section](https://github.com/airbnb/javascript#strings) the standard is:

> "Strings longer than 80 characters should be written across multiple lines using string concatenation."

I understand the clarity of code - but the performance is 80-100% worse. [jsPerf](http://jsperf.com/ya-string-concat)

The need for such long strings is debatable anyway, but isn't this kind of a steep performance penalty for code style?

Nemoden:
So, why exactly strings concatenation is preferred by the Guide if it's 95% slower than joining it with the break symbol (\\)? It doesn't add much readability and extremely slow.

![js_perf_long_strings](https://cloud.githubusercontent.com/assets/795034/7037567/3e7dffe2-dde8-11e4-99ce-bcd1aa5996dc.png)

nkbt:
@Nemoden maybe because of readability. When string concatenation becomes a bottleneck in the app, then it can be easily converted to the fastest way available.

joooche:
Ok

yonghu86:
well,it's not length.

michaelmior:
I think concatenation is much more readable for the sole fact that you can keep the correct level of indentation. As I noted in my previous comment, I think concatenation of static strings should be removed by build tools so performance should not really ever be an issue.

ibc:
> I think concatenation is much more readable for the sole fact that you can keep the correct level of indentation.

Or you can use a proper editor that properly indents extra lines if they belong to the same line.

jonasninja:
It seems that the latest versions of Chrome have made performance a moot point. Here's a screenshot of revision 6 results. 
![image](https://cloud.githubusercontent.com/assets/4935079/9977082/02f0af1c-5ec9-11e5-9f79-0deb4bf9cb3c.png)

ljharb:
Nowaways it's preferred to use an ES6 template string literal - which transpiles to a long string (with or without breaks, i forget). This entire issue seems like a moot point now :-)

GrayedFox:
Not quite a moot point.... basically, referencing variables inside of a template literal is still much slower than string concatenation, however without referencing a variable (and simply using a template literal for line breaks) seems to have no performance impact... https://jsperf.com/es6-string-literals-vs-string-concatenation/4

ljharb:
But also, "much slower" won't make any difference unless you're building millions of strings per second. It's identically fast for most use cases.

PR #995, opened 2016-08-04
Reverse rule on string concatenation for long lines

lencioni:
Broken and concatenated long strings are painful to work with and
produce less readable and searchable code. I think we should reverse
this rule.

Unfortunately, the max-len rule currently does not allow for this, but
there is currently a proposal to add an option to ESLint's max-len rule
that would allow for strings to be ignored.

  https://github.com/eslint/eslint/issues/5805

There have also been discussions around performance of string
concatenation (https://github.com/airbnb/javascript/issues/40), but I
don't think that is very relevant here so I removed the links to them.

### DIFF
--- a/README.md
+++ b/README.md
@@ -483,25 +483,24 @@ Other Style Guides
     ```
 
   <a name="strings--line-length"></a><a name="6.2"></a>
-  - [6.2](#strings--line-length) Strings that cause the line to go over 100 characters should be written across multiple lines using string concatenation.
+  - [6.2](#strings--line-length) Strings that cause the line to go over 100 characters should not be written across multiple lines using string concatenation.
 
-  <a name="strings--concat-perf"></a><a name="6.3"></a>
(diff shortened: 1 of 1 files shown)


ljharb:
LGTM

chrisngobanh:
While you bring up a good point about it being less searchable, I disagree with your opinion that it makes it easier to read. I always make sure that my line lengths are within the 100 character limit because it's obnoxious to have to scroll to the right in my text editor to read a line of code.

I propose that we go forth with adding the `ignoreStrings` option to the `max-len` rule when it becomes available, while changing this rule to say something along the lines of "we don't care if you want to concatenate your strings or not as long as you concatenate them correctly if you choose to do so". In other words, we should change this rule to be less strict with how we want the developer to format their strings.

ljharb:
@chrisngobanh the string's contents aren't something that needs to be read - strings are data, a black box. The *code* is what matters.

We'll definitely add that option when it becomes available, but strings broken across lines are very hard to read, so I don't think we should change the guide in that manner.

chrisngobanh:
@ljharb We need to update [18.12](https://github.com/airbnb/javascript#whitespace--max-len)

ljharb:
@chrisngobanh see #1027

Issue #999, opened 2016-08-04
Add topic about spread over apply


felipethome:
Add a topic, maybe in section 7 (Functions), about using spread operator `...` instead of `apply()`. Like:

// bad
var args = [1, 2, 3, 4];
Math.max.apply(Math, args);

// good
var args = [1, 2, 3, 4];
Math.max(...args);

Related ESlint rule: `prefer-spread`

Don know if worth adding this, because my reason would be just: it is cleaner this way.

ljharb:
Sure! Reasons include: it doesn't depend on `Function#apply` existing or working properly in the environment, so it's more robust; it's syntax, so the engine may be able to optimize it more; it's less verbose, so it's less likely to be done wrong.

felipethome:
Nice, thanks for the reasons!

ljharb:
Closed by #1001

PR #1001, opened 2016-08-05
[guide] Add a topic about spread operator vs apply

felipethome:
Related to #999 
I added the topic as 7.14, but maybe this topic should come close to 7.6?

### DIFF
--- a/README.md
+++ b/README.md
@@ -764,6 +764,27 @@ Other Style Guides
     }
     ```
 
+  <a name="functions--spread-vs-apply"></a><a name="7.14"></a>
+  - [7.14](#functions--spread-vs-apply) Prefer the use of the spread operator `...` to call variadic functions. eslint: [`prefer-spread`](http://eslint.org/docs/rules/prefer-spread)
+
+    > Why? It's cleaner, you don't need to supply a context, and you can not easily compose `new` with `apply`.
+
+    ```javascript
+    // bad
+    const x = [[1, 2], [3, 4], [5, 6]];
+    console.log(Array.prototype.concat.apply([], x));
+
+    // good
+    const x = [[1, 2], [3, 4], [5, 6]];
+    console.log([].concat(...x));
+
+    // bad
+    new (Function.prototype.bind.apply(Date, [null, 2016, 08, 05]));
+
+    // good
+    new Date(...[2016, 08, 05]);
+    ```
+
 **[⬆ back to top](#table-of-contents)**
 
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -764,6 +764,27 @@ Other Style Guides
...
+    > Why? It's cleaner, you don't need to supply a context, and you can not easily compose `new` with `apply`.
+
+    ```javascript
+    // bad
+    const x = [[1, 2], [3, 4], [5, 6]];
+    console.log(Array.prototype.concat.apply([], x));
+
+    // good
+    const x = [[1, 2], [3, 4], [5, 6]];
+    console.log([].concat(...x));
```
ljharb: I think this should just be `[...x]`?

ljharb: or maybe just `...x`?

README.md
```diff
@@ -764,6 +764,27 @@ Other Style Guides
...
 
+  <a name="functions--spread-vs-apply"></a><a name="7.14"></a>
+  - [7.14](#functions--spread-vs-apply) Prefer the use of the spread operator `...` to call variadic functions. eslint: [`prefer-spread`](http://eslint.org/docs/rules/prefer-spread)
+
+    > Why? It's cleaner, you don't need to supply a context, and you can not easily compose `new` with `apply`.
+
+    ```javascript
+    // bad
+    const x = [[1, 2], [3, 4], [5, 6]];
+    console.log(Array.prototype.concat.apply([], x));
```
ljharb: and i'd think this would be `console.log.apply(console, x)`

README.md
felipethome: That is because it isn't very clear what I'm trying to do. I was trying to show an example of how to flatten an array. I should point that out with comments, or do you prefer that I change the example to just a simple console.log print?

ljharb: ah, in that case i'd probably make a `flatten` function in both examples?

felipethome: I can do that, but I actually liked the simplicity of just doing a print with console.log() like you suggested. What do you prefer?

ljharb: I'm not sure if either's really necessary. The point of the examples is to demonstrate uses of "apply" that are better used with the spread operator - using concat to flatten seems somewhat obscure :-)

felipethome:
> The point of the examples is to demonstrate uses of "apply" that are better used with the spread operator - using concat to flatten seems somewhat obscure

True! I simplified the examples now.

PR #1010, opened 2016-08-11
Altered a "bad" example that didn't actually work under 8.2.

Gisleburt:
This previously gave the example:

```
[1, 2, 3].map(number => {
});
```
However, in both node 5, and node 6, this example doesn't actually work. It certainly is "bad" but not for the given reason. It also gave me (and perhaps others) the impression that the final statement was implicitly returned when it's not.

```
$  docker run -it --rm node:6 node
> const test = [1, 2, 3].map(number => {
... const nextNumber = number + 1;
... `A string containing the ${nextNumber}.`;
... });
undefined
> test
[ undefined, undefined, undefined ]
>
(To exit, press ^C again or type .exit)
>

$  docker run -it --rm node:5 node
> const test = [1, 2, 3].map(number => {
... const nextNumber = number + 1;
... `A string containing the ${nextNumber}.`;
... });
undefined
> test
[ undefined, undefined, undefined ]
>
(To exit, press ^C again or type .exit)
>
```

I've chosen another example that does work and seems "bad" based on the explanation given in the guide.

```
& docker run -it --rm node:5 node
> let test = [1, 2, 3].map(number => { return `A string containing the ${number}.`; });
undefined
> test
[ 'A string containing the 1.',
  'A string containing the 2.',
  'A string containing the 3.' ]
>
(To exit, press ^C again or type .exit)
>});


$ docker run -it --rm node:6 node
> let test = [1, 2, 3].map(number => { return `A string containing the ${number}.`; });
undefined
> test
[ 'A string containing the 1.',
  'A string containing the 2.',
  'A string containing the 3.' ]
>
(To exit, press ^C again or type .exit)
>});
```

### DIFF
--- a/README.md
+++ b/README.md
@@ -817,10 +817,7 @@ Other Style Guides
 
     ```javascript
     // bad
-    [1, 2, 3].map(number => {
-      const nextNumber = number + 1;
-      `A string containing the ${nextNumber}.`;
-    });
+    [1, 2, 3].map(number => { return `A string containing the ${number}.`; });
 
     // good
     [1, 2, 3].map(number => `A string containing the ${number}.`);


ljharb:
That it doesn't work is the exact reason this example is bad. It's meant to point out the hazards of forgetting the `return` keyword when you're using the explicit-return form.

Gisleburt:
Surely a style guide is about style, not syntax. That's what the language specification is for. 😄 

I would say as it stands, between the example and the explanation, this suggests that the final line is an implicit return. The code is not functional and would be picked up by a developer as such regardless of code style.

ljharb:
Right, but the current example is not invalid syntax - it's just a common mistake developers do make, and have made - that's why it's included.

Gisleburt:
Sorry, by invalid syntax I mean it's not a valid way to get a return from the closure. To me the example says, don't do this because it's confusing, not don't do this because it won't work. That's not a style issue.

Anyway, this mislead me, not in code obviously, in a conversation. Given that it's the adopted defacto style for ES6+, I'm sure I'm not the only one to make the mistake, but in the end it's your style and it's up to you guys.

nathankunicki:
Just chiming in, I feel the the example should be changed.

1. The code won’t work. This isn’t “style”, the code just simply won’t work. Your interpreter or transpiler will pick it up. It's not demonstrating best practices, it's just demonstrating non-functioning code.

2. The code isn’t demonstrating what the section heading is saying anyway.

The heading says: “8.2 If the function body consists of a single expression, omit the braces and use the implicit return. Otherwise, keep the braces and use a return statement.”

The example has two expressions within the braces, therefore the braces cannot be omitted, and so the implicit return cannot be used.

KooBiGitHub:
Đã

Ngày 12-08-2016, vào lúc 01:45, Nathan Kunicki <notifications@github.com> viết:

ljharb:
What's "style" and not isn't really relevant - our style guide recommends plenty of practices to avoid non-working code.

The problem that we are hoping to avoid is that the code *will parse* but will not actually do what programmers expect.

This example parallels the good example two below it. I'd be happy to accept a PR that adds a comment, or expounds on the problem, but this is a very good example of a common and easy-to-make mistake.

Issue #540, opened 2015-10-07
Disallowing Unary Increment and Decrement (++ and --)


chrisngobanh:
In Douglas Crockford's Javascript: The Good Parts, he writes:

>  **++ and --**
> The ++ (increment) and -- (decrement) operators have been known to contribute to bad code by encouraging excessive trickiness. They are second only to faulty architecture in enabling to viruses and other security menaces. There is a plusplus option [in JSLint] that prohibits the use of these operators.

In addition, his replacement for ++ and -- is += 1 and -= 1, respectively.

Apparently Crockford's book is pretty popular, but I don't see anyone really talking about this part.

What does everyone think?

EDIT: This is a little off-topic, but Ruby does not support ++ and --.

ljharb:
I think `i++;` and `i += 1;` are mostly equivalent, but since our style guide also recommends avoiding loops, this doesn't come up often in practice. Where do you find yourself needing increment and decrement operators besides loops?

justjake:
I'm in. I don't like assignments like `foo = ++this.id`

ljharb:
I also don't like those, but because it's conflating assignment with expressions, not because of the `++`

chrisngobanh:
Similarly to loops, should we, instead of banning them (`++` `--`) outright, recommend not to use them?

jacobp100:
Python also doesn't allow this.

Also worth noting that `i++` is not the same as `++i`.

goatslacker:
Throughout our style guide we mention that mutation is bad, maybe this should be explicitly mentioned. `i += 1; i++;` or friends are all mutating.

sstern6:
It has been a while since someone has commented on this.  Has a PR already been made or would the community still want one.  I would be happy to make one for this.

Thanks

ljharb:
None has been made; I'd be happy to review one!

sstern6:
Thanks for the prompt response @ljharb!  Ill have that up as soon as I can, within the next 48 hours Im guessing.

This is my first contribution to the code base so I have a few questions.

1. I will only need to update `eslint-config-airbnb-base` correct? I was going to apply the no-plusplus rule from eslint http://eslint.org/docs/rules/no-plusplus.
[...]

PR #1012, opened 2016-08-14
[eslint config] [base] [breaking] Add no-plusplus in base errors.js and desc in README

sstern6:
@ljharb  Added `no-plusplus: 'error'` to eslint-config-airbnb-base/errors.js in order to disallow unary increment and decrement statements.  In addition, added description under Variables in the main README. 

This PR references https://github.com/airbnb/javascript/issues/540.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1278,6 +1278,35 @@ Other Style Guides
     }
     ```
 
+    - [13.5](#variables--unary-increment-decrement) Avoid using unary increments and decrements (++, --). eslint [`no-plusplus`](http://eslint.org/docs/rules/no-plusplus)
+
+      > Why? Per the eslint documentation, unary increment and decrement statements are subject to automatic semicolon insertion and can cause silent errors with incrementing or decrementing values within an application. It is also more expressive to mutate your values with statements like `num += 1` instead of `num ++`. Disallowing unary increment and decrement statements also prevents you from pre-incrementing/pre-decrementing values unintentionally which can also cause unexpected behavior in your programs.
+
+      ```javascript
+        // bad
+
+        let array = [1, 2, 3];
+        let num = 1;
+        let increment = num ++;
+        let decrement = -- num;
+
+        for(let i = 0; i < array.length; i++){
+          let value = array[i];
+          ++value;
+        }
+
+        // good
+
+        let array = [1, 2, 3];
+        let num = 1;
+        let increment = num += 1;
+        let decrement = num -= 1;
+
+        array.forEach((value) => {
+          value += 1;
+        });
+      ```
+
 **[⬆ back to top](#table-of-contents)**
 
 
--- a/packages/eslint-config-airbnb-base/rules/style.js
+++ b/packages/eslint-config-airbnb-base/rules/style.js
@@ -180,7 +180,8 @@ module.exports = {
     'no-new-object': 'error',
 
     // disallow use of unary operators, ++ and --
-    'no-plusplus': 'off',
+    // http://eslint.org/docs/rules/no-plusplus
+    'no-plusplus': 2,
 
     // disallow certain syntax forms
     // http://eslint.org/docs/rules/no-restricted-syntax


README.md
```diff
@@ -1418,6 +1418,35 @@ Other Style Guides
     }
     ```
 
+    <a name="variables--unary-increment-decrement"></a><a name="13.5"></a>
```
ljharb: there's no need to add a link to "13.5" - the numbers only exist for legacy reasons.

README.md
```diff
@@ -1418,6 +1418,35 @@ Other Style Guides
...
+    - [13.5](#variables--unary-increment-decrement) Avoid using unary increments and decrements (++, --). Eslint [`no-plusplus`](http://eslint.org/docs/rules/no-plusplus)
```
ljharb: `eslint` should be lowercased

README.md
```diff
@@ -1418,6 +1418,35 @@ Other Style Guides
...
+
+      > Why? Per the Eslint documentation, unary increment and decrement statements are subject to automatic semicolon insertion and can cause silent errors with incrementing or decrementing values within an application.  It is also more expressive to mutate your values with statements like `num += 1` instead of `num ++`.  Disallowing unary increment and decrement statements also prevents you from pre-incrementing/pre-decrementing values unintentionally which can also cause unexpected behavior in your programs.
```
ljharb: `eslint` should be lowercased

README.md
```diff
@@ -1418,6 +1418,35 @@ Other Style Guides
...
+        }
+
+        // good
+
+        let array = [1, 2, 3];
+        let num = 1;
+        let increment = num += 1;
+        let decrement = num -= 1;
+
+        for(let i = 0; i < array.length; i += 1){
```
ljharb: in this case, the guide discourages ever using loops, so the "good" example should be
```
array.forEach((item) => {
  doSomethingWithItem(item);
});
```

packages/eslint-config-airbnb-base/rules/errors.js
```diff
@@ -74,6 +74,10 @@ module.exports = {
     // http://eslint.org/docs/rules/no-prototype-builtins
     'no-prototype-builtins': 2,
 
+    // disallow use of unary operators, ++ and --
+    // http://eslint.org/docs/rules/no-plusplus
+    'no-plusplus': 'error',
```
ljharb: please use `2` here to remain consistent with the rest of the file

ljharb: Also, according to http://eslint.org/docs/rules/#stylistic-issues this rule is in "style", so please delete the relevant lines from this file, and move them to lines 183-185 in `rules/style.js`

README.md
```diff
@@ -1418,6 +1418,38 @@ Other Style Guides
...
+
+      ```javascript
+        // bad
+
+        let array = [1, 2, 3];
+        let num = 1;
+        let increment = num ++;
+        let decrement = -- num;
+
+        array.forEach((value) => {
```
ljharb: sorry if i was unclear before; i think having a loop in the "bad" example is actually a good thing :-)

sstern6: Ah ok, I apologize, I misunderstood. I interpreted it incorrectly.  Fixing now.

ljharb:
Thanks! I'd prefer this rebased down to a single commit, if you don't mind.

I'm also going to wait a while before merging this, since it's a breaking change.

sstern6:
Sure no problem, will have the rebase done tonight!  Thanks for the feedback through this PR

tleunen:
What do you think of allowing it in loops?

Sometimes, a basic for loop is used in order to do things and break/return something without doing through the entire array. And usually people are using `i++`, not really `i += 1`.

ljharb:
@tleunen if there was a way for the linter rule to allow it in loops, i'd consider that, but since this guide forbids any and all loops at any time for any reason, I'm not sure why someone following the guide would even run into the problem? (There's `Array#some`, `Array#every`, and `Array#find` for "stop partway through" logic)

sstern6:
@ljharb just for reference, there is an arg you can send in to no-plusplus that addresses the for loop issue.` "allowForLoopAfterthoughts": true `.  I understand that the guides forbid the use of  loops.  Just pointing it out for reference.

ljharb:
I'll think about whether we should add that, since there's not really as much of a need to avoid unary increment/decrement in for loop position.

README.md
```diff
@@ -1418,6 +1418,35 @@ Other Style Guides
     }
     ```
 
+    - [13.5](#variables--unary-increment-decrement) Avoid using unary increments and decrements (++, --). eslint [`no-plusplus`](http://eslint.org/docs/rules/no-plusplus)
+
+      > Why? Per the eslint documentation, unary increment and decrement statements are subject to automatic semicolon insertion and can cause silent errors with incrementing or decrementing values within an application.  It is also more expressive to mutate your values with statements like `num += 1` instead of `num ++`.  Disallowing unary increment and decrement statements also prevents you from pre-incrementing/pre-decrementing values unintentionally which can also cause unexpected behavior in your programs.
```
ljharb: sorry to be picky, but outside of typewriters there's no need for two spaces after a period :-)

packages/eslint-config-airbnb/package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "eslint-config-airbnb",
-  "version": "10.0.0",
+  "version": "10.0.1",
```
ljharb: i think you may need to rebase instead of merge?

README.md
sstern6: No worries, thanks for the feedback, will fix.

packages/eslint-config-airbnb/package.json
sstern6: Yep, rebased and repushed. Should be good now

ljharb:
@sstern6 hm, why does this commit have me listed as the author?

sstern6:
Ive never seen that before, this morning there were merge conflicts, I:

1) Pulled origin
2) Rebased against master
3) Pushed

It could have reassigned authorship when I rebased?  I will look into possible reasons and solutions.  Do you have any ideas?

ljharb:
Simplest might be to soft reset back to master, and make a new commit (on the same branch), and then force push that up?

I can think of any number of very unlikely ways this would happen, but I have no idea how this one would have, unless you started by cherry-picking one of my commits and amended it?

sstern6:
Ya sounds good!  No I did not do that.  Im also going to look at the reflogs and see if I can find where it went wrong for future reference.  I will reset soft back to master and make a new commit with the final changes then push!

Thanks

sstern6:
@ljharb just force pushed after resetting the commit, looked like in the rebase interactive, first squashing the commits amended your commit which is why I think you were granted authorship.  Everything should be up to date, let me know if you have any feedback or recommend any additional changes.

Thanks

sstern6:
@ljharb with this being a breaking change, how long do you wait until merging and what does that process look like, "merging an update that contains a breaking change in a large codebase".  1) for curiosity and 2) in order to time my rebase to make sure were green when you're ready to merge.

Thanks

ljharb:
@sstern6 totally fair question - generally i'd want to wait until we have a few major changes queued up for the base config, and then merge them all at once.

It might be a couple weeks - if you don't mind waiting, I'll ping you when the time is right :-)

sstern6:
Perfect thank you for the clarification.

ljharb:
@sstern6 do you mind rebasing? I'll merge and release this now.

sstern6:
REBASING NOW, STANDBY!

sstern6:
@ljharb should be solid, if anything needs to be adjusted please let me know.

Thanks

PR #1017, opened 2016-08-17
Readme: No variable assignment chaining

### LINKED ISSUE #221: Advise against multiple assignment in var declaration
I'd suggest to include a section in the guide explaining the scoping problems of this:

``` js
function foo() {
    var a = b = c = 1;
}
```

and to advise against it.

See http://stackoverflow.com/questions/1758576/multiple-left-hand-assignment-with-javascript

awesomeaniruddh:
Closes #221

### DIFF
--- a/README.md
+++ b/README.md
@@ -1417,6 +1417,38 @@ Other Style Guides
       return name;
     }
     ```
+  <a name=variables--no-chain-assignment""></a><a name="13.5"></a>
+  - [13.5](#variables--no-chain-assignment) Don't chain variable assignments. eslint: [`one-var`](http://eslint.org/docs/rules/one-var.html)
+
+    > Why? Chaining variable assignments creates implicit global variables.
+
+    ```javascript
+    // bad
+    (function example() {
+      // JavaScript interprets this as
+      // let a = ( b = ( c = 1 ) );
+      // The let keyword only applies to variable a; variables b and c become
+      // global variables.
+      let a = b = c = 1;
+    }());
+
+    console.log(a); // undefined
+    console.log(b); // 1
+    console.log(c); // 1
+
+    // good
+    (function example() {
+      let a = 1;
+      let b = a;
+      let c = a;
+    }());
+
+    console.log(a); // undefined
+    console.log(b); // undefined
+    console.log(c); // undefined
+
+    // the same applies for `const`
+    ```
 
 **[⬆ back to top](#table-of-contents)**
 


lencioni:
Thanks for the PR! Could you please rebase this down to one commit?

lencioni:
Are we sure that the one-var rule enforces this? I don't see any mention of that in the documentation for the rule and trying it out locally I am unable to get this rule to warn on things like `let a = b = c = 1;`. I do, however, get `no-undef` errors because `b` and `c` are not defined before being used. This can be resolved by doing:

```js
let b;
let c;
let a = b = c = 1;
```

which produces no warnings.

ljharb:
Hmm, interesting. I think `one-var` and `no-undef` are related, but don't directly forbid this usage. Perhaps we should ask eslint to create a rule to specifically catch this?

README.md
```diff
@@ -1417,6 +1417,38 @@ Other Style Guides
...
+
+    ```javascript
+    // bad
+    (function example() {
+      // JavaScript interprets this as
+      // let a = ( b = ( c = 1 ) );
+      // The let keyword only applies to variable a; variables b and c become
+      // global variables.
+      let a = b = c = 1;
+    })();
```
ljharb: let's put the invoking parens inside the containing parens, in both places

awesomeaniruddh:
Rebased and fixed the parens.

ljharb:
@awesomeaniruddh Thanks, this LGTM pending one last rebase :-)

README.md
```diff
@@ -1417,6 +1417,38 @@ Other Style Guides
       return name;
     }
     ```
+  <a name=variables--no-chain-assignment""></a><a name="13.5"></a>
+  - [13.5](#variables--no-chain-assignment) Don't chain variable assignments. eslint: [`one-var`](http://eslint.org/docs/rules/one-var.html)
```
lencioni: Since this eslint rule does not enforce this, does it make sense to mention it here? It seems misleading.

ljharb: I figured since it was related, it would apply, but if you think it's misleading we can remove it.

lencioni: I think it will make people think that this eslint rule enforces this part of the style guide, which it does not. More accurate would be `no-undef`, but that still does not fully enforce this. I think we should remove this reference and open an issue to either add a new rule or expand an existing rule to enforce this.

ljharb: Linking to #1021

PR #1021, opened 2016-08-18
Remove misleading reference to one-var eslint rule

lencioni:
This eslint rule does not enforce this style guide rule. I believe that
referencing it here will cause people to believe that this is enforced
via linting when it is not. As far as I know, there is no rule that
currently enforces this, but it would be nice to add one.


lencioni:
I opened a proposal for this rule here: https://github.com/eslint/eslint/issues/6932

PR #1023, opened 2016-08-19
Correct es2015 named module example

wdoug:
See http://exploringjs.com/es6/ch_modules.html for an in-depth explanation

### DIFF
--- a/README.md
+++ b/README.md
@@ -1097,7 +1097,7 @@ Other Style Guides
 
     // best
     import { es6 } from './AirbnbStyleGuide';
-    export default es6;
+    export es6;
     ```
 
   <a name="modules--no-wildcard"></a><a name="10.2"></a>


README.md
```diff
@@ -1097,7 +1097,7 @@ Other Style Guides
 
     // best
     import { es6 } from './AirbnbStyleGuide';
-    export default es6;
+    export es6;
```
ljharb: this is actually incorrect, as it will export a named export named "es6". We want `export default es6` so that there is only a default export.

ljharb:
In other words, we prefer a single default export, and zero named exports.

wdoug:
Ah now I see, the example is for one file. I was reading it as two different files, where one was exporting a default, and the other was trying to import a named export. 😬

Issue #1024, opened 2016-08-19
Why are underscore prefixes for internal methods of a React component considered bad?


ttmarek:
The third bullet under [react#methods](https://github.com/airbnb/javascript/tree/master/react#methods) says:

> Do not use underscore prefix for internal methods of a React component.

I was just wondering what the reasoning behind this was?

lencioni:
They give people a false sense of "private"ness that could lead to bugs.

ttmarek:
Okay. Even though JavaScript doesn't support private variables, I still think the underscores help communicate the programmer's intent. If possible, @lencioni could you give an example of this false sense of "private"ness that lead to a bug? Or maybe a scenario where the underscores would confuse another programmer and lead them astray. I'm not disagreeing with you, I just want to have a better understanding of how this rule came about.

ljharb:
Private means inaccessible. Your intent to privacy is irrelevant if the value is reachable, ie public.

For example, npm broke node once by removing an underscore-prefixed variable.

ttmarek:
> npm broke node once by removing an underscore-prefixed variable

If the variable _wasn't_ prefixed they would have used it either way, and node would have still broke. When a meant-to-be-private variable is prefixed with an underscore there is at least one line of defence (a weak one, granted) preventing its use. Whereas when a meant-to-be-private variable isn't prefixed with an underscore, there is no line of defence (not even a weak one).

ljharb:
@ttmarek yes, but had they not prefixed it with an underscore, nobody would have been fooled into thinking it was actually private, and there would have been test coverage, and changing it would have prompted a semver-major version bump.

An underscore is in no way, whatsoever, a line of defense - it's just an incorrect assumption some developers make about a language that simply does not have the concept of privacy outside of closures.

ttmarek:
@ljharb So, in other words: underscore prefixed methods might confuse programmers from other languages into thinking that JavaScript supports private methods. Right?

ljharb:
That, or programmers who think that convention is the same as privacy.

ttmarek:
Okay, cool. Thanks @ljharb, I see what you're getting at. Do you think its worth adding a short blurb about this under the third bullet in [react#methods](https://github.com/airbnb/javascript/tree/master/react#methods)?

lencioni:
I think so. I'd be happy to merge a PR that does that.

ljharb:
Sure, more helpful info is always good.

ericelliott:
[...]

Issue #490, opened 2015-08-27
Underscore prefixes for "private" properties


taion:
From the main section:
>Use a leading underscore `_` when naming private properties.

From the React section:
>Do not use underscore prefix for internal methods of a React component.

These two seem a bit inconsistent.

goatslacker:
A bit. The thing about React components is that all methods should be private so we just save ourselves the typing and don't insert the leading underscore.

taion:
What do you mean? If I have e.g. a ref to an element, I can call methods on it just fine. I've seen this used to implement e.g. `getValue` methods on things like complex, uncontrolled input components for forms.

If you consider this an anti-pattern, it could be useful to add a clarifying comment to say to not do this. There are definitely problems with this pattern - it's not container-safe, for one.

af7:
I am also interested in an explanation on that. Not all methods are "internal" if we consider that we may call a component's method from the parent.

Example:

**parent.jsx**
````
    _handleOpenChildDialog() {
        this.refs.childComponent.showDialog();
    }
````

**childComponent.jsx**
````
    showDialog() {
        this.refs.myDialog.show();
    }
````

So, what is our general take on this? Do we still write all methods without any underscore?

taion:
:+1: Facebook, for example, use this pattern extensively: https://github.com/facebook/relay/blob/v0.4.0/src/container/RelayRootContainer.js

hbrls:
@taion what about v0.14

taion:
What about it?

ljharb:
Personally, I think this pattern is horrifically bad. Things that are not private are not private - sticking an underscore in front of them doesn't make them private, it just lulls you into the false belief that you can make a breaking change without bumping a major version.

It's not private if it's accessible in the runtime. If it's private - ie, via closure or WeakMap - no convention or unit tests are needed. If it's not, it's public, and you *must* test it, and whether you document it or not, *it is public* and breaking it counts as a breaking change. At that point, using a leading underscore is just fluff, so why would you need a convention for it?

taion:
@ljharb 

That's totally fine - I personally don't mind writing my code either way (as long as I can point to a style guide and say that I'm following it).

The issue is that this style guide appears to be inconsistent between https://github.com/airbnb/javascript#22.4 and https://github.com/airbnb/javascript/tree/master/react#methods. I don't see how it's possible to follow both.

ljharb:
[...]

PR #1025, opened 2016-08-21
Explain why not to prefix methods with underscores

ttmarek:
Fixes issue #1024

### DIFF
--- a/react/README.md
+++ b/react/README.md
@@ -450,6 +450,7 @@
     ```
 
   - Do not use underscore prefix for internal methods of a React component.
+  > Why? Underscore prefixes are sometimes used as a convention in other languages to denote privacy. But, unlike those languages, there is no native support for privacy in JavaScript, everything is public. Regardless of your intentions, adding underscore prefixes to your properties does not actually make them private, and any property (underscore-prefixed or not) should be treated as being public. See issues #1024, and #490 for a more in-depth discussion.
 
     ```jsx
     // bad


react/README.md
```diff
@@ -450,6 +450,7 @@
     ```
 
   - Do not use underscore prefix for internal methods of a React component.
+  > Why? As Doug Crockford writes: "Do not use _ *underbar* as the first or last character of a name. It is sometimes intended to indicate privacy, but it does not actually provide privacy. If privacy is important, use closure. Avoid conventions that demonstrate a lack of competence." Source - http://javascript.crockford.com/code.html.
```
ljharb: I like Crockford's advice, but I don't think we need to propagate his tone here by quoting him. Could we write our own statement that conveys the same meaning?

ttmarek: Sure, I'll think of something that'll encapsulate all the main ideas expressed in #1024, #490, and Crockford's notes.

ttmarek: This might take me a couple days, its a busy work week this week. Also want to run this by a few developers I work with.

ttmarek:
image

react/README.md
```diff
@@ -450,6 +450,7 @@
     ```
 
   - Do not use underscore prefix for internal methods of a React component.
+  > Why? Underscore prefixes are sometimes used as a convention in other languages to denote privacy. But, unlike those languages, there is no native support for privacy in JavaScript, everything is public. Regardless of your intentions,  adding underscore prefixes to your properties does not actually make them private, and any property (underscore prefixed or not) should be treated as being public. See issues #1024, and #490 for a more in-depth discussion.
```
ljharb: there shouldn't be two spaces after "intentions,", and let's do "underscore-prefixed".

In what other languages is a single underscore used as a convention for privacy? I'm only aware of it in JS. PHP (and I think Python) use double underscores, but that's not the convention we're addressing here.

Otherwise, looks good!

skyrpex: PHP uses protected and private keywords. there's no need to use underscores

ljharb: That's true, I was referring to PHP's magic methods.

ttmarek: I'm only aware of the convention in three other languages (there may be more):

**Python**
> _single_leading_underscore : weak "internal use" indicator. E.g. from M import * does not import objects whose name starts with an underscore. - [PEP8 StyleGuide](https://www.python.org/dev/peps/pep-0008/#descriptive-naming-styles)

**C#**
I'm not an ex-C# programmer, but my colleagues are, and they tell me this is where their use of the underscore prefixes to denote privacy came from.
> We use _camelCase for internal and private fields and use readonly where possible. Prefix instance fields with _, static fields with s_ and thread static fields with t_. When used on static fields, readonly should come after static (i.e. static readonly not readonly static). - [.NET team styleguide](https://github.com/dotnet/corefx/blob/368fdfd86ee3a3bf1bca2a6c339ee590f3d6505d/Documentation/coding-guidelines/coding-style.md)

**Objective-C**
> Make sure the name of the instance variable concisely describes the attribute stored. Usually, you should not access instance variables directly, instead you should use accessor methods (you do access instance variables directly in init and dealloc methods). To help to signal this, prefix instance variable names with an underscore (_), for example - [Apple Guidelines for Cocoa?](http://stackoverflow.com/questions/3521254/prefixing-property-names-with-an-underscore-in-objective-c)

ljharb:
LGTM - would you mind rebasing this down to a single commit?

ttmarek:
@ljharb reabsed into one commit, and up-to-date with master.

PR #1032, opened 2016-08-25
static propTypes and defaultProps

zhongjie-chen:


ljharb:
This is using static class fields, which is not yet a standard part of JavaScript - it's still [stage 2](https://github.com/tc39/proposals).

Our style guide does not recommend using anything prior to stage 3.

Issue #1015, opened 2016-08-16
[Request] use constants for linter error levels


braco:
My team has forked this project a couple times because as linting rules are mostly 2/error instead of 1/warning,  it obscures genuine errors.

Something like this:
```
<Foo b={} a={} /> // props sorted wrong
```
Should not have the same severity as a critical bug! This adds a ton of mental overhead to every line when red dots are popping up everywhere: "will this code throw an error or is there some minor formatting issue?"

Needing to be so precious with every line slows down development, especially when you may just be testing some additions or wading into third party code. 

Doubly with autofix enabled, Atom currently fixes a ton of formatting inconsistencies on save, I can safely ignore some warnings now.

However, I understand your logic in #853 – when the team is huge and you need to enforce consistency, saying "you can't proceed without fixing this niggle" is important. I would argue that should go in a commit hook or similar, but...

why not use a custom constant to define the error levels? This would allow the linting logic to be more flexible:

`{ rule: airbnb.ERROR }`

ljharb:
Everything is a [genuine error](https://github.com/airbnb/javascript/issues/853#issuecomment-238640651), and yes, that most certainly should have the same [severity](https://github.com/airbnb/javascript/issues/853#issuecomment-238669832) as a critical bug. No formatting issues are minor, because they can obscure serious bugs and architectural issues in the code that are otherwise unlintable. In other words, you DO need to be so precious with every line - and if that slows down development, so be it: speed is not more important than correctness and rigor.

If you have errors autofixed on save, and you don't want to see them, can you not just hit ⌘-S or ctrl-S and remove them?

I'm leaving this open because it is an interesting idea to use environment variables to override each rule category's severity - it would not be more granular than "one env var per [file](https://github.com/airbnb/javascript/tree/master/packages/eslint-config-airbnb-base/rules)", but that would indeed make it easy to override some or all of the groups of rules' severities, without having to fork and/or replicate the config.

braco:
@ljharb: thanks for the discussion.

The logic you're using in that first link is a bit like the [brown M&M test](http://www.snopes.com/music/artists/vanhalen.asp) – if you suck at this part, you probably suck at other parts. That's a good rule of thumb, but that could also be enforced by rejecting PR/pushes with any warnings, among other things. In the meantime, rephrasing "warning, this might be obscuring an error" as "this is an error" is an extreme style in itself and broadly impacts everyone that uses this popular and thoughtful ruleset.

[...]

PR #1037, opened 2016-08-26
[base] use env flags to config severity level

hshoff:
Best viewed ignoring whitespace: https://github.com/airbnb/javascript/pull/1037/files?w=1
addressing: https://github.com/airbnb/javascript/issues/1015

:warning: WIP... just looking for implementation feedback. 

Currently it just reads from `process.argv` and parses out severity level flags. Flag options: `--warn`, `--off`, `--error`.

Usage would be from the command line like `--warn=some-rule,another-rule --off=semi`

Not positive if this is the best way to control severity levels. Would prefer to have this as a field in `.eslintrc` called `overrides` or something. Open to other ideas.

before:
```js
$ node
> require('./best-practices')
    rules: {
      'accessor-pairs':  'off',
      'array-callback-return': 'error',
      'block-scoped-var':  'error',
// ...
```

after:
```js
$ node
> process.argv.push('', "--warn=accessor-pairs", "--off=block-scoped-var")
4
> require('./best-practices')
{ rules:
   { 'accessor-pairs': 'warn',
     'array-callback-return': 'error',
     'block-scoped-var': 'off',
// ...
```

@ljharb @lencioni

### DIFF
--- a/packages/eslint-config-airbnb-base/rules/best-practices.js
+++ b/packages/eslint-config-airbnb-base/rules/best-practices.js
@@ -1,240 +1,244 @@
-module.exports = {
-  rules: {
-    // enforces getter/setter pairs in objects
-    'accessor-pairs': 'off',
+var sev = require('../utils/severity')();
 
-    // enforces return statements in callbacks of array's methods
-    // http://eslint.org/docs/rules/array-callback-return
-    'array-callback-return': 'error',
+module.exports = (function() {
+  return {
+    rules: {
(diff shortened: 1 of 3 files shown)


ljharb:
I'm not sure if this is the approach we want.

Primarily, I don't think we want per-rule granularity - override comments are best for those. I'm going to close this; but let's discuss this in person.

hshoff:
:+1:

PR #1041, opened 2016-08-28
Fix a typo in commas section

heejongahn:
Example code block following blockquote(line 2223) was broken as a line break was missing. So I just added single blank line.


ljharb:
Thanks!

PR #1044, opened 2016-08-29
Example in 6.6, violates 6.1.

gburtini:
The example showed a template string being used with no interpolation as "good". While it met the rule being described, it violated another rule for no reason.

### DIFF
--- a/README.md
+++ b/README.md
@@ -565,7 +565,7 @@ Other Style Guides
 
     // good
     const foo = '\'this\' is "quoted"';
-    const foo = `'this' is "quoted"`;
+    const foo = `my name is '${name}'`;		
     ```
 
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -544,7 +544,6 @@ Other Style Guides
 
     // good
     const foo = '\'this\' is "quoted"';
-    const foo = `'this' is "quoted"`;
```
ljharb: instead of removing it, can we add an interpolation?

gburtini: Done.

ljharb:
@gburtini would you mind checking the box "allow maintainers to edit"?

gburtini:
Done.

PR #1053, opened 2016-09-05
Update shell commands in readmes

niedzielski:
• Remove exports when variable is not used by subprocesses.

• Remove command keyword where no conflicting shell functions exist.

• Remove needless trailing semicolons for consistency with second
  statement.

### DIFF
--- a/packages/eslint-config-airbnb-base/README.md
+++ b/packages/eslint-config-airbnb-base/README.md
@@ -15,8 +15,8 @@ Our default export contains all of our ESLint rules, including ECMAScript 6+. It
 1. Ensure packages are installed with correct version numbers by running:
   ```sh
   (
-    export PKG=eslint-config-airbnb-base;
-    npm info "$PKG" peerDependencies --json | command sed 's/[\{\},]//g ; s/: /@/g' | xargs npm install --save-dev "$PKG"
+    PKG=eslint-config-airbnb-base
+    npm info "$PKG" peerDependencies --json | sed 's/[\{\},]//g ; s/: /@/g' | xargs npm install --save-dev "$PKG"
   )
   ```
 
@@ -35,8 +35,8 @@ Lints ES5 and below. Requires `eslint` and `eslint-plugin-import`.
 1. Ensure packages are installed with correct version numbers by running:
   ```sh
   (
-    export PKG=eslint-config-airbnb-base;
-    npm info "$PKG" peerDependencies --json | command sed 's/[\{\},]//g ; s/: /@/g' | xargs npm install --save-dev "$PKG"
+    PKG=eslint-config-airbnb-base
+    npm info "$PKG" peerDependencies --json | sed 's/[\{\},]//g ; s/: /@/g' | xargs npm install --save-dev "$PKG"
   )
   ```
 
--- a/packages/eslint-config-airbnb/README.md
+++ b/packages/eslint-config-airbnb/README.md
@@ -15,8 +15,8 @@ Our default export contains all of our ESLint rules, including ECMAScript 6+ and
 1. Ensure packages are installed with correct version numbers by running:
   ```sh
   (
-    export PKG=eslint-config-airbnb;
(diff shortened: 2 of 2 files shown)


packages/eslint-config-airbnb-base/README.md
```diff
@@ -15,8 +15,8 @@ Our default export contains all of our ESLint rules, including ECMAScript 6+. It
 1. Ensure packages are installed with correct version numbers by running:
   ```sh
   (
-    export PKG=eslint-config-airbnb-base;
-    npm info "$PKG" peerDependencies --json | command sed 's/[\{\},]//g ; s/: /@/g' | xargs npm install --save-dev "$PKG"
+    PKG=eslint-config-airbnb-base
+    npm info "$PKG" peerDependencies --json | sed 's/[\{\},]//g ; s/: /@/g' | xargs npm install --save-dev "$PKG"
```
ljharb: We can't know if conflicting commands exist - anyone could have a `sed` binary or an alias for it. `command` is always required.

ljharb:
The trailing semicolon is in case newlines are lost during copy-pasting. I prefer to keep it.

ljharb:
Regarding `export`, I'm reasonably confident it's needed, and that I added it in because things didn't work otherwise. On what shells did you verify this? Note you'll need to check dash, sh, bash, zsh, and ksh.

packages/eslint-config-airbnb-base/README.md
niedzielski: You can say the same of every single command. It's confusing to present what is likely only necessary to override shell personalization of a single user's private config.

ljharb: Yes, and that's why every command that's not `npm` here is prefixed with `command`. If there were others, they all would have `command` as well.

I'm not sure why it's confusing - most will just copy-paste the command and don't need to understand it. Those that do, will understand `command`.

niedzielski: Just to make sure we're on the same page, from the Dash manual:

> Execute the specified command but ignore shell functions when searching for it.  (This
> is useful when you have a shell function with the same name as a builtin command.)

Then I should change this to `command npm`, `command xargs`, ... in case a user defines their own functions for those as well? If a user has their own sed binary, they probably want to use it. In which case, they've either intentionally overridden the builtin and command will be unwanted or they haven't and PATH will be used so command is unneeded. Regardless, a user can't expect to override shell standards without consequence. They should choose a unique name instead. It seems foolish and obfuscating to attempt to code for someone who has sabotaged their own shell so severely.

ljharb: The goal is to achieve the desired results. Whenever possible, all code should be robust against all circumstances, including "the user intentionally changed something they probably shouldn't have" - this holds in JS as well as in shell scripting.

I don't see how it's "obfuscating" at all. I agree it should be on `xargs` however - and I'm ok with it not being on `npm` because that's not a shell built-in.

niedzielski:
> copy-pasting

When I copy and paste into my shell, newlines are preserved. Bash's `shopt -s cmdhist` will preserve multiline. When this option isn't enabled, I believe Bash correctly insert semicolons where they are needed when collapsing multiline commands. Just to confirm, do you still want me to revert this?

> export

export means "I need access to this variable in a child process". Unless you're running scripts that explicitly depend on this environment variable (a quick grep of this repo shows none), it's completely unnecessary and **extremely** confusing.

ljharb:
@niedzielski to confirm that it's unnecessary, please test the command in all five of the shells mentioned - if it works identically in all of them without `export`, i'm happy to remove it.

Bash is not the only shell I'm concerned about - and anything that makes the command more robust is something I'd want to keep - including the semicolon.

packages/eslint-config-airbnb-base/README.md
niedzielski: Hm, I completely disagree with your approach. Let's part ways.

PR #1056, opened 2016-09-06
File naming proposal 

mozartdiniz:


lencioni:
Thanks for the contribution, but we generally want our filenames to match what they export, which is either camelCase (e.g. a function) or PascalCase (e.g. a React component). We have this documented in rules 22.6-22.8.

PR #1060, opened 2016-09-08
Add JavaScript Air to list of Podcasts

lencioni:
It probably wouldn't be a bad idea for us to make a full pass on
updating the resources section, but I noticed the podcasts portion was
missing JS Air.

cc @kentcdodds


kentcdodds:
Wahoo!

PR #1062, opened 2016-09-08
remove unnecessary semicolon

kentcdodds:
From my best guess this:

```
Airbnb JavaScript Style Guide() {
  ...
}
```

is a function declaration, which does not need a semicolon :)


kentcdodds:
In case it's not obvious. This is a total joke and you can close this PR if you want :)

ljharb:
Based on https://github.com/airbnb/javascript#functions--declarations, we should change it to an NFE then :-p

kentcdodds:
Makes sense, but This is one case where I have the exact opposite opinion of this styleguide :P

PR #1070, opened 2016-09-12
Add notes on acronyms and initialisms to naming

kesne:
This was previously ambiguous. Let's reduce decision fatigue and just pick one.

### DIFF
--- a/README.md
+++ b/README.md
@@ -2761,6 +2761,36 @@ Other Style Guides
     export default AirbnbStyleGuide;
     ```
 
+  <a name="naming--Acronyms-and-Initialisms"></a>
+  - [22.9](#naming--Acronyms-and-Initialisms) Acronyms and initialisms should always be all capitalized, or all lowercased.
+
+    > Why? Names are for readability, not to appease a computer algorithm.
+
+    ```javascript
+    // bad
+    import SmsContainer from './containers/SmsContainer';
+
+    // bad
+    const HttpRequests = [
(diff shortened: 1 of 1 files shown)


goatslacker:
👍

README.md
```diff
@@ -2573,6 +2573,29 @@ Other Style Guides
...
+    // bad
+    import SmsContainer from './containers/SmsContainer';
+    
+    // bad 
+    const HttpRequests = [
+      // ...
+    ];
+    
+    // good
+    import SMSContainer from './containers/SMSContainer';
```
ljharb: i wonder if for both of these, we should include a "best" example as well that renames these things so as not to run into this question?

kesne: Maybe we can do that for double-acronyms? i.e. `XMLHTTPRequest` renamed to `HTTPRequest`?

ljharb: I'm thinking that for example, "TextMessageContainer" might be better than "SMSContainer", or "Request" better than "HTTPRequest" (but with better names).

kesne: Sounds good. I'll also add a note about allowing all-lowercase to support non-constructors.

mvolk: If we're going with uppercase acronyms, perhaps we should also provide guidance regarding running multiple acronyms together in a name?

ljharb: Definitely - my preferred guidance is "pick a name where that's not a problem" :-p

PR #1072, opened 2016-09-13
Change bad example of 8.2

learning:


README.md
```diff
@@ -823,8 +823,7 @@ Other Style Guides
     ```javascript
     // bad
     [1, 2, 3].map(number => {
+      return `A string containing the ${number}.`;
```
ljharb: why this change? The purpose is to make an arrow function that is not simply returning a single expression, and mistakenly omits the `return`.

learning: I think anyone reading this style guide will not forgot the `return` keyword, this is a **wrong** example, not **bad** example. 

```javascript
// wrong
[1, 2, 3].map(number => {
});
```

So, this **bad** example would be better (for me).

```javascript
// bad
[1, 2, 3].map(number => {
  return `A string containing the ${number}.`;
});
```

And the **good** examples make sense.

If only one expression, use one line arrow function with implicit `return`, compare with **bad** example.

```javascript
// good
[1, 2, 3].map(number => `A string containing the ${number}.`);
```

And more than 1 expressions, use normal arrow function with curly braces, compare with **bad** example.

```
// good
[1, 2, 3].map((number) => {
});
```

I'm not sure how people think of this original **bad** example, but for me, I have read it and couldn't get its point.

ljharb: wrong examples are bad examples, and bad examples are wrong examples. This is explicitly here because implicit return is a new feature in JS, and many people have misunderstood it.

learning: ok, maybe you're right, I close this now.

PR #1076, opened 2016-09-15
Set 'allowTemplateLiterals' to true on 'quotes' rule.

Jarwain:
Added allowTemplateLiterals option to quotes rule. This prevents a quotes error on a template string.

Current Behavior:
```
const foo = 'foo';
const foobar = foo + 'bar';
// Error Thrown: prefer-template
const foobar_template = `@{foo}bar`;
// Error Thrown: quotes (strings must be in single quotes)
```

New Behavior:
```
const foo = 'foo';
const foobar = foo + 'bar';
// Error Thrown: prefer-template
const foobar_template = `@{foo}bar`;
// No error thrown
```

### DIFF
--- a/packages/eslint-config-airbnb-base/rules/style.js
+++ b/packages/eslint-config-airbnb-base/rules/style.js
@@ -276,7 +276,7 @@ module.exports = {
     'quote-props': ['error', 'as-needed', { keywords: false, unnecessary: true, numbers: false }],
 
     // specify whether double or single quotes should be used
-    quotes: ['error', 'single', { avoidEscape: true }],
+    quotes: ['error', 'single', { avoidEscape: true , allowTemplateLiterals: true }],
 
     // do not require jsdoc
     // http://eslint.org/docs/rules/require-jsdoc


ljharb:
This is intentionally not turned on. Template literals should only be used when there are interpolations.

ljharb:
Specifically, "@{foo}" is not an interpolation - that'd be "${foo}".

Jarwain:
It was my mistake when writing up the pull request. They _should_ be interpolations. I messed up the syntax due to only discovering them maybe an hour ago. Pull Request is edited

ljharb:
Hmm - I can't reproduce that problem. `quotes` ignores template literals that have interpolations in them.

ljharb:
`echo 'const foo = `foo${bar}`;' | eslint --stdin --no-eslintrc --rule prefer-template:2 --rule quotes:2 --parser-options '{ecmaVersion:6}'` has no errors (with latest eslint)

Jarwain:
Turns out I only triggered the error initially because I had messed up the interpolation... Using `@{}` instead of `${}`. Thanks for humoring me

PR #1083, opened 2016-09-19
Add rule to avoid misusing DOM component prop names

lencioni:
People expect props like `style` and `className` to mean one specific
thing. Varying this API for a subset of your app makes the code less
readable and less maintainable, and may cause bugs.

### DIFF
--- a/react/README.md
+++ b/react/README.md
@@ -127,6 +127,19 @@
       return WithFoo;
     }
     ```
+
+  - **Props Naming**: Avoid using DOM component prop names for different purposes.
+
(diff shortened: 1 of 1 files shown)


react/README.md
```diff
@@ -127,6 +127,19 @@
       return WithFoo;
     }
     ```
+
+  - **Props Naming**: Avoid using DOM component prop names for different purposes.
```
ljharb: should this be linked to any eslint rules?

lencioni: Although react/forbid-component-props is similar, it does not exactly enforce this. I'm not aware of any that do. It would be nice if there was one, though. ;)

PR #1094, opened 2016-09-27
class-methods-use-this: exempt React getChildContext

vdh:
Added `getChildContext` to the exceptions list for `class-methods-use-this`


ljharb:
Thanks!

PR #1097, opened 2016-09-30
Add missing return statement for arrow function implicit return documentation

albertstill:


ljharb:
This is a duplicate of #1072, and is *intentionally* missing the return statement, which is what makes it a bad example.

PR #1101, opened 2016-10-02
Upgrade to import plugin v2

SimenB:
[Breaking changes](https://github.com/benmosher/eslint-plugin-import/blob/master/CHANGELOG.md#200---2016-09-30) seems to just involve changing defaults, which is already defined in this project, so no difference.
Still a breaking change for this module because of the changed peer I suppose, but rules wise, it's not.

### DIFF
--- a/packages/eslint-config-airbnb-base/rules/imports.js
+++ b/packages/eslint-config-airbnb-base/rules/imports.js
@@ -33,7 +33,7 @@ module.exports = {
 
     // ensure imports point to files/modules that can be resolved
     // https://github.com/benmosher/eslint-plugin-import/blob/master/docs/rules/no-unresolved.md
-    'import/no-unresolved': ['error', { commonjs: true }],
+    'import/no-unresolved': ['error', { commonjs: true, caseSensitive: true }],
 
     // ensure named imports coupled with named exports
     // https://github.com/benmosher/eslint-plugin-import/blob/master/docs/rules/named.md#when-not-to-use-it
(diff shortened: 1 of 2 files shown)


SimenB:
@ljharb updated 😄 

I only saw one added option (`caseSensitive `), am I missing something?
Or do you want to copy the new default options?

packages/eslint-config-airbnb-base/rules/imports.js
```diff
@@ -93,8 +93,8 @@ module.exports = {
     // Style guide:
 
     // disallow non-import statements appearing before import statements
-    // https://github.com/benmosher/eslint-plugin-import/blob/master/docs/rules/imports-first.md
-    'import/imports-first': ['error', 'absolute-first'],
+    // https://github.com/benmosher/eslint-plugin-import/blob/master/docs/rules/first.md
+    'import/first': ['error', 'absolute-first'],
```
ljharb: Let's also keep a line that disables `import/imports-first`, in case anyone is overriding it.

SimenB: Added (I think 😄 )

SimenB:
There was a conflict, rebased now.

I think at least the new webpack rule should be activated. Do you want that as a separate PR, or included in this?

ljharb:
That would also be a breaking change, and should be done in a separate PR/commit after this one is merged.

Arcanemagus:
Is there a timeline on when support for this version will be released? I've had to revert 3 PR's updating to this that got carelessly merged already.

SimenB:
I rebased and updated to 2.0.1, at least

ljharb:
@Arcanemagus no timeline yet. I'd recommend your tests run `npm ls` if you find bad updates to be a problem.

PR #1102, opened 2016-10-02
Add editorconfig

SimenB:
It silences IntelliJ a least 😄 
I think this should match the style in the repo.

image

### DIFF
--- a/.editorconfig
+++ b/.editorconfig
@@ -1,0 +1,11 @@
+root = true
+
+[*]
+indent_style = space
+indent_size = 2
+charset = utf-8
+trim_trailing_whitespace = true
+insert_final_newline = true
+end_of_line = lf
+# editorconfig-tools is unable to ignore longs strings or urls
+max_line_length = null
--- a/packages/eslint-config-airbnb-base/package.json
+++ b/packages/eslint-config-airbnb-base/package.json
@@ -4,6 +4,7 @@
   "description": "Airbnb's base JS ESLint config, following our styleguide",
   "main": "index.js",
   "scripts": {
+    "prelint": "editorconfig-tools check * rules/* test/*",
     "lint": "eslint .",
     "tests-only": "babel-tape-runner ./test/test-*.js",
     "prepublish": "(in-install || eslint-find-rules --unused) && (not-in-publish || npm test) && safe-publish-latest",
@@ -47,6 +48,7 @@
   "devDependencies": {
     "babel-preset-airbnb": "^2.1.1",
     "babel-tape-runner": "^2.0.1",
+    "editorconfig-tools": "^0.1.1",
     "eslint": "^3.12.1",
     "eslint-find-rules": "^1.14.3",
     "eslint-plugin-import": "^2.2.0",
--- a/packages/eslint-config-airbnb-base/.editorconfig
+++ b/packages/eslint-config-airbnb-base/.editorconfig
@@ -1,0 +1,1 @@
+../../.editorconfig
--- a/packages/eslint-config-airbnb/.editorconfig
+++ b/packages/eslint-config-airbnb/.editorconfig
@@ -1,0 +1,1 @@
+../../.editorconfig
--- a/packages/eslint-config-airbnb/package.json
+++ b/packages/eslint-config-airbnb/package.json
@@ -4,6 +4,7 @@
   "description": "Airbnb's ESLint config, following our styleguide",
   "main": "index.js",
   "scripts": {
+    "prelint": "editorconfig-tools check * rules/* test/*",
     "lint": "eslint .",
     "tests-only": "babel-tape-runner ./test/test-*.js",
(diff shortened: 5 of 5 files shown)


ljharb:
I don't think it's a good idea to include this file unless we can validate it against our existing code. Do you know of a tool that can do that?

SimenB:
Well, there is https://github.com/jedmao/eclint. I'm unable to make it work properly, though...

But the files use 2 space indent, trailing newline, and no trailing spaces.

All of these are validated by eslint already (maybe not utf-8, but that's usually default in modern editors, and I doubt this repo uses anything other than ascii either way), this file just automatically sets the IDE/text editor to those same settings.

Vadorequest:
This may be what you need, I haven't looked deeply into it but it seems good.

https://github.com/sun1x/eslint-to-editorconfig

SimenB:
@Vadorequest so the test would be to run the tool, and expect `git status` to be clean?

Vadorequest:
I think I misunderstood what you wanted, I was looking for a tool to generate a .editorconfig based on ESLint. 

But yeah, you could generate the file using this module and match it against yours.

SimenB:
@ljharb Found a tool to run a check on CI (same that webpack uses)

.editorconfig
```diff
@@ -0,0 +1,14 @@
...
+indent_size = 2
+charset = utf-8
+trim_trailing_whitespace = true
+insert_final_newline = true
+end_of_line = lf
+# editorconfig-tools is unable to ignore longs strings or urls
+max_line_length = null
+
+[*.md]
+trim_trailing_whitespace = false
```
ljharb: why would we not want to trim trailing whitespace on markdown files?

packages/eslint-config-airbnb-base/package.json
```diff
@@ -4,6 +4,7 @@
   "description": "Airbnb's base JS ESLint config, following our styleguide",
   "main": "index.js",
   "scripts": {
+    "prelint": "editorconfig-tools check *.js rules/*.js test/*.js",
```
ljharb: huh, i wasn't aware you could "pre" any old script. Seems this works all the way back to npm 1.

ljharb: This should probably cover jsx files too.

.editorconfig
```diff
@@ -0,0 +1,14 @@
+root = true
+
+[*]
```
ljharb: if this is `*`, then the editorconfig check should also use `*` - if it's just js/jsx, then i'd like both places to match.

SimenB: that's a good point

.editorconfig
SimenB: MD has significant whitespace, so if you want e.g. a linebreak without starting a new paragraph, you add two trailing spaces.

image

Want me to remove it?

SimenB:
Changed to run against every file

.editorconfig
ljharb: Interesting, I was not aware of that.

It's fine to leave as-is then.

SimenB:
Yes, it walks up the tree. I know it worked because it gave me failures on too long lines before I disabled that rule

SimenB:
I don't think the file needs to bi included in a publish, it's just useful for contributors, having their IDE configured automatically

ljharb:
@SimenB `npm explore foo && npm install && npm test` should work for every package - it definitely has to be included in the publish.

SimenB:
hm, ok. add a symbolic link, duplicate it, or copy it over during pubilsh (and gitignore it)?

ljharb:
I think the symlink is probably the simplest solution.

SimenB:
Ok, symlinks pushed

SimenB:
@ljharb rebased

ljharb:
hmm - when i try this locally, even if i add a hard tab, it doesn't error out.

SimenB:
Hmm, it doesn't seem to be able to check indent style... In the source code it's called `indent-char`, but changing to that doesn't help.

It does correctly yell at me for trailing spaces, trailing newline, CRLF and size of indentation, though. And it also yelled at me for line length, but as it\s unable to ignore long strings or comments, it gave a lot of false positives.

A bummer it doesn't seems to check type of indent, but extra indent gets yelled at

SimenB:
Also, setting `indent-style=tab` makes it fail for spaces, so it just doesn't seem able to detect a single tab in all the spaces

ljharb:
gotcha - seems i just got lucky and picked the one thing it can't detect. oh well, this seems fine :-)

SimenB:
Hmm, If I set _all_ indentation to tabs, it yells at me for not using spaces... Seems like a bug

ljharb:
@SimenB please file that upstream if you don't mind ;-)

SimenB:
Done, slang800/editorconfig-tools#36

Thanks for merging!

PR #1105, opened 2016-10-04
Make section numbering consistent

randallreedjr:
* Renumber items where numbers were skipped

### DIFF
--- a/README.md
+++ b/README.md
@@ -155,7 +155,7 @@ Other Style Guides
     ```
 
   <a name="es6-computed-properties"></a><a name="3.4"></a>
-  - [3.4](#es6-computed-properties) Use computed property names when creating objects with dynamic property names.
+  - [3.2](#es6-computed-properties) Use computed property names when creating objects with dynamic property names.
 
     > Why? They allow you to define all the properties of an object in one place.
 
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -154,8 +154,8 @@ Other Style Guides
     const item = {};
     ```
 
-  <a name="es6-computed-properties"></a><a name="3.4"></a>
-  - [3.4](#es6-computed-properties) Use computed property names when creating objects with dynamic property names.
+  <a name="es6-computed-properties"></a><a name="3.2"></a>
```
ljharb: the anchor links are intentionally left as they are, to ensure old links continue to work forever. same throughout.

README.md
```diff
@@ -1260,7 +1260,7 @@ Other Style Guides
 
     > Why? They don't transpile well to ES5.
 
-  <a name="generators--spacing"></a>
+  <a name="generators--spacing"></a><a name="11.3"></a>
```
ljharb: please do not add any new links that have numbers. that's the old-style legacy format, and we won't be adding or changing numbered links forevermore.

README.md
randallreedjr: Thanks for pointing this out @ljharb. Does this mean that you would prefer to keep the numbering as is (out of sequence), or update the numbering but leave the links alone?

README.md
randallreedjr: Ah, I suspected that might be the case, but couldn't confirm.

README.md
randallreedjr: @ljharb In that case, I think this PR is ready to be reviewed again.

PR #1114, opened 2016-10-09
Update section 8.1

CalebAnderson2014:
The "good" example on line 832 breaks the one arg parens rule for the style guide.

### DIFF
--- a/README.md
+++ b/README.md
@@ -829,7 +829,7 @@ Other Style Guides
     });
 
     // good
-    [1, 2, 3].map((x) => {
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -829,7 +829,7 @@ Other Style Guides
     });
 
     // good
-    [1, 2, 3].map((x) => {
```
ljharb: Actually this is correct - the guide requires that one-arg arrow parens be omitted only for implicit return - and requires that they be included for explicit return.

PR #1129, opened 2016-10-14
eslint-import version update

KeKs0r:


ljharb:
This is intentional until eslint-config-airbnb-base is released.

PR #1131, opened 2016-10-15
Update `import/no-extraneous-dependencies` to alow test files

SimenB:
It now allows devDependencies in all files called `test.js`, `spec.js`,
`*.test.js` or `*.spec.js`


ljharb:
Thanks :-) I already have this change locally, but it's blocked on https://github.com/benmosher/eslint-plugin-import/issues/602

ljharb:
`eslint-config-airbnb` `v13.0.0` is now released.

stevemao:
@ljharb Would you consider `*.spec.js`?

ljharb:
@stevemao a PR that adds that seems just fine.

PR #1132, opened 2016-10-16
Update eslint to 3.8.0

SimenB:
New rule called `func-name-matching` added

Based on https://github.com/airbnb/javascript#functions, where you have

```js
// good
const foo = function bar() {
};
```

`off` seems correct here.

I also added a missing link

### DIFF
--- a/packages/eslint-config-airbnb-base/rules/style.js
+++ b/packages/eslint-config-airbnb-base/rules/style.js
@@ -32,7 +32,13 @@ module.exports = {
     // http://eslint.org/docs/rules/func-call-spacing
     'func-call-spacing': ['error', 'never'],
 
+    // requires function names to match the name of the variable or property to which they are
+    // assigned
+    // http://eslint.org/docs/rules/func-name-matching
+    'func-name-matching': ['off', { includeCommonJSModuleExports: false }],
+
     // require function expressions to have a name
+    // http://eslint.org/docs/rules/func-names
     'func-names': 'warn',
 
     // enforces use of function declarations or expressions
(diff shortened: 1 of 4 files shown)


SimenB:
So replace `'always-multiline'` with 4x `'always-multiline'`, and an `ignore` with TODO for `always` (not `'always-multiline'`?)?

SimenB:
Added the two others, and rebased.

For `space-before-function-paren` i did not follow the default of eslint as that would be against this guide. And as the change is breaking anyways, no issue there

ljharb:
whoops, you're right, `always-multiline` would be the desired setting for `functions`

ljharb:
The eslint settings we want for `space-before-function-paren`:
```js
    {
        "anonymous": "never",
        "named": "never",
        "asyncArrow": "ignore" // TODO: semver-major; set to "always"
    }
```

packages/eslint-config-airbnb-base/rules/style.js
```diff
@@ -303,7 +309,11 @@ module.exports = {
 
     // require or disallow space before function opening parenthesis
     // http://eslint.org/docs/rules/space-before-function-paren
-    'space-before-function-paren': ['error', { anonymous: 'always', named: 'never' }],
+    'space-before-function-paren': ['error', {
+      anonymous: 'always',
+      named: 'never',
+      asyncArrow: 'always'
```
ljharb: this needs to stay `ignore` since eslint 3.7 parses `async () => {}` just fine, so `always` would be a breaking change.

SimenB: Even though the changed peer already is a breaking change? Along with import v2 already in master?

ljharb: Oh, that is a good point. ok, then in that case, let's also update the other TODO :-) thanks!

PR #525, opened 2015-09-25
Unnecessary semicolon

luqmaan:
Remove unnecessary semicolon.

:trollface:

PR #1136, opened 2016-10-18
no-extra-semi on function declation style decoration.

graingert:


ljharb:
Duplicate of #525. The name isn't a valid identifier either, but whimsy is important.

graingert:
https://github.com/airbnb/javascript/pull/1137

ljharb:
I don't think we need extra documentation that "not code" shouldn't need PRs to "correct" its code.

PR #1137, opened 2016-10-18
Prevent PRs attempting a semicolectomy on the docs

graingert:


graingert:
@ljharb an intentional PR trap?

ljharb:
It's not code. If you're compelled to make a PR making a code correction on "not code", I don't think there's any way to avoid you making an unnecessary PR.

Issue #1008, opened 2016-08-10
Inconsistent Opinion on Function Expressions


peterkrieg:
Just wanted to clarify something that was confusing to me and perhaps others reading.

[7.1](https://github.com/airbnb/javascript#functions--declarations) states that the following is "bad":

```
// bad
const foo = function () {};
```

However, [7.11](https://github.com/airbnb/javascript#functions--signature-spacing) states the following is "good":

```
// good
const x = function () {};
```

I understand [7.11](https://github.com/airbnb/javascript#functions--signature-spacing) is talking about spacing alone, but the fact that the example used was what _not_ to do in the previous section [7.1](https://github.com/airbnb/javascript#functions--declarations) seemed strange.

ljharb:
7.1 probably needs to be revised a bit.

PR #1140, opened 2016-10-25
[guide] Fix inconsistency in function expression rules.

AndrewSouthpaw:
The link to a discussion around anonymous expressions and whether they're easy to find in ES6 ([diff](https://github.com/airbnb/javascript/pull/1036/commits/1541503fd47626e2d4b03ff01bca76940081edeb)) was lost in the PR #1008 , and introduced a slight inconsistency in the "good" example where the function expression is different from the named function, which I'm guessing was not intentional.

### DIFF
--- a/README.md
+++ b/README.md
@@ -576,7 +576,7 @@ Other Style Guides
   <a name="functions--declarations"></a><a name="7.1"></a>
   - [7.1](#functions--declarations) Use named function expressions instead of function declarations. eslint: [`func-style`](http://eslint.org/docs/rules/func-style) jscs: [`requireFunctionDeclarations`](http://jscs.info/rule/requireFunctionDeclarations)
 
-    > Why? Function declarations are hoisted, which means that it’s easy - too easy - to reference the function before it is defined in the file. This harms readability and maintainability. If you find that a function’s definition is large or complex enough that it is interfering with understanding the rest of the file, then perhaps it’s time to extract it to its own module! Don’t forget to name the expression - anonymous functions can make it harder to locate the problem in an Error's call stack.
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -588,7 +588,7 @@ Other Style Guides
     }
 
     // good
-    const foo = function bar() {
+    const foo = function foo () {
```
ljharb: this is actually intentional - having the lexically bound name reference be different than the variable containing the function means that things are more greppable, debuggable, and easier to maintain.

AndrewSouthpaw: Thanks. Is there a rule in the style guide that explains the part about using a different lexically bound name reference? I haven't come across that before...

ljharb: No prose - this example constitutes the preference. There's actually an eslint rule that enforces that they be different, but I'm not planning on enabling it in the config by default.

AndrewSouthpaw: Got it. So I'm just really curious, how does having the names be different make things easier to maintain?

ljharb: Primarily, it makes it easier to locate when you `grep` for the name of the function as read in a stack trace - it also lets you use a short, simple name for the local variable, but a long, verbose name for the function, for better debugging.

AndrewSouthpaw: Oh, so you would do something like:

```js
const doIt = function doSomethingWithSomethingElse {
  ...
};
...
doIt();
```
?

ljharb: yep! just like that.

If you're using recursion, it also helps disambiguate where the recursive call occurs.

AndrewSouthpaw: Thanks for taking the time to explain it. :) TIL!

PR #1143, opened 2016-10-26
Fix broken rule reference link `import/first`

1pete:
`import/imports-first` is renamed to `import/first`.

Also old link has been deleted, got 404 when navigate.

---

This change is [Reviewable](https://reviewable.io/reviews/airbnb/javascript/1143)

### DIFF
--- a/README.md
+++ b/README.md
@@ -1239,7 +1239,7 @@ Other Style Guides
 
   <a name="modules--imports-first"></a>
   - [10.7](#modules--imports-first) Put all `import`s above non-import statements.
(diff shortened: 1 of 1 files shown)


ljharb:
This is part of eslint-plugin-import v2.0, which the main config has not updated to yet. #1101 upgraded it for the base config, but not for the react config.

ljharb:
(as a side note, Github has PR reviews, so I'm not sure why we'd want to use reviewable.io on a per-PR basis - that seems like a project-level decision)

PR #1148, opened 2016-10-30
Allows unary operators ++ and -- in the afterthought (final expression) of a for loop

frenzzy:
RE: [13.6 Avoid using unary increments and decrements](https://github.com/airbnb/javascript#variables--unary-increment-decrement)
Let's allow to use canonical for-loop syntax with `i++`.
```js
for (i = 0; i < l; i--) {
    return;
}

function* recursion(/* ... */) {
  for (let i = 0; i < length; i++) {
    yield* recursion(/* ... */);
  }
}
```
Using addition assignment here looks strange to say the least.
ESLint [no-plusplus](http://eslint.org/docs/rules/no-plusplus) rule has a [special case](http://eslint.org/docs/rules/no-plusplus#allowforloopafterthoughts) for allowing unary operators in the for-loops.

### DIFF
--- a/packages/eslint-config-airbnb-base/rules/style.js
+++ b/packages/eslint-config-airbnb-base/rules/style.js
@@ -212,7 +212,7 @@ module.exports = {
 
     // disallow use of unary operators, ++ and --
     // http://eslint.org/docs/rules/no-plusplus
-    'no-plusplus': 'error',
+    'no-plusplus': ['error', { allowForLoopAfterthoughts: true }],
 
     // disallow certain syntax forms
     // http://eslint.org/docs/rules/no-restricted-syntax


ljharb:
Thanks for the contribution!

Both generators and loops are forbidden by this style guide, so this shouldn't even come up (and I don't agree that `i += 1` looks particularly strange).

I don't think it would be a good idea to loosen this rule - if you want to use loops, you have to override a number of our rules, and I think that's something that should be difficult.

everdimension:
@ljharb Well, have you thought about how you would alter these rules once generators do become stable?

> and I don't agree that i += 1 looks particularly strange

It looks like someone tried to artificially satisfy a weird linter rule :) There is no reason why `i += 1` is better as the loop condition than the conventional `i++`.

ljharb:
Even when generators can be transpiled without regenerator-runtime, we'd still want to discourage loops.

frenzzy:
So, until this style guide prohibits the unary operators and `for` loops, it does not matter whether the `allowForLoopAfterthoughts` flag is turned on or off. But for style guides based on it and which allow `for` loops will need to redefine less rules. I think it's a good point, do not you agree?

PR #1153, opened 2016-11-01
Add comma-dangle for functions Fixes #951

### LINKED ISSUE #951: Trailing commas in function arguments
What about trailing commas here?

``` javascript
const hero = (
  firstName,
  lastName,
  birthYear,
  superPower, // <-- this one
) => {
  // ....
};
```

It needs [http://babeljs.io/docs/plugins/syntax-trailing-function-commas/](http://babeljs.io/docs/plugins/syntax-trailing-function-commas/) and it's based on [this proposal](https://github.com/jeffmo/es-trailing-function-commas)

Does AirBnB encourages the use of trailing commas here?

graingert:


packages/eslint-config-airbnb-base/rules/errors.js
```diff
@@ -1,7 +1,12 @@
 module.exports = {
   rules: {
     // require trailing commas in multiline object literals
-    'comma-dangle': ['error', 'always-multiline'],
+    'comma-dangle': ['error', {
+      objects: 'always-multiline',
```
ljharb: this should also include `arrays` set to "always-multiline"

ljharb:
Also, we'll need to update https://github.com/airbnb/javascript#commas--dangling to have examples of both multiline function signatures and invocations

README.md
```diff
@@ -2397,6 +2397,45 @@ Other Style Guides
...
+      firstName,
+      lastName,
+      inventorOf,
+    ) {
+      // does nothing
+    }
+    
+    // good
+    function createHero(firstName,
+                        lastName,
```
ljharb: this is actually a bad example.

graingert: It's not, because the rule is specifically designed to allow it.

ljharb: It's a bad example for our styleguide - that style is absolutely not permitted, even if `comma-dangle` works with it. Other linter rules should be forbidding it.

ljharb: Specifically, "good" doesn't mean "passes the linter rule", "good" means "conforms to our styleguide"

graingert:
Where is this style banned?

ljharb:
It seems we don't have a specific section on it in https://github.com/airbnb/javascript#functions - I'll add one now.

In general, in any case, any time the entire `( )`, `[ ]`, or `{ }` doesn't fit on a single line, the `(`/`[`/`{` must end a line, the `)`/`]`/`}` must begin a line, and every item must be on a line by itself.

graingert:
You should totally request a rule for that from eslint

ljharb:
Section added in c51251d17b8e38f1e8975a5e4eb13d7986229410

README.md
graingert: I've marked this branch as editable by maintainers. Otherwise you can wait a few hours for me to get to a keyboard

PR #1155, opened 2016-11-02
2 corrections

vp2177:
* I think you meant "Always use `let` to declare variables", as `const` declares a constant, not a variable. Although one could argue that a constant is a subtype of a variable, hence my more generic wording
* The only valid ES syntax is parenthesis around multiple arrow-function arguments, reworded accordingly

### DIFF
--- a/README.md
+++ b/README.md
@@ -918,7 +918,7 @@ Other Style Guides
     ```
 
   <a name="arrows--one-arg-parens"></a><a name="8.4"></a>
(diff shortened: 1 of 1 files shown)


ljharb:
`const` is still called a variable, and sometimes omitting the arg parens will be treated as a value, the comma operator, and an arrow function - the current wording is correct.

Thanks for your contribution!

PR #1156, opened 2016-11-02
Use num++ instead of num ++

vanillajonathan:
I've never seen anyone do `num ++`, people don't do that. So I changed it to `num++` so it is more realistic and reflects actual common practices.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1547,15 +1547,15 @@ Other Style Guides
   <a name="variables--unary-increment-decrement"></a><a name="13.6"></a>
   - [13.6](#variables--unary-increment-decrement) Avoid using unary increments and decrements (++, --). eslint [`no-plusplus`](http://eslint.org/docs/rules/no-plusplus)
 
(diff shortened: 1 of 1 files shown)


ljharb:
People absolutely use both spacings, which is why there's an eslint rule to prevent having the separating space.

Thanks for the contribution! Happy to reopen if you want to instead adapt it to include both patterns, but as-is the current spacing is what we'll stick with.

vanillajonathan:
I added to patch-1 some appeasement.
vanillajonathan/javascript@09fc01195cbfaa4bd4ea3d54e31886586bdf0de0

Perhaps this is satisfactory.

ljharb:
@vanillajonathan thanks, this is great! Would you mind rebasing down to a single commit?

vanillajonathan:
I use the GitHub website to commit and it does not support rebasing.

I checked the "Allow edits from maintainers" checkbox. Maybe that allows something.

ljharb:
@vanillajonathan no problem, i'll take care of it for you

PR #1157, opened 2016-11-02
Corrected case

BusbyActual:
Updated case to read the same in multiple spots. Added single quote example

PR #1158, opened 2016-11-03
Updated case of bullets and added a quotes example

BusbyActual:


ljharb:
Already taken care of in https://github.com/airbnb/javascript/pull/1157#issuecomment-258056448 :-)

PR #1167, opened 2016-11-09
Fixed a Arguement Error

CyberiumShadow:
func_name_matching had off and always together.

### DIFF
--- a/packages/eslint-config-airbnb-base/rules/style.js
+++ b/packages/eslint-config-airbnb-base/rules/style.js
@@ -35,7 +35,7 @@ module.exports = {
     // requires function names to match the name of the variable or property to which they are
     // assigned
     // http://eslint.org/docs/rules/func-name-matching
-    'func-name-matching': ['off', 'always', {
+    'func-name-matching': ['off', {
       includeCommonJSModuleExports: false
     }],
 


packages/eslint-config-airbnb-base/rules/style.js
```diff
@@ -35,7 +35,7 @@ module.exports = {
     // requires function names to match the name of the variable or property to which they are
     // assigned
     // http://eslint.org/docs/rules/func-name-matching
-    'func-name-matching': ['off', 'always', {
```
ljharb: actually, this is correct. the string option is optional, but our configs prefer to be explicit.

ljharb:
If you're running into an eslint error, it's because you have to upgrade your version of eslint.

CyberiumShadow:
@ljharb What would need to be updated then?

ljharb:
I'd try restarting atom - it's got a bunch of bugs around changing node_modules while remaining open.

CyberiumShadow:
Turns out that installing Node 6.9.1 wiped my global node modules.
All good now, I also had Global ESLint enabled for my linter.

Sorry for the trouble!

ljharb:
Glad to hear it - you definitely want everything installed locally.

Issue #1168, opened 2016-11-09
ignore 'import/no-extraneous-dependencies' devDependencies checking for webpack files?


chentsulin:
Maybe `gulpfile.js`, `Gruntfile` or other task runner files also can be listed to ignore.

```js
'import/no-extraneous-dependencies': ['error', {
  devDependencies: [
    'spec/**', 
    'test/**', 
    'tests/**', 
    '**/__tests__/**',
    'webpack.config.js',
    'webpack.config.*.js',
    'gulpfile.js',
    'Gruntfile',
  ],
  optionalDependencies: false,
}],
```

Any feedback?

ljharb:
That sounds like a great idea. I'd be happy to accept a PR.

Willibaur:
Hi all,
I've been through many of the Issues on this repo, and could not find what I'm looking for so I'm going to leave my question here, I have this rule in my `.eslintrc.js`

`'import/no-extraneous-dependencies': ["error", { devDependencies: true, }],`

But is still giving my a bunch of errors, all of them like this one

` 1:1  error  'ember' should be listed in the project's dependencies. Run 'npm i -S ember' to add it  import/no-extraneous-dependencies`

Any help would be appreciated
Thanks in advance

ljharb:
@Willibaur a new issue on eslint-plugin-import is the appropriate place to file that.

Willibaur:
Thanks @ljharb I did that, so answer for this can be found here 
https://github.com/benmosher/eslint-plugin-import/issues/422#issuecomment-292606152
just in case some else has the same problem.

PR #1170, opened 2016-11-09
Added some ignore pattern for import/no-extraneous-dependencies

chentsulin:
Close https://github.com/airbnb/javascript/issues/1168

### DIFF
--- a/packages/eslint-config-airbnb-base/rules/imports.js
+++ b/packages/eslint-config-airbnb-base/rules/imports.js
@@ -68,7 +68,17 @@ module.exports = {
     // https://github.com/benmosher/eslint-plugin-import/blob/master/docs/rules/no-extraneous-dependencies.md
     // paths are treated both as absolute paths, and relative to process.cwd()
     'import/no-extraneous-dependencies': ['error', {
-      devDependencies: ['spec/**', 'test/**', 'tests/**', '**/__tests__/**'],
+      devDependencies: [
+        'spec/**',
+        'test/**',
+        'tests/**',
+        '**/__tests__/**',
+        '**/webpack.config.js',
+        '**/webpack.config.*.js',
+        '**/rollup.config.js',
+        '**/gulpfile.js',
+        '**/Gruntfile',
+      ],
       optionalDependencies: false,
     }],
 


packages/eslint-config-airbnb-base/rules/imports.js
```diff
@@ -68,7 +68,17 @@ module.exports = {
     // https://github.com/benmosher/eslint-plugin-import/blob/master/docs/rules/no-extraneous-dependencies.md
     // paths are treated both as absolute paths, and relative to process.cwd()
     'import/no-extraneous-dependencies': ['error', {
-      devDependencies: ['spec/**', 'test/**', 'tests/**', '**/__tests__/**'],
+      devDependencies: [
+        'spec/**', 
+        'test/**', 
+        'tests/**', 
+        '**/__tests__/**',
+        '**/webpack.config.js',
```
ljharb: why would anyone have their webpack config not at the root? the leading "**/" should be removed - same for all the other config files.

chentsulin: I ever see some usage case like:

`examples/webpack.config.js` 

or 

```
webpack/webpack.config.js
webpack/webpack.config.dev.js
webpack/webpack.config.prod.js
webpack/webpack.config.test.js
```

But I can take those changes if it is too much.

ljharb: hmm - i guess that makes sense. let's keep them for now.

packages/eslint-config-airbnb-base/rules/imports.js
```diff
@@ -68,7 +68,17 @@ module.exports = {
     // https://github.com/benmosher/eslint-plugin-import/blob/master/docs/rules/no-extraneous-dependencies.md
     // paths are treated both as absolute paths, and relative to process.cwd()
     'import/no-extraneous-dependencies': ['error', {
-      devDependencies: ['spec/**', 'test/**', 'tests/**', '**/__tests__/**'],
+      devDependencies: [
+        'spec/**',
+        'test/**',
+        'tests/**',
+        '**/__tests__/**',
```
stevemao: Would be great to include `*.spec.js` as well :)

PR #1173, opened 2016-11-10
Add information about NaN comparisons

alleroux:
You mention NaN evaluates to false but don't explicitly mention that any comparison to NaN evalutes to false, including NaN === NaN.  Added a note about this at the bottom of that section.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1695,6 +1695,7 @@ Other Style Guides
     + **Booleans** evaluate to **the value of the boolean**
     + **Numbers** evaluate to **false** if **+0, -0, or NaN**, otherwise **true**
     + **Strings** evaluate to **false** if an empty string `''`, otherwise **true**
+    + **NaN** comparisons always evaluate to false, even **NaN === NaN**
 
     ```javascript
     if ([0] && []) {


README.md
```diff
@@ -1695,6 +1695,7 @@ Other Style Guides
     + **Booleans** evaluate to **the value of the boolean**
     + **Numbers** evaluate to **false** if **+0, -0, or NaN**, otherwise **true**
     + **Strings** evaluate to **false** if an empty string `''`, otherwise **true**
+    + **NaN** comparisons always evaluate to false, even **NaN === NaN**
```
ljharb: This isn't actually true. This section is about *if statement* comparisons, so `NaN` is covered here under "**Numbers** evaluate to **false** if **+0, -0, or NaN**, otherwise **true**"

I'm fine including more information about `NaN !== NaN`, and `NaN` in general, but "if comparisons" aren't the section for it.

ljharb: (Additionally, `NaN !== NaN` is a NaN comparison that evaluates to true)

ljharb (changes requested):

README.md
alleroux: That's true. Perhaps there is a more concise way to say it.

PR #1175, opened 2016-11-12
Extend `import/no-extraneous-dependencies` ignore list

### LINKED ISSUE #1174: Extend `import/no-extraneous-dependencies` ignore list
What do you think about adding follow patterns to ignore list?

```js
'test.js',
'test-*.js',
'**/*.test.js',
```

Like in [AVA](https://github.com/avajs/ava-files/blob/master/index.js#L22-L26)

andrepolischuk:
Fix #1174


SimenB:
@ljharb Could you do a patch release with this? Or are breaking changes also merged, so we'll have to wait for a major?

ljharb:
The latter.

PR #1177, opened 2016-11-15
Update readme with yarn usage

gblejman:


ljharb:
Thanks for the contribution! Not only do I not think this is accurate (I think yarn has a command to install the peer deps for you), but yarn isn't quite production ready yet, and Airbnb is not yet recommending it's use.

hkdobrev:
@ljharb Do you think AirBnB is open to recommending Yarn usage now as it has a greater adoption now and is a bit more mature? Thanks!

ljharb:
No, it's still not fully production-ready.

oshalygin:
@hkdobrev there have been a number of reported issues(some of which I've experienced myself) when using private NPM repositories and yarn.

PR #1181, opened 2016-11-21
Fixed invalid JS example.

tamasfabi:
image


ljharb:
This is perfectly valid ES2017. You're expected when using our guide and config to be using https://npmjs.com/babel-preset-airbnb.

PR #1187, opened 2016-11-28
Remove wrong example for 11.1

apsavin:
Removed example had nothing in common with the "Why" section and was too much like "bad" one.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1313,11 +1313,6 @@ Other Style Guides
     sum === 15;
 
     // good
-    let sum = 0;
-    numbers.forEach(num => sum += num);
-    sum === 15;
-
-    // best (use the functional force)
     const sum = numbers.reduce((total, num) => total + num, 0);
     sum === 15;
     ```


ljharb:
The (not wrong) example shows the evolution from loops, to forEach, to functional iteration methods. I'd prefer to keep it.

apsavin:
But there is no sense in such evolution - from for loop with side effect to forEach with side effect.

km256:
>The (not wrong) example shows the evolution from loops, to forEach, to functional iteration methods. 

But why it's better?

apsavin:
correctly used `for` loop -> bad usage of functional iteration method -> good usage of functional iteration method

Sorry, but I just can't get it.

ljharb:
@apsavin i understand you don't see the sense in it - however, that progression is exactly the way many people learn functional patterns. `forEach` is similar enough to a `for` loop that it serves as a good stepping stone to learning about `map`, `filter`, etc.

jifeon:
maybe just replace `good` with `better`

apsavin:
@ljharb makes sense, thanks.

PR #1188, opened 2016-11-29
Remove "no-duplicate-imports" rule

jongold:
As far as I can tell, `no-duplicate-imports` is _replaced_ by `[import/no-duplicates](https://github.com/benmosher/eslint-plugin-import/blob/master/docs/rules/no-duplicates.md)`.

`no-duplicate-imports` breaks with Flow types; `import/no-duplicates` works.

e.g. this should work, but it currently breaks with `no-duplicate-imports` enabled.
```js
import { foo } from './bar';
import type { Baz } from './bar';
```

Reference:
https://github.com/babel/eslint-plugin-babel/issues/59


ljharb:
It's intentional to leave both enabled; you can use `"no-duplicate-imports": 0` to turn it off for your config.

PR #1190, opened 2016-12-01
Avoid i++ in array spreads example

lencioni:
This style guide recommends using `i += 1` over `i++`. Even though this
appears in a "bad" example, the example will be clearest if it is "bad"
in fewer ways.

I removed some trailing whitespace for free.


ljharb:
My thinking had been that with ++, it looks more similar to code users will already have in their codebase. This is fine tho.

Issue #1068, opened 2016-09-12
disapprove forEach


bergus:
The [no-iterators rule](/airbnb/javascript#iterators--nope) seems to be well-reasoned, but the provided example entitles the `forEach` method as `// good`. This doesn't seem to fit together, all what `forEach` is about are side effects, it doesn't `return` anything. The example uses a mutable `sum` variable just like the `// bad` one. I don't see much difference (except for making it look less structured and more object-oriented). Is this really worth it?

I think that example should be entitled as `// bad` as well or deleted altogether. Instead an example for correct usage of `map` should be added.

ljharb:
I'm fine with a better example using `map` - but using `forEach` is strictly MUCH better than a loop, if only because it's a gateway drug to using the rest of the iterator methods.

You'll note that the "good" forEach example is followed by a "best" reduce example - I'm not sure what an example of "map" would add.

bergus:
A gateway drug, aha, well OK. I still wouldn't call it "good", maybe `// better` (than "bad") would be more appropriate.

As for a `map` example, why not just take one of the existing and write
```
// bad
const nums = [1, 2, 3];
const incs = [];
nums.forEach(x => {
  incs.push(x + 1);
});

// good
const nums = [1, 2, 3];
const incs = nums.map(x => x + 1);
```

ljharb:
Adding a "map" example set, alongside the "sum" example set, seems great.

ljharb:
Fixed by #1201.

PR #1200, opened 2016-12-09
editorial - adds map example to iterators

sara-gudeman:
closes issue #1068

### DIFF
--- a/README.md
+++ b/README.md
@@ -1317,6 +1317,9 @@ Other Style Guides
     numbers.forEach(num => sum += num);
     sum === 15;
 
+    const sum = numbers.map(num => num + 1);
+    sum === 15;
+
     // best (use the functional force)
     const sum = numbers.reduce((total, num) => total + num, 0);
     sum === 15;


README.md
```diff
@@ -1317,6 +1317,9 @@ Other Style Guides
     numbers.forEach(num => sum += num);
     sum === 15;
 
+    const sum = numbers.map(num => num + 1);
```
ljharb: I'm not sure what the benefit of this example is - are you implying someone might try to use `map` to add up the numbers in an array? This will produce an array of `[2, 3, 4, 5, 6]`, which is not `=== 15`.

sara-gudeman: Sorry! I was working off the example in the issue. I should have paid more attention the sum itself. The idea was that `forEach` was in some ways suboptimal because it had side effects. The example provided was:

```
// good
const nums = [1, 2, 3];
const incs = nums.map(x => x + 1);
```

However, in retrospect, if it is to match the previous examples exactly, `map` wouldn't work in this case and reduce is the optimal solution. I see why `forEach` was used. Is the example above useful?

sara-gudeman: Sorry! I was taking the example from the issue itself. I realize this was the incorrect place to put it. I will close this PR.

sara-gudeman: Absolutely! I can work on that later today :)

ljharb: (No big deal if you open a new PR, but it'd be nice to reuse this one)

ljharb:
Replaced by #1201.

PR #1201, opened 2016-12-10
Adds new examples to iterators section

sara-gudeman:
closes issue #1068

### DIFF
--- a/README.md
+++ b/README.md
@@ -1309,7 +1309,6 @@ Other Style Guides
     for (let num of numbers) {
       sum += num;
     }
-
     sum === 15;
 
     // good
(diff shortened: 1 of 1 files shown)


ljharb:
Thanks!

churchie317:
Not to be pedantic, but should we consider calling the mutated arrays something other than `modified`? Can I recommend `newArray`?

I just wonder if it might confuse newer users who would think that it's the original arrays that are being modified.

ljharb:
Sure, a different name is fine. Maybe instead of `newArray`, something like `increasedByOne`?

churchie317:
#1202 Hope this helps!

PR #1202, opened 2016-12-11
Change "modified" to "increasedByOne" to minimize ambiguity

churchie317:
issue #1068


churchie317:
Also, am I referencing this correctly?

PR #1206, opened 2016-12-14
Add gulpfile.babel.js (ie ES6 gulpfiles)

mikelambert:


ljharb:
Thanks! This is better solved by #1204, though.

PR #1208, opened 2016-12-15
Added flowtype linting support through the use of a new package

milesj:
I use flowtype on all my personal projects, as well as a few projects at Airbnb. It gets pretty tiring to repeat the same setup across all projects, so I thought it would be best to standardize this through a new official Airbnb ESLint package.

I wasn't sure what to call this package, so I called it "eslint-config-airbnb-next", as it implies it's the next stage / future support (and follows the naming of ESNext). Furthermore, it inherits from "eslint-config-airbnb", so includes react, a11y, and import support automatically. Not sure if this is wanted, but it avoids the following 4 packages: base, react, flow, react+flow, but I can switch to that too if need be.

Also didn't modify the readme as I'm not sure how to phrase it.

### DIFF
--- a/package.json
+++ b/package.json
@@ -3,15 +3,18 @@
   "version": "2.0.0",
   "description": "A mostly reasonable approach to JavaScript.",
   "scripts": {
-    "preinstall": "npm run install:config && npm run install:config:base",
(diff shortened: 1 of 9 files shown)


ljharb:
Thank you, but Airbnb officially does not use Flow at this time, so this is very premature on the public repo.

Issue #1212, opened 2016-12-18
react-a11y rules require object input, not array


arsenypoga:
Eslint throws errors about invalid configuration for input values in react-a11y
The input value in:
`'jsx-a11y/anchor-has-content': ['error', ['']]`
should be:
`'jsx-a11y/anchor-has-content': ['error', {components: ['']}],`
[https://github.com/evcohen/eslint-plugin-jsx-a11y/blob/master/docs/rules/anchor-has-content.md](url)
For 
`'jsx-a11y/href-no-hash': ['error',['a'] ],`
where it should be:
`'jsx-a11y/href-no-hash': ['error', { component: ['a'] }],`
[https://github.com/evcohen/eslint-plugin-jsx-a11y/blob/master/docs/rules/href-no-hash.md](url)

arsenypoga:
Another one at:
`'jsx-a11y/label-has-for': ['error', ['label']],`
should've been:
`'jsx-a11y/label-has-for': ['error', {components: ['label']}],`
Another one at:
`'jsx-a11y/heading-has-content': ['error', ['']],`
Should:
`'jsx-a11y/heading-has-content': ['error', {components: ['']}],`

ljharb:
@arsenypoga this is intentional. `eslint-config-airbnb` is only compatible with v2 of `eslint-plugin-jsx-a11y` right now - if you use the `npm install` command in the package readme, you'll get the proper version.

In master, all these rules are already updated to be compatible with v3 of `eslint-plugin-jsx-a11y` and not yet published - but in the meantime, please downgrade to v2.

arsenypoga:
Ok, I see. Well, I switched the values in he jsx-a11y config, and Eslint doesn't have errors anymore, so I can call it success anyways.

ljharb:
Just published v14.0.0 of the main config, and v11.0.1 of the base config.

(#1162, #1166, #1182, #1186, #1207, #1212, #1222)

Issue #1207, opened 2016-12-14
UNMET PEER DEPENDENCY eslint-plugin-jsx-a11y@^2.2.3


khankuan:
Hi, I noticed that the version for eslint-plugin-jsx-a11y shows `^3.0.2` but the npm package installed uses `^2.2.3`.

ljharb:
Yes, that is intentional - the package hasn't been updated yet. If you use the `npm install` command in the readme for `eslint-config-airbnb`, you'll always get the right peer dependency versions installed.

khankuan:
Will there be any new releases?

ljharb:
At some point, certainly. Is there a reason that you need v3 of eslint-plugin-jsx-a11y right now?

khankuan:
No, just got confused why its out of sync with npm. Thanks for update :)

khankuan:
And atom starts popping out error: `Configuration for rule "jsx-a11y/anchor-has-content" is invalid:
	Value "" is the wrong type.`

ljharb:
Atom has a bug where you need to restart it when you make any node_modules change.

khankuan:
Oh interesting, seems like thats the problem. Thanks!

Aid19801:
I've tried restarting Atom and i'm still getting the same error. Goes away if i switch to Google lint. but unfortunately our team are aligned with AirBnB style.  Any ideas how i can fix this relatively quickly? 

![image](https://cloud.githubusercontent.com/assets/15033409/21639788/c498a1a0-d269-11e6-88e0-f382179bcb6f.png)

ljharb:
@Aid19801 do you have v3 of `eslint-plugin-jsx-a11y` installed? Our config isn't compatible with that yet - stick to v2.

UnsungHero97:
@ljharb is a fix for this on the roadmap?

ljharb:
It's not something that needs fixing - it's just the nature of peer deps in npm - but it's up to v3 on master, so whenever the next release happens, it'll be included.

Aid19801:
@ljharb thanks mate, will take a look shortly and come back to you.

ljharb:
Just published v14.0.0 of the main config, and v11.0.1 of the base config.

(#1162, #1166, #1182, #1186, #1207, #1212, #1222)

PR #1186, opened 2016-11-24
Update react-a11y.js

salzhrani:
`jsx-a11y` was bumped to `3.0.1` but `react-a11y.js` was not updated to accommodate the changes.

Issue #1182, opened 2016-11-21
jsx-a11y breaks linter


johntran:
Hi there!

I got the following errors for the following rules:

- jsx-a11y/anchor-has-content
- jsx-a11y/href-no-hash
- jsx-a11y/label-has-for
- jsx-a11y/heading-has-content

They are all type errors. I've removed the rules, and my linter started working again.

For those who want to fix their linter, see below. I wasn't sure what the expected behavior was for those rules, so I didn't attempt to fix the rules. I opted to remove them instead.
https://gist.github.com/johntran/26eb28be5ed30edde5f706525245324c

Setup:
OS: Mac OS X 10.11.6
Atom: 1.12.5
Node: 7.1.0

Atom Packages:
linter-eslint: 8.0.0
Nuclide: 0.183.0

npm packages:
babel-eslint: 7.1.1
eslint-plugin-jsx-a11y: 3.0.1
eslint-config-airbnb: 13.0.0
eslint: 3.10.2

See screenshots. I enabled "linter" atom package to get these screenshots. I disabled "linter" afterwards and Nuclide Diagnostics started picking up eslint errors again.

![screenshot 2016-11-21 05 16 00](https://cloud.githubusercontent.com/assets/3813391/20484190/d9b91d12-afaa-11e6-8453-2e8ade96cf00.png)
![screenshot 2016-11-21 05 17 14](https://cloud.githubusercontent.com/assets/3813391/20484187/d9b4ad72-afaa-11e6-9271-dcecdbdaef03.png)
![screenshot 2016-11-21 05 18 45](https://cloud.githubusercontent.com/assets/3813391/20484188/d9b6823c-afaa-11e6-90d7-ce7d85616f7c.png)
![screenshot 2016-11-21 05 26 44](https://cloud.githubusercontent.com/assets/3813391/20484240/2053c79a-afab-11e6-9d05-4bf60b615712.png)

johntran:
Ah... duplicate of: https://github.com/airbnb/javascript/issues/1163

Closing.

johntran:
Reopening:

https://github.com/airbnb/javascript/commit/72eb3e443e49da634a83b56660fbe56698775cbb

Wasn't this commit supposed to fix it?

andykenward:
@johntran you are referring to a commit that is beyond `eslint-config-airbnb: 13.0.0` version that it looks like you are using.

[...]

PR #1222, opened 2016-12-28
Updates peerDependencies for eslint-config-airbnb

kamronbatman:
Updates peerDependencies for eslint-config-airbnb.
Also since the last NPM release has a severely outdated peerDependency for eslint-plugin-jsx-a11y, can we please get a new NPM release with this update?

Thanks!

### DIFF
--- a/packages/eslint-config-airbnb-base/package.json
+++ b/packages/eslint-config-airbnb-base/package.json
@@ -49,7 +49,7 @@
     "babel-preset-airbnb": "^2.1.1",
     "babel-tape-runner": "^2.0.1",
     "editorconfig-tools": "^0.1.1",
-    "eslint": "^3.12.1",
+    "eslint": "^3.12.2",
     "eslint-find-rules": "^1.14.3",
     "eslint-plugin-import": "^2.2.0",
     "in-publish": "^2.0.0",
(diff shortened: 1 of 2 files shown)


ljharb:
v2 to v3 of eslint-plugin-jsx-a11y is not "severely" outdated; if you read the changelog you'll see the differences are minor.

It is explicitly intentional to not yet have released an update of `eslint-config-airbnb` that supports v3 of `eslint-plugin-jsx-a11y`, and I will release an update when it is ready.

Thanks for the contribution; I'll rebase it down to one commit with the desired commit message convention, and merge it in.

ljharb:
A better solution imo is to never globally install anything, especially a linter, which should always only be installed per-project :-)

kamronbatman:
I agree that is an option.  It is definitely a pain in the ass when it comes to IDE setup.  Webstorm in particular is not smart enough to say... 'Lets find the project root directory, lets find the rc file, then lets find node_modules, then lets find eslint'.

You are right though, just need to manually set it up, sigh.

ljharb:
Just published v14.0.0 of the main config, and v11.0.1 of the base config.

(#1162, #1166, #1182, #1186, #1207, #1212, #1222)

Issue #1089, opened 2016-09-25
What rules does the community tend to override?


ljharb:
I'm opening this issue specifically to gather a list of the rules that those in the community feel are controversial or problematic, such that they override them, and most importantly, *why*.

This is not an invitation for "+1s" (those belong as reactions on individual posts), or an invitation to hear about everyone's subjective aesthetic preferences, nor is it an indication that any of these rules necessarily will - or will not - change. I'm simply hoping to primarily gather civil, well-reasoned explanations for why certain rules are overridfen, disabled, etc.

Any comments mentioning spaces vs tabs, or semicolons, will be summarily deleted :-)

We're aware of the following:
 - `react/jsx-filename-extension` - filed as #985. Why is it controversial? Some believe that `.js` files should be allowed to contain JSX. Others believe this is an app-level concern, and does not belong in a shared config (like "env" settings).

Any others? I'll update the original post with new examples as I'm made aware of them.

jonathanong:
- `arrow-body-parenthesis` or whatever, because it makes converting regular functions to arrow functions difficult as it's not fixed automatically. 
- `camelcase` - because many times i'm bound by other APIs
- `no-param-reassign` - makes working with `koa`/`express`/node.js difficult
- `no-plusplus` - because `i += 1` is annoying
- `prefer-template` - because it isn't convenient 5% of the time
- `consistent-return` - annoying with early returns
- `new-cap` - bound by other APIs
- `no-underscore-dangle` - disallows "private" properties. bound by mongodb.
- `no-use-before-define` - disallows function declarations at the bottom

ljharb:
@jonathanong Thanks! I'd forgotten `camelcase`, we have that problem too. Re `arrow-body-style`, would the existence of an autofixer make it not be a problem?

Can you elaborate on when `prefer-template` isn't convenient?

ljharb:
Thanks - I've added a few of your examples to the OP. (and fixed your markdown, hope you don't mind)

chadwatson:
`no-use-before-define` for functions

steida:
https://github.com/este/este/blob/master/.eslintrc

```
// Soft some rules.
"global-require": 0, // Used by webpack-isomorphic-tools and React Native.
"new-cap": [2, {"capIsNew": false, "newIsCap": true}], // For Record() etc.
"no-class-assign": 0, // Class assign is used for higher order components.
"no-nested-ternary": 0, // It's nice for JSX.
[...]

PR #1224, opened 2016-12-28
linebreak-style makes it impossible for Windows users to contribute

paulcbetts:
Any project that currently uses eslint-config-airbnb checked out on Windows results in thousands of linter errors since the line endings will be CRLF, making it impossible to contribute to these projects as a Windows user without nerfing the linter entirely.

A better way to enforce this is not via a linter, but via a .gitattributes file (https://help.github.com/articles/dealing-with-line-endings). When you do this, Git will ensure the repo internally is LF, but on Win32 will check out files as CRLF (this is a builtin case of a clean/smudge filter)

Refs zeit/hyper#1230, https://github.com/sindresorhus/eslint-config-xo/pull/35

### DIFF
--- a/packages/eslint-config-airbnb-base/rules/style.js
+++ b/packages/eslint-config-airbnb-base/rules/style.js
@@ -121,10 +121,6 @@ module.exports = {
       applyDefaultPatterns: true,
     }],
 
-    // disallow mixed 'LF' and 'CRLF' as linebreaks
-    // http://eslint.org/docs/rules/linebreak-style
-    'linebreak-style': ['error', 'unix'],
-
     // enforces empty lines around comments
     'lines-around-comment': 'off',
 


SimenB:
https://github.com/airbnb/javascript/issues/1089#issuecomment-254025333

Also, it should be set to `off`, not removed.

ljharb:
Indeed, https://github.com/airbnb/javascript/issues/1089#issuecomment-254025333 is the official response - all users should always only use LF, including on Windows.

paulcbetts:
@ljharb This is not how Git works on Windows, can you please read https://help.github.com/articles/dealing-with-line-endings? What you're asking is imposing a huge burden on Windows users. To contribute to any AirBnB repo, users will have to:

1. Edit their *system-wide Git settings* to set autocrlf=false
1. Check out the repo (remember, it can't be done after-the-fact, because all the files will be checked out as CRLF)
1. Remember to set autocrlf=true back before working with any other repo on their system.
1. Continue to remember to toggle this setting back-and-forth constantly

You're entirely correct that editors on Windows will be perfectly happy to work in LF, but this isn't an editor problem - this is a fundamental conflict with how Git works on Windows. I don't know why ESLint added this feature but imho I agree with @SimenB, the only thing ESLint should be checking for is _inconsistent_ line endings.

If I add this feature to ESLint, will you consider using it? Line Ending Purity isn't worth throwing 50%+ of node.js users under the bus.

ljharb:
Certainly a `consistent` option would make sense to me, and I'd vastly prefer a discussion about changing "LF only" to "consistent, LF preferred" over the current one of "LF only" vs "off". (however, I'd want a way where on non-Windows systems, CRLF was strictly forbidden - there's no reason for a CRLF character to exist outside of a Windows machine)

What I'm suggesting is that every repo on their system, and that they work with, should be using LF, and not CRLF. I agree that it's highly inconvenient that they'd have to re-clone all their repos after setting their systemwide Git settings to `autocrlf=false` - but I'm then suggesting that they'd never set it back to `true`, and that any Windows projects they work with should be using only LF as well.

What are the arguments (besides "legacy", which carries weight but isn't the best reason to make choices) why any project should ever use CRLF?

paulcbetts:
> What are the arguments (besides "legacy", which carries weight but isn't the best reason to make choices) why any project should ever use CRLF?

Whether LF or CRLF is _philosophically right_ or not doesn't actually matter, what matters is that this setting system-wide is both non-default (and relatively rare tbh), and will definitely *corrupt repos* that have CRLF encoded in the repo (also non-default, but far more common, from the last time I ran the statistics on the top 250 C# repos, as part of my work on the Windows GitHub client). 

So in the _very best_ case scenario, you're asking every Windows user to reclone every repo on their system (and to understand all of these nuances around line endings), and in the worst case, you're encouraging users to set a setting that will corrupt every PR they submit to certain repos.

In the meantime, I'm _desperately_ trying to get Windows users to contribute to node.js projects - all I hear from JS maintainers are "Ugh, Windows people never contribute, they just complain". Help me fix that by getting this obstacle out of the way!

ljharb:
If the "consistent" option were available, I'd be willing to conditionally set it to "unix" or "consistent, unix preferred" based on `process.platform === 'win32'` - that way at least non-Windows users (and almost every CI system) would require LF, but windows users would have a more relaxed setting.

Do you think that would address the line endings footguns on Windows you're describing, without completely sacrificing the desired guarantees?

paulcbetts:
That would do it, yeah - lemme work on an ESLint PR today

paulcbetts:
@ljharb https://github.com/eslint/eslint/pull/7823 should get us what we need here, then we can write something like:

```js
"linebreak-style": ["error", process.platform === "win32" ? "consistent" : "unix"]
```

paulcbetts:
I'm not interested in pursuing this further. Thanks all.

ljharb:
I'm sorry to hear that.

jacobq:
I'm not interested in having / starting a flame war, but to anyone who thinks there is no use case for allowing / supporting different EOL encodings, I can tell you that as embedded software engineer I have to work with it all the time as many devices (typically closed-source / black boxes) require it as part of their interface / API (e.g. data sent via USB/serial port or TCP). Just because you have the privilege of working in an environment where you can have control and can make things the way you want does not mean that all developers can or should. Having consistent code standards is great, but if it's taken too far the cost/benefit ratio can explode due to lost productivity dealing with minutia.

ljharb:
@jacobq Unless you do your *typing* on those machines, I'm not sure I see the problem - you can always transform your code before you load it onto a device.

daxelrod:
The [sanest way](https://github.com/daxelrod/jowl/commit/ea0ad1c6c6726c32b548ec6ce0cf04ece27d7d4f) I've found to configure projects so that this rule is followed on Windows is:

A `.gitattributes` in the project with

```
* text=auto eol=lf
```

which will represent files as LF in the object database and write them with LF on checkout, and

A `.editorconfig` in the project with

```
end_of_line = lf
```

to cause many editors to initially create files with LF endings.

ljharb:
An editorconfig is a great idea, as soon as you can hook up a tool that can enforce that the editorconfig doesn't drift - iow, that it matches (or is compatible with) your eslint config, and all your files comply with it. I'm not aware of a reliable one.

iamakulov:
Folks, what’s the status of this? I’d love to see this merged.

A lot of projects use the AirBnB config, and it’s very annoying to constantly add `linebreak-style: 'off'` to their `.eslintrc` (and be careful to not commit this change) when I try to contribute to them.

ljharb:
@iamakulov at the moment, it's blocked by eslint's refusal to allow https://github.com/eslint/eslint/pull/7823 to be merged.

iamakulov:
@ljharb Thanks!

While ESLint is blocking this, what if we solve this issue with 

```
"linebreak-style": ["error", process.platform === "win32" ? "windows" : "unix"]
```

?

My thinking is as follows:
1. By default, Git uses CRLF for Windows files and LF for Unix files. That code ↑ matches this default behavior.
2. The default behavior could be changed either by i) adding a `.gitattributes` file or ii) changing the `core.autocrlf` option globally.
   1. The `.gitattributes` is added by a maintainer. The maintainer would know the effect of the `* eol=lf` statement and would likely remember to change the `linebreak-style` rule accordingly. Even if they won’t, Windows users would report a new ESLint error, and the maintainer would likely remember that they added `.gitattributes` before that.
   2. The `core.autocrlf` option is changed by a specific user. The user would know what they do, so if they configure Git to check out in LF, they’d likely understand why ESLint tells them line endings should be CRLF. In this case, they might either disable this setting or submit a PR with the `.gitattributes` files.

This would keep the majority of people that use the default settings happy. The power users receiving the error should understand why it’s happening.

ljharb:
@iamakulov the concern there is then a repo might not add the .gitattributes file in the first place, and conflicts could arise between windows and linux developers using the airbnb config.

iamakulov:
> the concern there is then a repo might not add the .gitattributes file in the first place, and conflicts could arise between windows and linux developers using the airbnb config.

Could you elaborate on that please? AFAIK, the repo (not the local checked-out code, but the commits) would always be in LF, so there should be no conflicts.

iamakulov:
> AFAIK, the repo (not the local checked-out code, but the commits) would always be in LF

I.e.:
* On Windows, Git uses `core.autocrlf=true` by default. This keeps local (checked out) files in CRLF, but converts them to LF when they are committed.
* On UNIX, Git uses `core.autocrlf=input` by default. This keeps local (checked out) files in LF + makes sure the files are also committed with LF (i.e. if CRLF is encountered, it’s converted to LF).

Here’s the related piece of docs: https://git-scm.com/book/en/v2/Customizing-Git-Git-Configuration (Ctrl+F → “core.autocrlf”)

So, _the repo_ (as in _the commits_) are always kept in LF by default, but _local files_ use CRLF on Windows and LF on Unix. This is what this rule matches:

```
"linebreak-style": ["error", process.platform === "win32" ? "windows" : "unix"]
```

ljharb:
@iamakulov my understanding is that the behavior you're describing only applies if windows users have set up their git properly using autocrlf; the documentation you linked doesn't say what the default is, but I'm relatively sure the motivation for https://github.com/eslint/eslint/pull/7823 was that it's not the default behavior, and that once changed, you have to delete and re-clone any previously cloned repo to fix it.

Issue #1233, opened 2017-01-06
Tool to set up eslint-config-airbnb without bash


nathanhleung:
I saw the creator of #1213 was having problems installing all the peer dependencies for the config on Windows, since the bash command doesn't work. I was having the same type of trouble, but I didn't want to manually copy the version of each peer dependency _every_ time I wanted to start a new project.

I wrote up a generalized cli (which works on all platforms, not just Windows!) to help install peerDeps (inspired by this use case) called [install-peerdeps](https://github.com/nathanhleung/install-peerdeps) - let me know if this helps!

ljharb:
Thanks, this looks great! What about `--save` and `--save-dev`? The usefulness isn't just installing it, it's putting the proper versions in your package.json.

ljharb:
Once that exists, I'd love a PR into the readmes for both packages that suggests using `install-peerdeps` as an alternative to the bash command.

nathanhleung:
I've got a `--dev` flag for devDependencies, but even better, it detects whether the project is running yarn or npm, and then runs either `yarn add` or `npm install --save`!

(here are the full usage instructions)
```
Usage: install-peerdeps <package>[@<version>], default version is 'latest'

Installs the specified package along with correct peerDeps.

Options:

  -h, --help     output usage information
  -V, --version  output the version number
  -d, --dev      Install the package as a devDependency
```

I'll submit a PR as soon as I can!

ljharb:
I'd definitely prefer it not to autodetect things - projects can use both npm and yarn, and I'd want to recommend a tool that asked the user to make the choice. Similarly, I'd want "modify package.json or yarn.lock" not to be the default; users should have to ask for it.

nathanhleung:
Got it, I'll add those options to the cli and let you know.

nathanhleung:
One note though,, Yarn purposely prohibits you from installing a "silent dependency" (i.e. one that is missing from package.json). In other words, there's [no alternative to `npm install`](https://yarnpkg.com/en/docs/migrating-from-npm) in Yarn and package.json/yarn.lock is always edited when a package is installed.

However, I'll add the yarn/npm option to the cli - if a user wants to install the eslint config silently, they'll just have to use npm.

ljharb:
That's fair - but npm users are the far larger group, and I don't want their expectations surprised :-)

nathanhleung:
I've added two new options, `--yarn` and `--silent`:

```
[...]

PR #1234, opened 2017-01-06
add note about install-peerdeps to readmes

nathanhleung:
discussed in #1233, windows users can use install-peerdeps to automatically install peerdeps for the config

### DIFF
--- a/packages/eslint-config-airbnb-base/README.md
+++ b/packages/eslint-config-airbnb-base/README.md
@@ -32,6 +32,19 @@ Our default export contains all of our ESLint rules, including ECMAScript 6+. It
     npm install --save-dev eslint-config-airbnb-base eslint@^#.#.# eslint-plugin-import@^#.#.#
   ```
 
+  Windows users can either install all the peer dependencies manually, or use the [install-peerdeps](https://github.com/nathanhleung/install-peerdeps) cli tool.
+
+  ```sh
+  npm install -g install-peerdeps
+  install-peerdeps --dev eslint-config-airbnb-base
+  ```
+
+  The cli will produce and run a command like:
+
+  ```sh
+  npm install --save-dev eslint-config-airbnb-base eslint@^#.#.# eslint-plugin-import@^#.#.#
+  ```
+
 2. Add `"extends": "airbnb-base"` to your .eslintrc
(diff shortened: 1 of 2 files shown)


SimenB:
Could something other than `#` be used? It's rendered as comments.

image

ljharb:
Sure, suggestions welcome - I just don't want it to be a valid command

nathanhleung:
@SimenB  If you omit the ```sh at the start of the block it'll disable the syntax highlighting:

With ```sh
```sh
npm install --save-dev eslint-config-airbnb eslint@^#.#.# eslint-plugin-jsx-a11y@^#.#.# eslint-plugin-import@^#.#.# eslint-plugin-react@^#.#.#
```

Without ```sh
```
npm install --save-dev eslint-config-airbnb eslint@^#.#.# eslint-plugin-jsx-a11y@^#.#.# eslint-plugin-import@^#.#.# eslint-plugin-react@^#.#.#
```

Do you want to submit a PR for that?

ljharb:
Sure, that seems fine too.

vamsiampolu:
@ljharb the install script is really great, but I noticed two things:

+ it does not install them as dev dependencies

+ it does not install the `eslint-config-airbnb` package itself

So, I made a few changes:

```sh
(
  export PKG=eslint-config-airbnb;
  npm info "$PKG@latest" peerDependencies --json | command sed 's/[\{\},]//g ; s/: /@/g ' | xargs npm install --save-dev;
  npm install eslint-config-airbnb --save-dev
)
```

ljharb:
@vamsiampolu As you can see [here](https://github.com/airbnb/javascript/tree/master/packages/eslint-config-airbnb#eslint-config-airbnb-1) and [here](https://github.com/airbnb/javascript/tree/master/packages/eslint-config-airbnb-base#eslint-config-airbnb-base-1), the command is:
```sh
(
  export PKG=eslint-config-airbnb;
  npm info "$PKG@latest" peerDependencies --json | command sed 's/[\{\},]//g ; s/: /@/g' | xargs npm install --save-dev "$PKG@latest"
)
```

which already absolutely installs the package itself, and installs everything as dev dependencies.

PR #1251, opened 2017-01-11
Update 4.5 documentation for good/bad flow consistency

oshalygin:
- This commit reorganizes the bad and good examples to
  match the flow with the rest of the documentation.
- The general flow is that a bad example is provided
  followed by a good or best example.
- To match the same convention, the good examples
  are funneled to the bottom of the block so that
  users can understand the wrong approach first

### DIFF
--- a/README.md
+++ b/README.md
@@ -370,15 +370,6 @@ Other Style Guides
   - [4.5](#arrays--callback-return) Use return statements in array method callbacks. It's ok to omit the return if the function body consists of a single statement following [8.2](#8.2). eslint: [`array-callback-return`](http://eslint.org/docs/rules/array-callback-return)
 
     ```javascript
-    // good
-    [1, 2, 3].map((x) => {
-      const y = x + 1;
-      return x * y;
-    });
-
-    // good
-    [1, 2, 3].map(x => x + 1);
-
     // bad
     const flat = {};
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -370,15 +370,6 @@ Other Style Guides
   - [4.5](#arrays--callback-return) Use return statements in array method callbacks. It's ok to omit the return if the function body consists of a single statement following [8.2](#8.2). eslint: [`array-callback-return`](http://eslint.org/docs/rules/array-callback-return)
 
     ```javascript
-    // good
```
ljharb: in this case, there's no bad "map" example. these are "map, reduce, filter", and should be grouped as such.

oshalygin: @ljharb  Agreed, I guess the only other scenario would be to have `.map` perform some kind of side effects.  

I can totally go both ways on this and appreciate your feedback but would you care to have this change in the documentation where `.map` is moved to the bottom of the block?  The benefit here would be to first observe things you shouldn't do with filter/reduce followed by good practices for both, followed with straight up good practices with `.map`.

I can change the commit message/title accordingly, but if think it's unnecessary, I can close this out as well.  

Thanks for the feedback

ljharb: No, I think `.map` belongs at the top.

PR #1253, opened 2017-01-11
typo in switch

wonism:
typo in 15.5 comparison--switch-blocks


ljharb:
This is actually intentional, since case blocks only need braces when they use a block-scoped variable.

PR #1254, opened 2017-01-11
typo in 15.5comparison--switch-blocks

wonism:
typo in 15.5comparison--switch-blocks


ljharb:
(from #1253)

ljharb:
This is actually intentional, since case blocks only need braces when they use a block-scoped variable.

PR #1255, opened 2017-01-12
[breaking] `padded-blocks`: enable `allowSingleLineBlocks` option

### LINKED ISSUE #1240: Airbnb's `padded-blocks` ESLint rule definition may be incorrect
In the Airbnb style guide, [section 18.8](https://github.com/airbnb/javascript#whitespace--padded-blocks) states to "not pad your blocks with blank lines". To enforce this via ESLint, the `padded-blocks` rule is used in [style.js](https://github.com/airbnb/javascript/blob/master/packages/eslint-config-airbnb-base/rules/style.js#L317).

Currently, `padded-blocks` is set to `['error', 'never']`. This, however, means that `switch` statements and `class` declarations are not being checked correctly (the documentation around this was a bit poor - see [this](https://github.com/eslint/eslint/issues/7868) issue).

In order to correctly enforce this for _all_ blocks, the `padded-blocks` rule needs to be set to `['error', { blocks: 'never', classes: 'never', switches: 'never' }]`. :)

91aman:
This Pull request is to fix #1240. This will enable no padding for switch statements and class declarations along with blocks.

Reference : http://eslint.org/docs/rules/padded-blocks

### DIFF
--- a/packages/eslint-config-airbnb-base/rules/style.js
+++ b/packages/eslint-config-airbnb-base/rules/style.js
@@ -419,12 +419,12 @@ module.exports = {
     'operator-linebreak': ['error', 'before', { overrides: { '=': 'none' } }],
 
     // disallow padding within blocks
-    // TODO, semver-major: uncomment option
     'padded-blocks': ['error', {
       blocks: 'never',
       classes: 'never',
       switches: 'never',
-      // allowSingleLineBlocks: true,
+    }, {
+      allowSingleLineBlocks: true,
     }],
 
     // Require or disallow padding lines between statements


ljharb:
It'd be ideal to at least do this in two commits - one that's a noop; converting the string form to the object form - and then one that adds the semver-minor change.

91aman:
@ljharb Sorry, I did not understand what you mean by `and then one that adds the semver-minor change.`

packages/eslint-config-airbnb-base/rules/style.js
```diff
@@ -314,7 +314,7 @@ module.exports = {
     'operator-linebreak': 'off',
 
     // enforce padding within blocks
-    'padded-blocks': ['error', 'never'],
+    'padded-blocks': ['error', { blocks: 'never', classes: 'never', switches: 'never' }],
```
ljharb: i mean that "classes" and "switches" should be set to null, or omitted - and that in a second commit/PR, they should be set to "never".

PR #1261, opened 2017-01-17
Update README.md

jinxac:
Looks like there is a typo.


gergoerdosi:
The original seems correct to me.

jinxac:
Yes it is @gergoerdosi  I read it incorrectly I think.. MY bad

PR #1263, opened 2017-01-17
[guide] Correct erroneous example

vsemozhetbyt:
Currently, the bad example in the [8.2](https://github.com/airbnb/javascript/blob/master/README.md#arrows--implicit-return) is just a wrong example:
```js
[1, 2, 3].map(number => {
});

// returns [ undefined, undefined, undefined ]
```
Maybe this is a copy-paste artifact. I dare to assume and restore what it should be initially.

Also, a trailing comma is added in the object literal.

### DIFF
--- a/README.md
+++ b/README.md
@@ -883,8 +883,7 @@ Other Style Guides
     ```javascript
     // bad
     [1, 2, 3].map(number => {
-      const nextNumber = number + 1;
-      `A string containing the ${nextNumber}.`;
+      return `A string containing the ${number}.`;
     });
 
     // good
@@ -898,7 +897,7 @@ Other Style Guides
 
     // good
     [1, 2, 3].map((number, index) => ({
-      [index]: number
+      [index]: number,
     }));
     ```
 


README.md
```diff
@@ -883,8 +883,7 @@ Other Style Guides
     ```javascript
     // bad
     [1, 2, 3].map(number => {
```
ljharb: This is intentional, please revert it.

vsemozhetbyt: Bad will not with this line the bad example be almost the same as the good one? Compare:
```js
// bad
[1, 2, 3].map(number => {
});
```
```js
// good
[1, 2, 3].map((number) => {
});
```

ljharb: Right. The intention is to call out the confusion that definitely happens between implicit and explicit return in arrow functions. In this case, the bad example is because it's forgetting the `return` keyword.

vsemozhetbyt: Sorry, I don't understand then what I should save in the commit. Should I remove `return` so that the code will throw as the other `reduce` example? Or should I just close the PR?

ljharb: The only change I think is good in this particular one is the added trailing comma.

vsemozhetbyt: Then I would better close it and add the comma in the #1267.

ljharb: Sounds good, thanks!

PR #1267, opened 2017-01-18
Cleanup code examples

vsemozhetbyt:
Run `airbnb-base` ESLint rules set against code examples (only accidental non-educational errors are fixed).

* `'''js` -> `'''javascript` for consistency.
* Fix indentation for consistency.
* Comment out ellipses.
* `const` -> `let`.
* Add / remove semicolons.
* Add spaces.
* Add trailing commas.
* Remove  trailing or leading underscores.
* Fix arrow IIFE call (currently, it throws).

### DIFF
--- a/README.md
+++ b/README.md
@@ -2073,17 +2073,17 @@ Other Style Guides
     ```javascript
     // bad
     function foo() {
-    ∙∙∙∙const name;
+    ∙∙∙∙let name;
     }
 
     // bad
     function bar() {
-    ∙const name;
+    ∙let name;
     }
 
     // good
     function baz() {
-    ∙∙const name;
+    ∙∙let name;
     }
     ```
 
--- a/README.md
+++ b/README.md
@@ -2546,10 +2546,10 @@ Other Style Guides
     }());
 
     // good, but legacy (guards against the function becoming an argument when two files with IIFEs are concatenated)
-    ;(() => {
+    ;((() => {
       const name = 'Skywalker';
       return name;
-    }());
+    })());
     ```
 
     [Read more](https://stackoverflow.com/questions/7365172/semicolon-before-self-invoking-function/7365214%237365214).
--- a/README.md
+++ b/README.md
@@ -907,7 +907,7 @@ Other Style Guides
 
     > Why? It shows clearly where the function starts and ends.
 
-    ```js
+    ```javascript
     // bad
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -696,7 +696,7 @@ Other Style Guides
     > Why? They are confusing to reason about.
 
     ```javascript
-    var b = 1;
+    let b = 1;
```
ljharb: These `vars` are intentional, and the examples do not need to execute nor pass a linter. Please excise this commit.

README.md
```diff
@@ -2545,7 +2545,8 @@ Other Style Guides
       return name;
     }());
 
-    // good, but legacy (guards against the function becoming an argument when two files with IIFEs are concatenated)
+    // good, but legacy (guards against the function becoming an argument
```
ljharb: there's no need to break this up. please excise this commit.

README.md
```diff
@@ -602,7 +602,7 @@ Other Style Guides
 
     ```javascript
     // immediately-invoked function expression (IIFE)
-    (function () {
+    (function iife() {
```
ljharb: IIFEs are rarely named. Let's excise this commit as well.

README.md
```diff
@@ -811,10 +811,10 @@ Other Style Guides
     console.log(...x);
 
     // bad
-    new (Function.prototype.bind.apply(Date, [null, 2016, 8, 5]));
+    const date = new (Function.prototype.bind.apply(Date, [null, 2016, 8, 5]));
```
ljharb: This doesn't add any value (if you're applying a linter rule, it'd error on the duplicate const definition anyways). Please excise this commit.

README.md
```diff
@@ -2550,7 +2550,7 @@ Other Style Guides
     ;(() => {
       const name = 'Skywalker';
       return name;
-    }());
+    })();
```
ljharb: This would need an additional set of parens wrapping the entire IIAFE - the invoking parens and the function being invoked must be a single unit.

vsemozhetbyt: So should it be
```js
((() => {
  const name = 'Skywalker';
  return name;
})());
```
?

ljharb: Yes.

ljharb:
I'd prefer a rebase; I don't plan to squash anything.

vsemozhetbyt:
OK, I will try, but I am not very good in rewriting history by git.

ljharb:
Thanks!

PR #1268, opened 2017-01-18
Improve getter/setter regex for react/sort-comp

iancanderson:
Before, the getter/setter regex was matching on methods named like `setFoo` and `getFoo`, which resulted in confusing warnings like:
```
error  methodFoo should be placed after setFoo  react/sort-comp
```

After this change, the regex only matches lines with whitespace in between `get` or `set` and the name of the setter.

### DIFF
--- a/packages/eslint-config-airbnb/rules/react.js
+++ b/packages/eslint-config-airbnb/rules/react.js
@@ -208,7 +208,7 @@ module.exports = {
         'static-methods',
         'lifecycle',
         '/^on.+$/',
-        '/^(get|set)(?!(InitialState$|DefaultProps$|ChildContext$)).+$/',
+        '/^(get|set)\s(?!(InitialState$|DefaultProps$|ChildContext$)).+$/',
         'everything-else',
         '/^render.+$/',
         'render'


ljharb:
Can you provide some example method lists (in a passing sorting) before, and after this change?

iancanderson:
@ljharb sure!

## Passing before:

```javascript
get foo() {}

getFoo() {}

methodFoo() {}
```

```javascript
set foo() {}

setFoo() {}

methodFoo() {}
```

## Passing after:

```javascript
get foo() {}

getFoo() {}

methodFoo() {}
```

```javascript
// failed before
get foo() {}

methodFoo() {}

getFoo() {}
```

```javascript
set foo() {}

setFoo() {}

methodFoo() {}
```

```javascript
// failed before
set foo() {}

methodFoo() {}

setFoo() {}
```

ljharb:
This change seems to break the intended behavior. The intention is to sort explicit getter/setter methods like "getFoo" and "setFoo" (as well as builtin getters/setters, which this guide forbids, so you shouldn't be using) before all other methods.

iancanderson:
@ljharb Whoops! I did not realize the guide forbids getters and setters. That does indeed make this change no good 😄 
For future reference, here is the relevant rule: https://github.com/airbnb/javascript#accessors--no-getters-setters

ljharb:
fwiw tho, even if we did allow getters and setters, we'd still want `getFoo` and `setFoo` methods to be grouped with them, above other methods.

PR #1269, opened 2017-01-18
Update russian translation

leonidlebedev:
:wave: 
I've made new Russian translation for ES6. I would like to add a link to it. The old one is for ES5 and shouldn’t be used.


ljharb:
Thanks! I'll have a Russian speaker review it, and then we'll merge :-)

PR #1278, opened 2017-01-24
The good case example on switch blocks did not use braces.

k1rd3rf:
`case 4` did not use braces.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1818,9 +1818,10 @@ Other Style Guides
         }
         break;
       }
-      case 4:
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -1818,9 +1818,10 @@ Other Style Guides
         }
         break;
       }
-      case 4:
+      case 4: {
```
ljharb: This is intentional - only `case` statements that have block-level variables need braces.

PR #1298, opened 2017-02-13
add more explanation on the `class vs React.createClass vs stateless` section

ivanzusko:
Normal function component has 1 kB smaller weight than class-based one (it could be noticed on the final bundle size), so more components built as normal functions - the smaller bundle size will be.


ljharb:
While this may be true, I don't think that filesize should ever be a reason one chooses a programming construct.

(separately, the PR would need a rebase)

ivanzusko:
@ljharb , of course it shouldn't . This just one more small argument in favor to choose one over another.

ljharb:
Although SFCs are great, and this performance argument favors them, it's not consistent to pick and choose some perf arguments to give credence to :-) It's far better to discount them all, and always choose the most conceptually appropriate and clean solution.

PR #1304, opened 2017-02-14
Changed from array to single value

scott-joe:
This being an array with only one value was causing an error in Atom's internal linter. It would throw errors on the 1st line of code stating there was no rule defined for `no-multi-assign`. When I added a rule in my local `.eslintrc.json` it simply delivered the same error in whatever level I declared. The only way for me to get rid of it was to set it to `0`; which was less than ideal since I wanted it to be where you have it, at `2` (`error`).

I eventually traced it all the way back to this config as our own config extends this one, changed the array to a string in the live project and it finally stopped throwing inline errors in Atom.

I will say that it's odd that the CLI didn't report an issue...so perhaps it's just an Atom interpretation issue of having an array with no other values.

### DIFF
--- a/packages/eslint-config-airbnb-base/rules/style.js
+++ b/packages/eslint-config-airbnb-base/rules/style.js
@@ -233,7 +233,7 @@ module.exports = {
 
     // disallow use of chained assignment expressions
     // http://eslint.org/docs/rules/no-multi-assign
-    'no-multi-assign': ['error'],
+    'no-multi-assign': 'error',
 
     // disallow multiple empty lines and only one newline at the end
     'no-multiple-empty-lines': ['error', { max: 2, maxEOF: 1 }],


ljharb:
This should work just fine - a single severity value works identically in eslint whether it's in an array or not. I'd prefer to file a bug on Atom, rather than modifying this package to work around it.

scott-joe:
@ljharb is there a particular reason this base setting is surrounded by brackets? Does that somehow imply that it can accept more values when simple string values do not?

ljharb:
It means that if I add an extra option later, I don't have to add the array brackets. Often I use that for consistency in my config files, so that every eslint rule always has brackets (that's not the case here atm, but it could be)

scott-joe:
Excellent. Thanks for the clarity. I appreciate your taking the time to help me understand.

I discussed this issue with the engineer who put together our eslint module and he explained that what was happening was that while our module pins to a specific version of eslint and your config, your changed to use looser dependency pinning than before. This created an *eslint* issue where it was, in so many poorly worded sentences, that it had a rule with no definition within *eslint*.

So, this policy of pushing minor and patch updates through to a specific version of the main module is good for keeping things up to date, but introduces the potential for breaking changes. Our solution was to let our version of eslint float a little more to catch the changes you're providing.

I only deliver all this in the likelihood that you find another wayward engineer with an editor throwing a fit over eslint errors that appear to trace back to your rule.

Thanks again for your patience and support. And thank you for publishing this module. It's a great service to the community.

ljharb:
Generally you don't want to be pinning specific versions anywhere but in a top-level app - everything else should use `^` semver ranges.

I'm going to close this for now, but please feel free to keep discussing here :-)

PR #1314, opened 2017-02-25
[resources] add link to neutrino preset

hshoff:
[Neutrino](https://neutrino.js.org/) is a cool new project from Mozilla for creating and building modern JS apps with zero configuration.

[What is the added value versus all the boilerplate projects out there like create-react-app?](https://neutrino.js.org/FAQ.html#what-is-the-added-value-versus-all-the-boilerplate-projects-out-there-like-create-react-app) is worth a read.

Neutrino already has an Airbnb base preset here: https://neutrino.js.org/presets/neutrino-preset-airbnb-base/ and supports customization as outlined in this "turning off semis required" example: https://neutrino.js.org/presets/neutrino-preset-airbnb-base/#simple-customization

### DIFF
--- a/README.md
+++ b/README.md
@@ -3099,6 +3099,7 @@ Other Style Guides
     + [ESlint](http://eslint.org/) - [Airbnb Style .eslintrc](https://github.com/airbnb/javascript/blob/master/linters/.eslintrc)
     + [JSHint](http://jshint.com/) - [Airbnb Style .jshintrc](https://github.com/airbnb/javascript/blob/master/linters/.jshintrc)
     + [JSCS](https://github.com/jscs-dev/node-jscs) - [Airbnb Style Preset](https://github.com/jscs-dev/node-jscs/blob/master/presets/airbnb.json) (Deprecated, please use [ESlint](https://github.com/airbnb/javascript/tree/master/packages/eslint-config-airbnb-base))
+  - Neutrino preset - [neutrino-preset-airbnb-base](https://neutrino.js.org/presets/neutrino-preset-airbnb-base/)
 
 **Other Style Guides**
 


ljharb:
Who maintains this preset?

Why does it need to be different than the eslint config? Does neutrino use something different from eslint for linting and autofixing?

hshoff:
@ljharb it's a [neutrino config module](https://github.com/mozilla-neutrino/neutrino-dev/blob/master/packages/neutrino-preset-airbnb-base/src/index.js) that makes it easy to add ESLint and set the rules to the Airbnb base preset for a neutrino project. [package.json](https://github.com/mozilla-neutrino/neutrino-dev/blob/master/packages/neutrino-preset-airbnb-base/package.json#L17)

Looks like it's maintained by @eliperelman (:wave: Hi Eli)

ljharb:
That package.json is lacking the proper peer dependencies - `eslint-plugin-import` is required for the base config, for example, and eslint itself needs to be available with the proper version.

hshoff:
that comes from the [neutrino-lint-base](https://github.com/mozilla-neutrino/neutrino-dev/blob/master/packages/neutrino-lint-base/package.json) package

eliperelman:
@hshoff hello!

> Why does it need to be different than the eslint config? Does neutrino use something different from eslint for linting and autofixing?

No, we don't use something different, it's merely a preset for plugging in the Airbnb style guide into a Neutrino project, which is Webpack-based. Behind the scenes it is still using the eslint-loader and thereby ESLint.

Put simply, it gives Neutrino users a fast path to lint quality via `yarn add --dev neutrino-preset-airbnb-base`, and adding it to their list of other presets. I'd be happy to answer any questions if you have any.

ljharb:
@eliperelman thanks for the explanation! the dependencies still need to be fixed (`npm install neutrino-preset-airbnb-base && npm ls` will exit nonzero) but that all makes sense.

PR #1320, opened 2017-03-02
Changed methods with '#' to '.' for consistency

Sylphony:
* `Array#push` to `Array.push`
* `Function#bind` to `Function.bind`


ljharb:
This is intentional. `Array#push` is documentational convention for `Array.prototype.push`, which is *not the same* as `Array.push`.

PR #1335, opened 2017-03-10
Arrow function should not use as props in Onclick event

reetesha:
I have below 3 suggestion to this on JSX onClick Events - 
    
1. Actually, we don't need to use .bind() or Arrow function in our code. You can simple use `<ul onClick={this.handleClick}>` in your code.
    
2. You can also move onClick event from th(or ul) to tr(or li) to improve the performance. Basically you will have n number of "Event Listeners" for your n li element.
    
        So finally code will look like this:
```jsx
        <ul onClick={this.onItemClick}>
            {this.props.items.map(item =>
                   <li key={item.id} data-itemid={item.id}>
                       ...
                   </li>
              )}
        </ul>
    
    // And you can access `item.id` in `onItemClick` method as shown below: 
    
        onItemClick = (event) => {
           console.log(e.target.getAttribute("item.id"));
        }
```

3. I agree with the approach mention above for creating separate React Component for ListItem and List. This make code looks good however if you have 1000 of li then 1000 Event Listeners will be created. Please make sure you should not have much event listener.

```jsx
        import React from "react";
        import ListItem from "./ListItem";
        export default class List extends React.Component {
    
            /**
            * This List react component is generic component which take props as list of items and also provide onlick
            * callback name handleItemClick
            * @param {String} item - item object passed to caller
            */
            handleItemClick = (item) => {
                if (this.props.onItemClick) {
                    this.props.onItemClick(item);
                }
            }
    
            /**
            * render method will take list of items as a props and include ListItem component
            * @returns {string} - return the list of items
            */
            render() {
                return (
                    <div>
                      {this.props.items.map(item =>
                          <ListItem key={item.id} item={item} onItemClick={this.handleItemClick}/>
                      )}
                    </div>
                );
            }
    
        }
    
    
        import React from "react";
    
        export default class ListItem extends React.Component {
            /**
            * This List react component is generic component which take props as item and also provide onlick
            * callback name handleItemClick
            * @param {String} item - item object passed to caller
            */
            handleItemClick = () => {
                if (this.props.item && this.props.onItemClick) {
                    this.props.onItemClick(this.props.item);
                }
            }
            /**
            * render method will take item as a props and print in li
            * @returns {string} - return the list of items
            */
            render() {
                return (
                    <li key={this.props.item.id} onClick={this.handleItemClick}>{this.props.item.text}</li>
                );
            }
        }
```


react/README.md
```diff
@@ -456,12 +456,39 @@
           {props.items.map((item, index) => (
             <Item
               key={item.key}
-              onClick={() => doSomethingWith(item.name, index)}
+              onClick={() => doSomethingWith(item.name, index)}
```
ljharb: please remove this trailing space

react/README.md
```diff
@@ -456,12 +456,39 @@
...
       );
     }
+    // We shouldnot use arrow function onClick as it create instance of method doSomethingWith every time in for loop in your example. //Instead we can function call with ==> and access the attribute or pass on as a props to Item React componenet. Let me know if have any other question
+    function ItemList(props) {
+      return (
+        <ul>
+          {props.items.map((item, index) => (
+            <Item
+              key={item.key}
+              onItemClick={this.doSomethingWith}
```
ljharb: this is an SFC (stateless functional component) - it has no `this`. What is `this.doSomethingWith` here?

react/README.md
```diff
@@ -456,12 +456,39 @@
...
+            <Item
+              key={item.key}
+              onItemClick={this.doSomethingWith} 
+            />
+          ))}
+        </ul>
+      );
+    }
+    
+   handleItemClick = (item) => {
```
ljharb: a) this is an SFC, so you can't use class properties in it.
b) class properties are a stage 2 proposal, and this guide forbids using syntax that is not stage 3 or higher.

react/README.md
```diff
@@ -456,12 +456,39 @@
...
             <Item
               key={item.key}
-              onClick={() => doSomethingWith(item.name, index)}
+              onClick={() => doSomethingWith(item.name, index)} 
             />
           ))}
         </ul>
       );
     }
+    // We shouldnot use arrow function onClick as it create instance of method doSomethingWith every time in for loop in your example. //Instead we can function call with ==> and access the attribute or pass on as a props to Item React componenet. Let me know if have any other question
```
ljharb: shouldnot → should not
componenet → component

ljharb (changes requested):
(i edited your comment to add backticks, so the code displayed properly)

For item 1, you can only do that as-is if you're using the obsolete and deprecated `React.createClass` - otherwise you need a constructor-bound instance method (ie, `this.handleClick = this.handleClick.bind(this);` in the constructor)

I'm not really clear what this PR is intending to improve - could you elaborate?

Untitled Diagram.xml
```diff
@@ -0,0 +1 @@
...uLAzJdI/BwoCcoBOkIXKuS0OCYQIe2m6pBx/iIJC2E+ENWsLxI95Bl5bkIXRp2QgH62CP82Ba51uD7/WCLOfB6sg6Q82O9nghgD/TC6bbNvqGcv+EyVOl+eL+bAlWd/hRDyfcHxfZVBw1LgK7OV1gFNi4TBwZ0YQ+egCMGBgFmD8538gIldEzfijrHSx45g6NjUhHBuOHW7dlPzIaaJI+YDZoZaceGR29a4xszKiRUWBnFKKER89VDlBHSDOdcwhfGSw157vWW62Oj7PCfdSgKbDD/GncVZdZcWRciCcT+Kt28N4ETJf/bjkIz42J4wZOpxK/u5iIrOsxcbIIvvEnW6SQf1ep5lssXuDnWv/k2UZB37Uc+u00WeAWicyetjAvF0ujwe7bCpwW9EAgxPNCy4D41VhIm5cud/hQfnUw2HaY+gYFrs/7dLSxU77S6vgKF+9VqL81439b4tU/yH+fe4PYONTRu8D7PQezDUc8x6BNbEvjINF8BaGOKDSXX08vJXXUehrhQ1Mo7lhXisXGZpWh7fASLY3/MGI+JFI8ofNu8D73f94NKNVSGiIVVYqmhoneALg6eHYVErIWuvoS/gDt+05FBEYSK5hfgzk2Ty8yXr20n0G9D0F39oValz1xxQ1HpiNMtZnphPdNiZm9DExvxJ8poZE6Wo5FuWEqSjphcxOtNJKdLJ6r8qPCvOd/PX5pMJ4igAlttzwqRLXWDuYlvPXkKvWSActllUF15ZhL9Kqv/RspOuYoYUp0f6804QW3IYWg1rDksmPhD5uS2rp3TUnMyqS60MAdfSo66mek9PFPRXpT7xBsThbtBc7Y3I3cL9fmiz7XwlvHceiI5TychoshtdmjT0OAonir4iy6qMKsKugapRzpoCqdCD2qImFOph5pVmkohD+2HGmGgqHZkbdQ0LCSiIOvUNDe0E09klQ8UT1ySYhOUY8M7Qh5U8iIl6Aq27k5MReGdDoYaiMAXdDJPIoHM+Sd2NN8JePUbNuF4f0vzBc1hxq3i8Asy5mlUU/W7YJmSHvcc6IrHFJp6pBriJ/nmT290xKtzukyan5LiYRjOWXaGLsnj7GpXTYb8DaJTht8zbskRxtkMxOwJH2O8DZJeCo1MUSt52I6XvogZU+QI7RTINMgRf+oeefdwNGqh3DZ/k+p3r/tP7/C+18=</diagram></mxfile>
```
ljharb: why is there an xml file in this PR?

ljharb (changes requested):

react/README.md
reetesha: Done. Please check

react/README.md
reetesha: Done. Please check

navyjax2:
Linting error occurs if I do 
`render() {
                return (
                    <li key={this.props.item.id} onClick={this.handleItemClick}>{this.props.item.text}</li>
                );
            }`
It will say:  "Visible, non-interactive elements should not have mouse or keyboard event listeners jsx-a11y/no-static-element-interactions"  The function will still work, but seems like the linting rules are saying there's supposed to be another way to do this?

ljharb:
That’s unrelated to the topic of this PR; but yes, it’s saying you shouldn’t put an onClick on an `<li>`

navyjax2:
Right, well, more than just on an <li>, but on any DOM object, I've found.  How is it unrelated?  Seems all of the code samples here all have it - you're using onClick on an element here, aren't you?  The only way I found out of the mess was to use `<Link to={"\MyPage\" + item.ID}>item.ID</Link>` - and set the routing to pick up the ID.  I had no clue how to avoid using an onClick on a DOM element, when that element needs a click to make a handler do its thing!  But because we're using this linting that apparently Facebook and UPS and others also use, seems like it would be applicable to the conversation - because you guys are saying "put this onClick in here", but the linting rules we're following at my job are saying not to.  Was hoping to see what the professionals like you guys thought the rule was trying to say to do/use, instead - because all the code samples I've seen so far stick an onClick in the element, and if we're trying to improve the framework, it seems like we should start by following the rules, shouldn't we?

ljharb:
`<button>` and `<a>` can certainly have an onClick - the purpose of that linting rule is to limit click behavior to things that are inherently clickable.

navyjax2:
Thanks for that clarification.  Seems they're rather strict on what can have an onClick, then, but at least now I know.

PR #1348, opened 2017-03-24
Fix dot-notation rule with non id props

aimicheng:
Fix dot-notation rule with non id props:
```javascript
const data = {};
data['foo-bar'] = 1;
```


ljharb:
Thanks for the contribution! I'm not sure I understand the reason for this PR. The `dot-notation` rule already allows properties to be quoted when they are invalid identifiers; and the point of the rule is to forbid quotes when the property name is a valid identifier, no matter its purpose.

Can you explain the motivation here?

PR #1361, opened 2017-04-06
.eslintrc deprecated

leftynaut:
`.eslintrc` is deprecated in favor of `.eslintrc.js`

http://eslint.org/docs/user-guide/configuring#configuration-file-formats

### DIFF
--- a/README.md
+++ b/README.md
@@ -3135,7 +3135,7 @@ Other Style Guides
 **Tools**
 
   - Code Style Linters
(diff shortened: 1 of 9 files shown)


ljharb:
Thanks for the contribution!

I'm aware, but `.eslintrc` will likely never be removed, and I believe it's better to keep the long-standard rc name than to pick an extension. I prefer to continue recommending `.eslintrc`.

PR #1372, opened 2017-04-13
Fix error on function within 8.2

Willibaur:
The function in place as `bad` example is
```javascript
[1, 2, 3].map(number => {
});
```
But the one proposed as `good` example missed the `+ 1` operator
```javascript
[1, 2, 3].map(number => `A ${number}`);
```

### DIFF
--- a/README.md
+++ b/README.md
@@ -927,7 +927,7 @@ Other Style Guides
     });
 
     // good
-    [1, 2, 3].map(number => `A string containing the ${number}.`);
+    [1, 2, 3].map(number => `A string containing the ${number + 1}.`);
 
     // good
     [1, 2, 3].map((number) => {


README.md
```diff
@@ -927,7 +927,7 @@ Other Style Guides
     });
 
     // good
-    [1, 2, 3].map(number => `A string containing the ${number}.`);
+    [1, 2, 3].map(number => `A string containing the ${number + 1}.`);
```
ljharb: I'm not sure this is an error; the examples don't all have to do identical things.

Willibaur: Oh I see, I will close this PR then @ljharb thanks

ljharb: Nobody's filed an issue with that confusion before.

PR #1376, opened 2017-04-19
Use dangling comma in the diff example for dangling commas

ashusta:
Seemed inconsistent to say "use dangling commas" but then not include dangling commas in the example diff.


tleunen:
It's only for multiline: https://github.com/airbnb/javascript/blob/master/packages/eslint-config-airbnb-base/rules/errors.js#L4-L10

PR #1381, opened 2017-04-21
Fix React code examples; ref cb and semicolons

ndresx:
- Added missing semicolons
- ~~Simplified `ref` callback example by removing parentheses and curly braces~~
- ~~Removed wrapping parentheses from `render` `return` element~~

### DIFF
--- a/react/README.md
+++ b/react/README.md
@@ -476,7 +476,7 @@
       }
 
       render() {
-        return <div onClick={this.onClickDiv.bind(this)} />
+        return <div onClick={this.onClickDiv.bind(this)} />;
       }
     }
 
@@ -493,7 +493,7 @@
       }
 
       render() {
-        return <div onClick={this.onClickDiv} />
+        return <div onClick={this.onClickDiv} />;
       }
     }
     ```
@@ -575,7 +575,7 @@
       }
 
       render() {
-        return <a href={this.props.url} data-id={this.props.id}>{this.props.text}</a>
+        return <a href={this.props.url} data-id={this.props.id}>{this.props.text}</a>;
       }
     }
 


react/README.md
```diff
@@ -386,7 +386,7 @@
 
     // good
     <Foo
-      ref={(ref) => { this.myRef = ref; }}
+      ref={ref => this.myRef = ref}
```
ljharb: this is invalid; implicit return arrow functions should not contain assignments; so this must be an explicit return; one-arg explicit return arrow functions must have parens around the arg. In other words, it's perfect as-is.

react/README.md
```diff
@@ -531,7 +531,7 @@
 
     // good
     render() {
-      return (<div />);
```
ljharb: single-line jsx are absolutely allowed to have parens - and i personally encourage it always - the guide only requires it around multiline parens. Please revert.

react/README.md
ndresx: According to the official React documentation (https://facebook.github.io/react/docs/refs-and-the-dom.html) it's a totally valid approach:

> There is even a shorter way to write it: ref={input => this.textInput = input}.

But ofc., if you want me to change it back I can do so!

react/README.md
ndresx: Reverting it is not a problem, _why I removed it_? I think it would be more consistent throughout all the examples to have it without the parentheses (there is not a single one which contains them) and I would expect that the Parentheses section (https://github.com/airbnb/javascript/tree/master/react#parentheses) is the guide's recommendation for the single line return - which is without parentheses.

Moreover, this particular section the change was made in is only referring to the return statement in render functions, so introducing another variation of a single line return could, under circumstances, lead to unnecessary confusion.

If you want me to change it back, just let me know please.

react/README.md
ljharb: It's valid in JavaScript. It's not valid according to this styleguide (read the main guide).

react/README.md
ljharb: Please do; our rules only don't require parens for single-line jsx because there isn't internal consensus at Airbnb on requiring them nor forbidding them - thus, it's important that this guide *be* inconsistent, to illustrate that either is perfectly acceptable.

ndresx:
@ljharb Have reverted both changes. I get your point with the arrow function, the question around parentheses will remain for me, but that's ok!

ljharb:
Thanks, this is great! Just needs a rebase down to 1 commit.

ndresx:
Should be ok now.

PR #1387, opened 2017-04-25
Remove "simply" from Readme

jondubin:
Our usage of "simply" in the Readme was called out on Twitter today https://twitter.com/iamsapegin/status/856880857570832384?s=09. Best to avoid its use since what is "simple" is relative.


jondubin:
And thanks @jongold for the spot!

jongold:
image

PR #1398, opened 2017-05-01
Fix react/prefer-stateless-function "ignorePureComponents" option typo

VladShcherbin:


rixth:
Whoops! Thanks @VladShcherbin

PR #1403, opened 2017-05-05
also disallow padding in classes and switches

graingert:


graingert:
I recently discovered that eslint have made the change to the rule default too

graingert:
https://github.com/eslint/eslint/issues/7879 might be worth just waiting until they release that

ljharb:
True - but actually shipping this change now will help make the change to eslint v4 not be breaking - when v4 comes out, it'd be ideal to support both v3 and v4 in a semver-minor release.

PR #1411, opened 2017-05-11
Disallow unused global variables

doshisid:
The current configuration allows unused global variables for unknown reason as stated [here](https://github.com/facebookincubator/create-react-app/issues/1719#issuecomment-286654527)

As stated [here](https://github.com/facebookincubator/create-react-app/issues/1719#issuecomment-285043565) the current config allows this behaviour because it was default in eslint.

But since eslint has changed the defaults to warn on any unused variable local/global, we should update this config to use new defaults.

### DIFF
--- a/packages/eslint-config-airbnb-base/rules/variables.js
+++ b/packages/eslint-config-airbnb-base/rules/variables.js
@@ -34,7 +34,7 @@ module.exports = {
     'no-undefined': 'off',
 
     // disallow declaration of variables that are not used in the code
-    'no-unused-vars': ['error', { vars: 'local', args: 'after-used', ignoreRestSiblings: true }],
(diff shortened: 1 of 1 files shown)


packages/eslint-config-airbnb-base/rules/variables.js
```diff
@@ -34,7 +34,7 @@ module.exports = {
     'no-undefined': 'off',
 
     // disallow declaration of variables that are not used in the code
-    'no-unused-vars': ['error', { vars: 'local', args: 'after-used', ignoreRestSiblings: true }],
```
ljharb: instead of removing "vars", we should set it to "all" explicitly.

ljharb:
Thanks, released in v11.2.0

PR #1420, opened 2017-05-20
Blacklist confusing globals

doshisid:
Blacklist confusing global variables set by browser env as discussed [here](https://github.com/facebookincubator/create-react-app/pull/2130#issuecomment-302808156).


doshisid:
We have changed the name to `confusing-browser-globals` at CRA. see [here](https://github.com/facebookincubator/create-react-app/pull/2286#discussion_r117611228)
Will send an update once that PR gets merged. So both has same data source as you wanted.

ljharb:
Sounds good, thanks!

PR #1424, opened 2017-05-24
[6.5] Add no-eval eslint rule to docs

daxelrod:
Add reference to the eslint `no-eval` rule to the README.

This rule is already set to `error` in best-practices.js in eslint-airbnb-config-base.


ljharb:
Thanks!

PR #1427, opened 2017-05-24
[eslint config] [eslint config base] Additional test extensions.

knpwrs:
I did some searching and didn't see any discussion about this. I'd like to have `*.test.jsx` files in my project so that I can have tests with JSX that don't interfere with `react/jsx-filename-extension`.

### DIFF
--- a/packages/eslint-config-airbnb-base/rules/imports.js
+++ b/packages/eslint-config-airbnb-base/rules/imports.js
@@ -73,10 +73,9 @@ module.exports = {
         'tests/**', // also common npm pattern
         'spec/**', // mocha, rspec-like pattern
         '**/__tests__/**', // jest pattern
-        'test.js', // repos with a single test file
-        'test-*.js', // repos with multiple top-level test files
-        '**/*.test.js', // tests where the extension denotes that it is a test
-        '**/*.spec.js', // tests where the extension denotes that it is a test
+        'test.{js,jsx}', // repos with a single test file
+        'test-*.{js,jsx}', // repos with multiple top-level test files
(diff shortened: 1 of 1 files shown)


knpwrs:
Alternatively this can be globbed with `{js,jsx}`.

knpwrs:
I searched and yet didn't see the PR _right before mine_. Updated in any case.

packages/eslint-config-airbnb-base/rules/imports.js
```diff
@@ -73,10 +73,10 @@ module.exports = {
...
         'spec/**', // mocha, rspec-like pattern
         '**/__tests__/**', // jest pattern
-        'test.js', // repos with a single test file
-        'test-*.js', // repos with multiple top-level test files
-        '**/*.test.js', // tests where the extension denotes that it is a test
-        '**/*.spec.js', // tests where the extension denotes that it is a test
+        'test.{js,jsx}', // repos with a single test file
+        'test-*.{js,jsx}', // repos with multiple top-level test files
+        '**/*.test.{js,jsx}', // tests where the extension denotes that it is a test
+        '**/*.spec.{js,jsx}', // tests where the extension denotes that it is a test
```
ljharb: is it worth globbing together test/spec as well?

knpwrs: Like `**/*.{test,spec}.{js,jsx}`?

ljharb: yep, just like that!

knpwrs: Just checked to make sure it would work with `minimatch`. It seems like we should be good. I'll push an update.

PR #1431, opened 2017-05-26
Adding Trailing Comma

mcmar:
If you're gonna make a styleguide, you should really lint your own guide.


packages/eslint-config-airbnb-base/rules/style.js
```diff
@@ -108,7 +108,7 @@ module.exports = {
       overrides: {
         return: { after: true },
         throw: { after: true },
-        case: { after: true }
+        case: { after: true },
       }
```
ljharb: you're missing one here too.

PR #1434, opened 2017-05-31
[7.7] Modify code comment error

getjll:
I think it's a mistake to write false in falsy.


fvgs:
Thanks for taking the time to submit this. The use of `falsy` is actually correct here. See [falsy](https://developer.mozilla.org/en-US/docs/Glossary/Falsy) for more information.

getjll:
Thank you for your answer.

PR #1436, opened 2017-06-02
Add linting for Markdown prose

daxelrod:
I hear you like linting! Why not enforce consistent style for your Markdown prose too?

Like JavaScript linting, Markdown linting can help identify errors or constructs that not all Markdown renderers handle correctly. The linter caught a few bugs in the README.

If you disagree with any of how I've gone about this, please let me know, and I'm happy to change things.

If you disagree with the intent of this completely, then oh well, it was a fun exercise. :smile: 

## Lint Rules

My intent is to codify the existing style being used, rather than to try to change how Markdown is written in this repo.

1. I've set a catchall to enable all of the [Markdownlint Rules](https://github.com/DavidAnson/markdownlint/blob/master/doc/Rules.md) and subsequently disabled ones that don't apply. If you'd instead like an explicit list of rules that do apply (to avoid updates to markdownlint silently adding new rules), I can do that instead.
1. I disabled a rule when it appeared that one of the documents deliberately did not conform to it. (Even if this nonconformance only happened once, I disabled the rule globally rather than for that one occurrence in the document).
1. I changed document content to fix rule nonconformance when it appeared that:
   * There was a legitimate bug in the document (meaning either GitHub or another Markdown renderer would not render something correctly)
   * A rule was followed consistently, except for a few isolated cases

## Implementation

A new `lint` npm script has been added to the top-level `package.json`, and set as a `pretest` script, to match how `lint` scripts in subpackages are handled. The top-level `travis` script now also explicitly invokes `lint`.

There is now a devDependency on `markdownlint-cli` but not `markdownlint`. The [JavaScript implementation of Markdownlint](https://www.npmjs.com/package/markdownlint) is a library only, and unfortunately the CLI wrapper `markdownlint-cli` has a dependency, rather than a peerDependency, on `markdownlint`.

Thanks!

### DIFF
--- a/linters/.markdownlint.json
+++ b/linters/.markdownlint.json
@@ -1,0 +1,154 @@
+{
+  "comment": "Be explicit by listing every available rule. https://github.com/DavidAnson/markdownlint/blob/master/doc/Rules.md",
+  "comment": "Note that there will be numeric gaps, not every MD number is implemented in markdownlint.",
+
+  "comment": "MD001: Header levels should only increment by one level at a time",
+  "header-increment": true,
+
+  "comment": "MD002: First header should be a top level header",
+  "first-header-h1": true,
+
+  "comment": "MD003: Header style: start with hashes",
+  "header-style": {
+    "style": "atx"
+  },
+
+  "comment": "MD004: Unordered list style",
+  "ul-style": {
+    "style": "dash"
+  },
+
+  "comment": "MD005: Consistent indentation for list items at the same level",
+  "list-indent": true,
+
+  "comment": "MD006: Consider starting bulleted lists at the beginning of the line",
+  "ul-start-left": false,
+
+  "comment": "MD007: Unordered list indentation: 2 spaces",
+  "ul-indent": {
+    "indent": 2
+  },
+
+  "comment": "MD009: Disallow trailing spaces",
+  "no-trailing-spaces": {
+    "br-spaces": 0,
+    "comment": "Empty lines inside list items should not be indented",
+    "list_item_empty_lines": false
+  },
+
+  "comment": "MD010: No hard tabs, not even in code blocks",
+  "no-hard-tabs": {
+    "code_blocks": true
+  },
+
+  "comment": "MD011: Prevent reversed link syntax",
+  "no-reversed-links": true,
+
+  "comment": "MD012: Disallow multiple consecutive blank lines",
+  "no-multiple-blanks": {
+    "maximum": 1
+   },
+
+  "comment": "MD013: Line length",
+  "line-length": false,
+
+  "comment": "MD014: Disallow dollar signs used before commands without showing output",
+  "commands-show-output": true,
+
+  "comment": "MD018: Disallow space after hash on atx style header",
+  "no-missing-space-atx": true,
+
+  "comment": "MD019: Dissalow multiple spaces after hash on atx style header",
+  "no-multiple-space-atx": true,
+
+  "comment": "MD020: No space inside hashes on closed atx style header",
+  "no-missing-space-closed-atx": true,
+
+  "comment": "MD021: Disallow multiple spaces inside hashes on closed atx style header",
+  "no-multiple-space-closed-atx": true,
+
+  "comment": "MD022: Headers should be surrounded by blank lines",
+  "comment": "Some headers have preceeding HTML anchors. Unfortunate that we have to disable this, as it otherwise catches a real problem that trips up some Markdown renderers",
+  "blanks-around-headers": false,
+
+  "comment": "MD023: Headers must start at the beginning of the line",
+  "header-start-left": true,
+
+  "comment": "MD024: Disallow multiple headers with the same content",
+  "no-duplicate-header": true,
+
+  "comment": "MD025: Disallow multiple top level headers in the same document",
+  "comment": "Gotta have a matching closing brace at the end",
+  "single-h1": false,
+
+  "comment": "MD026: Disallow trailing punctuation in header",
+  "comment": "Gotta have a semicolon after the ending closing brace",
+  "no-trailing-punctuation": {
+    "punctuation" : ".,:!?"
+  },
+  "comment": "MD027: Dissalow multiple spaces after blockquote symbol",
+  "no-multiple-space-blockquote": true,
+
+  "comment": "MD028: Blank line inside blockquote",
+  "comment": "Some 'Why?' and 'Why not?' blocks are separated by a blank line",
+  "no-blanks-blockquote": false,
+
+  "comment": "MD029: Ordered list item prefix",
+  "ol-prefix": {
+    "style": "one"
+  },
+
+  "comment": "MD030: Spaces after list markers",
+  "list-marker-space": {
+    "ul_single": 1,
+    "ol_single": 1,
+    "ul_multi": 1,
+    "ol_multi": 1
+  },
+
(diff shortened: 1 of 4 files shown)


lencioni:
@daxelrod Thanks for the PR! The markdown changes make sense to me and in general I think it is a good idea to lint our markdown. I don't have any experience with this specific tool, but I'm definitely on board for adding this or something like it to this project.

It seems like this is the sort of thing we might want to mention in a CONTRIBUTING.md document (which I'm actually a little surprised we don't already have).

daxelrod:
I'm glad you like the idea!

> I don't have any experience with this specific tool

Here's the markdown linter space as far as I'm aware:

The granddaddy of Markdown linters is [markdownlint/markdownlint](https://github.com/markdownlint/markdownlint), implemented in Ruby. It has a [nice list of rules](https://github.com/markdownlint/markdownlint/blob/master/docs/RULES.md) based upon the [CommonMark](http://commonmark.org/) spec and knowledge of pecularities about various parsers.

It was subsequently ported to JavaScript as [DavidAnson/markdownlint](https://github.com/DavidAnson/markdownlint). (This is the package I used for this PR). I have some experience using this package in [my own open source project](https://github.com/daxelrod/jowl) as well as in Visual Studio Code via [its markdownlint extension](https://marketplace.visualstudio.com/items?itemName=DavidAnson.vscode-markdownlint) by the same author.

In answering your question, I also just discovered [wooorm/remark-lint](https://github.com/wooorm/remark-lint). It uses [its own set of rules](https://github.com/wooorm/remark-lint/blob/master/doc/rules.md) but in practice there's [a good deal of overlap with markdownlint rules](https://github.com/wooorm/remark-lint/blob/3b11ff4a34313e33b50e8dc401d068ba5ebff424/doc/comparison-to-markdownlint.md). I have no personal experience with it, but it does look more pluggable, and there are a [bunch of nonofficial rules others have written](https://github.com/wooorm/remark-lint#list-of-external-rules) as well.

> It seems like this is the sort of thing we might want to mention in a `CONTRIBUTING.md` document (which I'm actually a little surprised we don't already have).

Me too! I looked for one when working on this.

Since I'm not a maintainer of this project, I don't think I'm the right person to write up an initial `CONTRIBUTING.md`, but I'll happily add information about Markdown linting to one.

linters/.markdownlint.json
```diff
@@ -0,0 +1,33 @@
+{
+  "comment": "Enable all rules, and then disable selective ones https://github.com/DavidAnson/markdownlint/blob/master/doc/Rules.md",
```
ljharb: i would prefer (just like our eslint config) to not "enable all", but to always explicitly define every single possible rule in our config.

linters/.markdownlint.json
```diff
@@ -0,0 +1,33 @@
...
+  "default": true,
+
+  "line-length": false,
+  "ul-start-left": false,
+  "no-multiple-blanks": false,
```
ljharb: this seems worth enabling?

linters/.markdownlint.json
```diff
@@ -0,0 +1,33 @@
...
+  "no-emphasis-as-header": false,
```
ljharb: what are the violations from this one being enabled?

linters/.markdownlint.json
```diff
@@ -0,0 +1,33 @@
...
+
+  "comment": "Needed for explicit anchors",
+  "no-inline-html": false,
+
+  "comment": "Rules use '-', some other lists use '+'",
```
ljharb: i'd probably be fine using `-` consistently

linters/.markdownlint.json
```diff
@@ -0,0 +1,33 @@
...
+  "ul-style": false,
+
+  "comment": "LICENCE.md does not conform",
```
ljharb: in fact i think we shouldn't be linting the license file at all, since that's legalese, not prose. can we exempt it?

hshoff:
A bit surprised that our Readme's markdown was only failing to pass the lint in two places after so many changes and contributions over the years. Not bad. 👍

linters/.markdownlint.json
daxelrod: Great point.

It turns out that `markdownlint-cli` wasn't handling recursive wildcards the way I expected, and I only have shell wildcards at my disposal. I'm going to restrict this PR to just linting style guides for now, rather than all of the `.md` files in the repo, and that will take care of not linting `LICENSE.md`.

I'm curious, though, why does the `LICENSE` file have a `.md` extension in the first place? It doesn't have any markup in it.

ljharb: That tells github how to display it; it's valid both with and without the extension.

ljharb: (ftr, to pass the glob to the tool, you'd need to quote it, otherwise the shell will try to expand it eagerly - and bash doesn't understand `**`)

linters/.markdownlint.json
daxelrod: This rule triggers whenever there is as paragraph that consists only of emphasized text.

Some examples are the document subhead

```
*A mostly reasonable approach to JavaScript*
```

and the sections under `##Resources` like

```
**Read This**

  - [Standard ECMA-262](http://www.ecma-international.org/ecma-262/6.0/index.html)
```

(which seem like they probably should either be list items or headings).

Here's the complete list.

```
README.md: 3: MD036 Emphasis used instead of a header
README.md: 3158: MD036 Emphasis used instead of a header
README.md: 3165: MD036 Emphasis used instead of a header
README.md: 3169: MD036 Emphasis used instead of a header
README.md: 3177: MD036 Emphasis used instead of a header
README.md: 3183: MD036 Emphasis used instead of a header
README.md: 3190: MD036 Emphasis used instead of a header
README.md: 3198: MD036 Emphasis used instead of a header
README.md: 3217: MD036 Emphasis used instead of a header
README.md: 3230: MD036 Emphasis used instead of a header
react/README.md: 3: MD036 Emphasis used instead of a header
```

daxelrod:
Thank you for the detailed feedback, @ljharb ! I've implemented all of your suggestions. Please see my [(now hidden) comment about `no-emphasis-as-header`](https://github.com/airbnb/javascript/pull/1436#discussion_r119975888) and let me know if you'd like me to turn that rule on as well and fix the document to comply with it.

Please let me know if you like my new approach of listing every markdownlint rule, and setting it either to its default, or changing it when necessary to comply with the document's style. I chose this over just specifying the applicable rules because being able to see which rules are disabled sparked a lot of good comments like yours above in the vein of "we *should* actually follow this".

@hshoff my hypothesis is that people contributing to a style guide are more vigilant than average about style :smile: 

I would also like to thank everyone involved. It's extremely rewarding to have three maintainers look at my PR within hours of me submitting it. It's also extremely impressive given the amount of activity that this project sees.

package.json
```diff
@@ -6,10 +6,12 @@
     "preinstall": "npm run install:config && npm run install:config:base",
     "install:config": "cd packages/eslint-config-airbnb && npm prune && npm install",
     "install:config:base": "cd packages/eslint-config-airbnb-base && npm prune && npm install",
+    "lint": "markdownlint --config linters/.markdownlint.json README.md react/README.md css-in-javascript/README.md",
```
ljharb: `README.md */README.md` might be ok too?

ljharb:
(This will need a rebase on the commandline prior to merging)

PR #1438, opened 2017-06-04
[guide] Add return in .map example

JPeer264:
I guess a `return` was missing in the examples of the arrow functions.


README.md
```diff
@@ -923,7 +923,7 @@ Other Style Guides
     // bad
```
ljharb: This is very intentional; the missing return is what makes it bad.

PR #1449, opened 2017-06-12
No arrow function implicit return with side effects

### LINKED ISSUE #1441: Distinguish between expressions with and without side-effects when using arrow function implicit return
Rule [8.2](https://github.com/airbnb/javascript#arrows--implicit-return) does not distinguish between [expressions](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Guide/Expressions_and_Operators#Expressions) with side-effects and expressions that only evaluate to a value. This can lead to subtle issues if an expression with a side-effect, such as an assignment, returns a value via the implicit return without the programmer's intention.

```javascript
function foo(callback) {
  const val = callback();
  if (val === true) {
    // do something if callback returns true
  }
}

let x = false;
// bad
foo(() => x = true);
```

In this example, the programmer may have intended for the callback function to only assign the value `true` to the variable `x`. However, due to the implicit return and the lack of curly braces, the arrow function not only performs the assignment, but also returns the value `true`. `foo` interprets this return value in a meaningful way and performs some action the programmer did not intend.

Rule 8.2 does not address this case where using the implicit return with an expression having a side-effect can lead to unintended behavior.

Unless the programmer intends for the value `true` to be returned from the callback, the above call to `foo` should be written as:

```javascript
// good
foo(() => {
  x = true;
});
```

The wording of [4.5](https://github.com/airbnb/javascript#arrays--callback-return) should also be revised.

[11.1](https://github.com/airbnb/javascript#iterators--nope) has a couple examples of expressions with side-effects being returned via implicit return. Though, in the case of a `forEach`, there is no risk of unintended behavior. Whether or not it is acceptable to make unnecessary use of the implicit return in such situations (e.g. `forEach`, `new Promise`) where the return value is ignored is a topic for discussion.

fvgs:
Resolves #1441

### DIFF
--- a/README.md
+++ b/README.md
@@ -369,7 +369,7 @@ Other Style Guides
     ```
 
   <a name="arrays--callback-return"></a><a name="4.5"></a>
-  - [4.5](#arrays--callback-return) Use return statements in array method callbacks. It’s ok to omit the return if the function body consists of a single statement following [8.2](#arrows--implicit-return). eslint: [`array-callback-return`](http://eslint.org/docs/rules/array-callback-return)
+  - [4.5](#arrays--callback-return) Use return statements in array method callbacks. It’s ok to omit the return if the function body consists of a single statement returning an expression without side effects, following [8.2](#arrows--implicit-return). eslint: [`array-callback-return`](http://eslint.org/docs/rules/array-callback-return)
 
     ```javascript
     // good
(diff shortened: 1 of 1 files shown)


ljharb:
Please also check "allow edits"; it allows me to normalize the commit message and rebase as needed so I can merge efficiently.

fvgs:
Gotcha, done

PR #1455, opened 2017-06-14
Fix iterator code example error

btpschroeder:
The first "good" example was missing a closing bracket (line 1377).


ljharb:
Thanks!

PR #1465, opened 2017-06-23
Use fetch over $.ajax

graingert:


oshalygin:
While good, hardly any browser supports this except everything thats greenfield.  IMO its clearer that it stay as `$.ajax` since most are at least familiar with jQuery AJAX calls, even if they don't use them.

graingert:
> hardly any browser


73% usage relative...

oshalygin:
> 73% usage relative...

That may be the case for browsers, but `fetch` is uncommon in most client code.  There are notable disadvantages to using `fetch` and I would wager that most developers have not used `fetch` in production.  

The guide is meant to be clear and instructive, as opposed to a forcing developers to research the fetch API.

https://medium.com/@thejasonfile/fetch-vs-axios-js-for-making-http-requests-2b261cdd3af5

ljharb:
`fetch`, unpolyfilled, is absolutely not broadly available enough.

However, `airbnb-browser-shims` polyfills it, and this guide strongly encourages you to always use it in browsers (and `airbnb-js-shims` in node).

Regardless, `$.ajax` is still a valid approach for those using jQuery, so while we may want to *also* add `fetch` examples, we would definitely not want to remove the jQuery examples.

ljharb (changes requested):
`fetch`, unpolyfilled, is absolutely not broadly available enough.

However, `airbnb-browser-shims` polyfills it, and this guide strongly encourages you to always use it in browsers (and `airbnb-js-shims` in node).

Regardless, `$.ajax` is still a valid approach for those using jQuery, so while we may want to *also* add `fetch` examples, we would definitely not want to remove the jQuery examples.

ljharb:
(separately, `fetch` isn't really meant to be used directly; it's a low-level primitive for libraries to abstract over)

PR #1471, opened 2017-06-28
Align JavaScript style guide with Newsela's style preferences.

sravanti:
This includes things like:

1. Not allowing spaces between curly braces
2. Newsela-specific import order
3. Using `_.size` and lodash methods in general over `.length` and inbuilt JavaScript methods.
4. A caveat for wildcard exports for Redux.


sravanti:
Yup, totally my bad!

Issue #1447, opened 2017-06-12
eslint v4 support


ljharb:
## Note: I filed this issue within minutes of eslint v4.0.0 being released.

`eslint-config-airbnb` and `eslint-config-airbnb-base` **do not yet support** eslint v4.

If it is easy/possible to support both v3 and v4, we will release a non-major bump of both packages that does so, before releasing a major bump that only supports v4.

This will likely take a few weeks at a minimum.

**Please** do not file additional issues and do not file pull requests attempting to add support - please, instead, continue to use eslint v3 until such time as our configs support v4.

Peer Dep eslint combined v3 and v4 support released:
 - [x] `eslint-plugin-import` [issue](https://github.com/benmosher/eslint-plugin-import/issues/870) | [PR](https://github.com/benmosher/eslint-plugin-import/pull/871) | released as v2.6.0
 - [x] `eslint-plugin-react` [issue for v7.1.0 release](https://github.com/yannickcr/eslint-plugin-react/issues/1251)
 - [x] `eslint-plugin-jsx-a11y` [issue](https://github.com/evcohen/eslint-plugin-jsx-a11y/issues/266) | [PR](https://github.com/evcohen/eslint-plugin-jsx-a11y/pull/267) | [release](https://github.com/evcohen/eslint-plugin-jsx-a11y/issues/269) (v5.1.0 and v6)
 - [x] `eslint-find-rules` [issue](https://github.com/sarbbottam/eslint-find-rules/issues/279) | released as v30.1
 - [ ] update and publish the base config
 - [ ] update and publish the main config

willin:
eslint must kidding...

ljharb:
@willin i'm not sure what your comment is saying nor what the point of it is, so i'm going to delete it.

ljharb:
@Geczy I've kept the OP updated; if there's updates, you'd see them here.

Since there are no updates posted, no, there are no updates. Please be patient - the OP said "a few weeks at minimum" and it's not yet been 3.

MFry:
It would be useful to post somewhere prominent on what would be recommended for new projects and the best way to stay on the 3x branch for eslint.

`yarn add eslint@3` for yarn users would go a long way to avoid setup hassles.

EloB:
Really would like this feature. The ability to use `overrides` would be awesome!

```
"overrides": [
  {
    "files": ["src/**/__tests__/*.js"],
    "env": {
      "jest": true
    }
  }
]
```

[...]

PR #1339, opened 2017-03-16
[guide] Add array-bracket-newline

JPeer264:
As discussed in #1338 I created a pull request. The naming, description and examples tend to the eslint rule `array-bracket-newline`, which hopefully comes soon (https://github.com/eslint/eslint/issues/6073):

```js
/*eslint array-bracket-newline: ["error", { "multiline": true }]*/
```

PR #1475, opened 2017-07-03
[eslint config][base][minor] add array-*-newline regarding to docs

JPeer264:
Follow PR of #1339. There are now 2 rules `array-bracket-newline` and `array-element-newline` which satisfies together the rule [4.6](https://github.com/airbnb/javascript#arrays--bracket-newline).

These rules are available since Eslint v4.0.0, so this PR should wait until Eslint v4 is supported by this styleguide (ref #1447).

### DIFF
--- a/README.md
+++ b/README.md
@@ -419,8 +419,8 @@ Other Style Guides
 
 **[⬆ back to top](#table-of-contents)**
 
-<a name="arrays--bracket-newline"></a>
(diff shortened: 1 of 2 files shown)


ljharb:
Thanks - but per #1447, PRs regarding eslint v4 are not needed nor desired. When adding support for v4, every new rule (of which these 2 are only a fraction) will be added.

PR #1484, opened 2017-07-06
fix minor error in  README.md

dong-jy:


css-in-javascript/README.md
```diff
@@ -19,7 +19,7 @@
     ```js
     // bad
     {
-      'bermuda-triangle': {
+      bermuda-triangle: {
```
ljharb: This change actually makes it incorrect; since in JS, `bermuda-triangle` is an invalid identifier, and thus it requires the quotes.

dong-jy: thanks

PR #1491, opened 2017-07-12
Typo fix on example

chytanya:


fvgs:
Thanks for spotting that!

PR #1503, opened 2017-07-25
[eslint config] Improve Gruntfile glob pattern

willclarktech:
To cover `Gruntfile`, `Gruntfile.js` and `Gruntfile.coffee` instead of just `Gruntfile`.

### DIFF
--- a/packages/eslint-config-airbnb-base/rules/imports.js
+++ b/packages/eslint-config-airbnb-base/rules/imports.js
@@ -82,7 +82,7 @@ module.exports = {
         '**/rollup.config.*.js', // rollup config
         '**/gulpfile.js', // gulp config
         '**/gulpfile.*.js', // gulp config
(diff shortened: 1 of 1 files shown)


packages/eslint-config-airbnb-base/rules/imports.js
```diff
@@ -82,7 +82,7 @@ module.exports = {
         '**/rollup.config.*.js', // rollup config
         '**/gulpfile.js', // gulp config
         '**/gulpfile.*.js', // gulp config
-        '**/Gruntfile', // grunt config
+        '**/Gruntfile{,.js,.coffee}', // grunt config
```
ljharb: I'm fine supporting the `.js` extension, but nobody should be using coffeescript. Can you remove `.coffee`?

willclarktech: Gladly!

PR #1507, opened 2017-07-29
Add no-param-reassign exception for AVA test contexts.

knpwrs:
See https://github.com/avajs/ava/tree/a0d5b373a19e9eb0022ad2936b46a1608f11b53f#test-context

Essentially `before` and `after` hooks in AVA are able to share data with tests through the `t.context` object:

```js
test.beforeEach(t => {
	t.context.data = generateUniqueData();
});

test(t => {
	t.is(t.context.data + 'bar', 'foobar');
});
```

### DIFF
--- a/packages/eslint-config-airbnb-base/rules/best-practices.js
+++ b/packages/eslint-config-airbnb-base/rules/best-practices.js
@@ -179,6 +179,7 @@ module.exports = {
         'res', // for Express responses
         'response', // for Express responses
         '$scope', // for Angular 1 scopes
+        't', // for AVA test contexts
       ]
     }],
 


ljharb:
Since for tests, it's easy to use a `.eslintrc` to add overrides; and since ava isn't nearly as popular as koa, express, and angular, I'm not sure it's worth adding this to airbnb's config (we use mocha, tape, and/or jest internally)

knpwrs:
Fair enough. Looks like [ESLint 4 added glob configuration](http://eslint.org/docs/user-guide/configuring#configuration-based-on-glob-patterns) which I haven't used yet. I'll give that a go.

Issue #1508, opened 2017-07-31
Yarn instructions for eslint-config-airbnb potentially installs incorrect version of peer dependencies


epan:
# Context
The Yarn instructions at [eslint-config-airbnb](https://www.npmjs.com/package/eslint-config-airbnb) readme currently installs the latest version of each peer dependency.

# Expected
Following the yarn instructions would install the correct peer dependencies.

# Actual
I encountered the error `Definition for rule 'jsx-a11y/href-no-hash' was not found` when running `yarn run eslint .` after following the Yarn instructions.

The Yarn instructions install the latest version of `eslint-plugin-jsx-a11y` which installs `eslint-plugin-jsx-a11y@6.0.2` but the peer dependency is `eslint-plugin-jsx-a11y@^5.1.1`

`eslint-config-airbnb` depends on a rule that no longer exists in `plugin-jsx-a11y@6.0.2`

# Workaround
1. Install package with `yarn add -D eslint-config-airbnb`
1. Run `npm info "eslint-config-airbnb@latest" peerDependencies` to get list of peer dependencies and their respective versions
1. `yarn add -D [dependency@version]` eg, `yarn add -D eslint-plugin-jsx-a11y@^5.1.1`
1. Repeat for each peer dependency listed in step 2

ljharb:
Airbnb doesn't use yarn. If you'd like to make a PR that improves the yarn instructions, that'd be great - if not, and they're incorrect, a PR to remove them would be fine too.

epan:
@ljharb I did my best - https://github.com/airbnb/javascript/pull/1511 Happy to revise as much as needed. Thanks for all this awesomeness around here.

PR #1511, opened 2017-07-31
Update yarn-specific install instructions

epan:
# Context
Attempts to clarify and address issue https://github.com/airbnb/javascript/issues/1508#issuecomment-319113807 to assist yarn users.

# What
New instructions recommend yarn users to list peer dependencies and then install each peer dependency with specified version.

# Why
Previous yarn instructions led users to install the latest version of each peer dependency which, in some cases, led to errors if the latest version of a peer dependency did not have rules that this package uses.


epan:
@ljharb good point! Updated the PR.

PR #1516, opened 2017-08-07
Remove polish translation

marhub:
Remove link to polish translation as it's outdated ( last commit from 2-3 years ago ). 
For example in polish translation you can read that you should use one "var" keyword to declare your vars.
example:
```js
// źle ( bad )
var items = getItems();
var goSportsTeam = true;
var dragonball = 'z';

// dobrze ( good )
var items = getItems(),
    goSportsTeam = true,
    dragonball = 'z';
```


mjurczyk:
@marhub @ljharb ah, I didn't think anybody used the polish translation, shame nobody put an issue there to update it. Do you think I should update the polish translation or is that maybe already underway?

ljharb:
@mjurczyk i don't think it's underway; if you're interested in replacing it that'd be great!

mjurczyk:
@ljharb sure, I will wait a few days to see if @marhub replies - I noticed he has a fork of the repo on his account and I wouldn't like to steal his progress or duplicate his efforts.
If he won't reply I will make sure to find some time to update the translation.

PR #1537, opened 2017-08-24
[guide] [react] HOC WrappedComponent static property

ifeature:
`WrappedComponent` static property should be added to custom High-Order component to get access to the initial(decorated component). In most cases it would allow to avoid issue [`import/no-named-as-default`](https://github.com/benmosher/eslint-plugin-import/blob/master/docs/rules/no-named-as-default.md)

### DIFF
--- a/react/README.md
+++ b/react/README.md
@@ -5,6 +5,7 @@
 ## Table of Contents
 
   1. [Basic Rules](#basic-rules)
+  1. [High-Order Components](#high-ordered-components)
   1. [Class vs `React.createClass` vs stateless](#class-vs-reactcreateclass-vs-stateless)
   1. [Mixins](#mixins)
   1. [Naming](#naming)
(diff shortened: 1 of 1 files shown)


react/README.md
```diff
@@ -26,6 +27,81 @@
     - However, multiple [Stateless, or Pure, Components](https://facebook.github.io/react/docs/reusable-components.html#stateless-functions) are allowed per file. eslint: [`react/no-multi-comp`](https://github.com/yannickcr/eslint-plugin-react/blob/master/docs/rules/no-multi-comp.md#ignorestateless).
   - Always use JSX syntax.
   - Do not use `React.createElement` unless you're initializing the app from a file that is not JSX.
+  
+## High-Order Components
+
+  - For [Higher-Order Components](https://facebook.github.io/react/docs/higher-order-components.html) define static property `WrappedComponent` as a reference to decorated component.
```
ljharb: This link doesn't suggest adding this static property. Where is this convention coming from?

react/README.md
```diff
@@ -26,6 +27,81 @@
     - However, multiple [Stateless, or Pure, Components](https://facebook.github.io/react/docs/reusable-components.html#stateless-functions) are allowed per file. eslint: [`react/no-multi-comp`](https://github.com/yannickcr/eslint-plugin-react/blob/master/docs/rules/no-multi-comp.md#ignorestateless).
   - Always use JSX syntax.
   - Do not use `React.createElement` unless you're initializing the app from a file that is not JSX.
+  
+## High-Order Components
+
+  - For [Higher-Order Components](https://facebook.github.io/react/docs/higher-order-components.html) define static property `WrappedComponent` as a reference to decorated component.   
+  
+    If your High-Order Components don't provide access to decorated components, you can get faced with issue of [`import/no-named-as-default`](https://github.com/benmosher/eslint-plugin-import/blob/master/docs/rules/no-named-as-default.md)
```
ljharb: This isn't really a problem due to enzyme's `.dive()` method - "wrapped" components no longer need to be directly exported for the same of tests.

react/README.md
```diff
@@ -26,6 +27,81 @@
...
+    }
+    export default bindBreakpoint(Title);
+    ```
+    
+    In our tests, we can get access to Pure, non-decorated `<Title />` component in this way:
+    
+    ```jsx
+    // Title.test.jsx
+    import Title from './Title';
+    shallow(<Title.WrappedComponent>Hello, World!</Title.WrappedComponent>
```
ljharb: specifically, `shallow(<Title />).dive()` is the proper way to do this.

PR #1482, opened 2017-07-05
Upgrade jsx-a11y to 6.0.2, enable more jsx-a11y rules

wyattdanger:
Upgrades eslint-plugin-jsx-a11y from 5 to 6, and enables more a11y related rules

PR #1546, opened 2017-09-01
Update "eslint-plugin-jsx-a11y": "^6.0.2"

inoas:
```bash
$ npm outdated
Package                 Current  Wanted  Latest  Location
eslint-plugin-jsx-a11y    5.1.1   5.1.1   6.0.2  React-Backend-Client
```

But, after updating package.json and running `npm install`:

    npm WARN eslint-config-airbnb@15.1.0 requires a peer of eslint-plugin-jsx-a11y@^5.1.1 but none was installed.

### DIFF
--- a/packages/eslint-config-airbnb/package.json
+++ b/packages/eslint-config-airbnb/package.json
@@ -57,7 +57,7 @@
     "eslint": "^3.19.0 || ^4.5.0",
     "eslint-find-rules": "^3.1.1",
     "eslint-plugin-import": "^2.7.0",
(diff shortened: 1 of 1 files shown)


inoas:
This may be related? @ Travis? https://github.com/evcohen/eslint-plugin-jsx-a11y/issues/283

ljharb:
We are currently not compatible with v6 of eslint-plugin-jsx-a11y.

If you use the install command in our readmes, you'll get the proper peer dep versions. `npm outdated` output is irrelevant.

Duplicate of #1482, which properly upgrades us to v6 as a semver-major change.

PR #1549, opened 2017-09-06
Update spread sections

bl00mber:


README.md
```diff
@@ -364,11 +364,16 @@ Other Style Guides
     ```
 
   <a name="arrays--from"></a><a name="4.4"></a>
-  - [4.4](#arrays--from) To convert an array-like object to an array, use [Array.from](https://developer.mozilla.org/en/docs/Web/JavaScript/Reference/Global_Objects/Array/from).
+  - [4.4](#arrays--from) To convert an array-like object to an array, use spreads `...` instead of [Array.from](https://developer.mozilla.org/en/docs/Web/JavaScript/Reference/Global_Objects/Array/from).
```
ljharb: this is good, but we should also include a "good" example that uses `Array.from` with a mapper function - specifically, `Array.from(foo, bar)` is much much better than `[...foo].map(bar)`.

README.md
```diff
@@ -364,11 +364,16 @@ Other Style Guides
...
 
     ```javascript
     const foo = document.querySelectorAll('.foo');
+
+    // bad
```
ljharb: this also isn't *bad*, it's just not ideal. I'd say make this "good" and the spread example "best".

README.md
```diff
@@ -420,7 +425,7 @@ Other Style Guides
     });
     ```
 
-<a name="arrays--bracket-newline"></a>
+  <a name="arrays--bracket-newline"></a>
```
ljharb: why this indentation change?

README.md
```diff
@@ -457,6 +462,17 @@ Other Style Guides
...
 
+  <a name="arrays--no-unnecessary-arrays"></a>
+  - [4.7](#arrays--no-unnecessary-arrays) Avoid create unnecessary arrays because of a performance reason.
+
+  ```javascript
+  // bad
+  const bar = [...foo].map(mapper);
+
+  // good
+  const bar = Array.prototype.map.call(foo, mapper);
```
ljharb: best is `Array.from(foo, mapper)` - `Array.prototype.map.call` isn't equivalent, since that only accepts arrays and arraylikes - spread and `Array.from` also accepts iterables.

README.md
```diff
@@ -2633,7 +2649,7 @@ Other Style Guides
       inventorOf,
     );
 
-    // good (note that a comma must not appear after a "rest" element)
+    // good (note that a comma must not appear after a "spread" element)
```
ljharb: this is incorrect; "rest" and "spread" are two different contextual meanings for the operator composed of three dots - and the comma is forbidding after *rest*, not after spread.

README.md
bl00mber: Because other similar parts of this guide have this indentation. Actually, this indentations don't affect the markdown preview on Github, so if needed I can remove this change.

README.md
```diff
@@ -364,15 +364,31 @@ Other Style Guides
...
+
+    // good
     const nodes = Array.from(foo);
+
+    // best
+    const nodes = [...foo];
+    ```
+
+  <a name="arrays--mapping"></a>
+  - [4.5](#arrays--mapping) Prefer [Array.from](https://developer.mozilla.org/en/docs/Web/JavaScript/Reference/Global_Objects/Array/from) over spread `...` for arrays mapping because of a performance reason.
```
ljharb: slight prose tweak: `for mapping over iterables; it avoids creating an intermediate array`

PR #1552, opened 2017-09-08
Removing functions from trailing comma rule

Jaybeetee:
As per https://github.com/airbnb/javascript/issues/951

This isn't standard syntax in JavaScript and results in 'Syntax Errors' in the event of a trailing comma in a function call.


ljharb:
Note, it won't cause syntax errors if you're using Babel (see the note st the top of the readme)

PR #1557, opened 2017-09-14
Allow multi space alignment (via no-multi-spaces)

epetrsn:
```javascript
// align end (requires additional rule change)
[  'foo',    'bar']
['apple', 'banana']

// align start (works as-is)
['foo',   'bar']
['apple', 'banana']
```


ljharb:
Seems like this is intended for your fork.

epetrsn:
Oops. Sorry about that.

PR #1563, opened 2017-09-20
Deleted duplicated example in ternaries.

cn007b:
In ternaries block present same code example: `const maybeNull = value1 > value2 ? 'baz' : null;` as `better` and `best`.
I've deleted this example from `better` example.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1914,16 +1914,15 @@ Other Style Guides
       ? "bar"
       : value1 > value2 ? "baz" : null;
 
-    // better
+    // split into 2 separated ternary expressions
     const maybeNull = value1 > value2 ? 'baz' : null;
 
+    // better
     const foo = maybe1 > maybe2
       ? 'bar'
       : maybeNull;
 
     // best
-    const maybeNull = value1 > value2 ? 'baz' : null;
-
     const foo = maybe1 > maybe2 ? 'bar' : maybeNull;
     ```
 


README.md
```diff
@@ -1899,8 +1899,6 @@ Other Style Guides
       : value1 > value2 ? "baz" : null;
 
     // better
-    const maybeNull = value1 > value2 ? 'baz' : null;
```
ljharb: this isn't actually a separate example, it's defining the code used a few lines down. Maybe it'd be better to just remove the empty lines between `const maybeNull` and `const foo`, in both sections?

cn007b: I'd understood the gist but was perplexed when I've found same code in better and good sections) Yes, it makes sense to delete empty lines, but still, same code will be in 2 sections(
Maybe it's possible to consider something like this:
````
// split into 2 separated ternary expression
const maybeNull = value1 > value2 ? 'baz' : null;

// better
const foo = maybe1 > maybe2
  ? 'bar'
  : maybeNull;

// best
const foo = maybe1 > maybe2 ? 'bar' : maybeNull;
````

ljharb: Sure, that's not bad. (`s/expression/expressions` tho)

ljharb:
Please don't open a second PR; let's reuse this one :-)

ljharb:
Happy to help if you're having issues rebasing!

cn007b:
OK. I've already closed 2nd PR) I don't use rebase in general...
So for me, it's just nightmare

ljharb:
np, i'll take care of it

cn007b:
I saw that you `surgeon with git rebase`)
I've squashed my commits but I was surprised with record `changed the title from`...
and decided that I'd done something wrong...

ljharb:
Next time it comes up, ping me in the gitter channel and I'll help walk you through it.

cn007b:
Ok, thank you so much! Sorry for this mess.

PR #1565, opened 2017-09-21
add rule

odykyi:
```
// bad 
if (foo === 123 || foo === 124 || foo === 125 || foo === 126) {
  thing1();
}

// good 
if ([123, 124, 125, 126].includes(foo)) {
  thing1();
}
```


aboyton:
Can we add a lint rule for this?

PR #1570, opened 2017-09-27
Fix broken example, severity of warning

thepeoplesbourgeois:
The problem with linting rules sometimes is that they can imply that not having a semicolon somewhere in thousands of lines of code (that runs perfectly fine without them) is equally as bad as accidentally writing a loop that crashes your application.

### DIFF
--- a/README.md
+++ b/README.md
@@ -381,10 +381,10 @@ Other Style Guides
 
     ```javascript
     // bad
-    const bar = [...foo].map(bar);
+    const baz = [...foo].map(bar);
 
     // good
-    const bar = Array.from(foo, bar);
+    const baz = Array.from(foo, bar);
     ```
 
   <a name="arrays--callback-return"></a><a name="4.5"></a>
@@ -400,18 +400,18 @@ Other Style Guides
     // good
     [1, 2, 3].map(x => x + 1);
 
-    // bad
+    // bad - no returned value means `memo` becomes undefined after the first iteration
     const flat = {};
     [[0, 1], [2, 3], [4, 5]].reduce((memo, item, index) => {
       const flatten = memo.concat(item);
-      flat[index] = flatten;
+      memo[index] = flatten;
     });
 
     // good
     const flat = {};
     [[0, 1], [2, 3], [4, 5]].reduce((memo, item, index) => {
       const flatten = memo.concat(item);
-      flat[index] = flatten;
+      memo[index] = flatten;
       return flatten;
     });
 


README.md
```diff
@@ -381,10 +381,10 @@ Other Style Guides
 
     ```javascript
     // bad
-    const bar = [...foo].map(bar);
```
ljharb: why do these need to change?

thepeoplesbourgeois: You're declaring `bar`. If you're declaring it here, how can it be what you pass in to `map`?

thepeoplesbourgeois: If this line of code works as it's written, I am more terrified of JavaScript than ever before

thepeoplesbourgeois:
And what misbehavior arises from a missing semicolon, exactly?

thepeoplesbourgeois:
I keep hearing nebulous fears from semicolon stalwarts that "bad things could happen", but I've yet to hear of any examples of such things

README.md
ljharb: aha, true :-)

ljharb:
I don't really have time to go into all the ways that ASI can and has caused numerous production bugs, but I think it's a bit presumptive to believe that something the majority of JS devs avoid isn't actually bad :-)

README.md
```diff
@@ -400,18 +400,18 @@ Other Style Guides
     // good
     [1, 2, 3].map(x => x + 1);
 
-    // bad
+    // very bad - error bad - no returned value means `memo` becomes undefined after the first iteration
```
ljharb: i think just "bad" here, with your english explanation, is sufficient

thepeoplesbourgeois:
I believe the logical fallacy you've just hit upon is "popularity confirmation bias", no?

I don't need "all the ways". If you can just tell the last time it's been a gotcha for you, I will concede the point.

README.md
thepeoplesbourgeois: That's fair. I've taken it back down just to "bad"

ljharb:
@thepeoplesbourgeois no, that would be if i claimed that it *was* true *because* lots of people do it. I'm claiming that someone who's new and/or uninformed should *assume by default* that if it's popular it's true, instead of assuming by default that it's some kind of unsubstantiated groupthink.

thepeoplesbourgeois:
Well... how new do you believe I am to Javascript? :)

ljharb:
lol, fair point, i apologize that i took your questions as implied newness.

thepeoplesbourgeois:
It happens 😁  I'll make the updates to the branch, but it'll be in a little while. I've been doing these edits in the online editor, so I need to find a moment to clone the repo and perform the terminal commands. Thank you for being my first approved open-source contrib!

ljharb:
oh no worries then; i'll take care of the rebase for you :-) Thanks for the contribution!

Issue #1567, opened 2017-09-23
why airbnb is using yet semicolons?


webmobiles:
I've read moderns apps don't use anymore semicolons, maybe in the future airbnb style  too ? maybe with https://standardjs.com/ ?

ljharb:
I'm not sure where you've read that, and it's quite false. Using semicolons is something the vast majority of JS devs do, despite the recent rise in usage of "standardJS" - something that is not actually a standard, nor something that aligns with what the majority of JS devs do in a number of ways.

The professional and modern thing to do is not to omit optional tokens, but to always use them, since explicit > implicit.

Bamieh:
Additionally, not using semi-colons can rise a series of unexpected bugs, take this line of code for example:

```
var obj = {n: 1};
var a = obj
[a].forEach(_ => console.log(_))
```

omidkrad:
I started a new project without using semicolons and after a few days I switched back to using semicolons for a simple reason: With semicolons, you get better parser, and in case of TypeScript, compiler messages.

Let's say you have forgotten a closing parenthesis on one line and went to the next line. The compiler may tell you that it expects a comma at the end of the line for an optional parameter, but if you closed the line with a semicolon, it will tell you that it expects a closing parenthesis. Also, I find it easier to skim through the code when lines are ended with semicolons.

mraak:
@omidkrad We have life critical production apps with 10'000's lines of code, I think it's around 100k now, all without semicolons. All these apps are heavily tested by test team and not a single bug could be traced back to the lack of semicolons. When a new developer joins and argues for the use of semicolons, I ask him to open a text editor and type 100k semicolons, each in new line, go to the CEO and explain to him what he achieved. Not a single one has done that and they all realized that typing useless stuff isn't productive.

ljharb:
@mraak punctuation isn't useless, and that kind of hostile comment, *again*, won't be tolerated here.

The absence of semicolon-related bugs in your codebase is anecdotal and proves nothing.

Luxcium:
> @omidkrad We have life critical production apps with 10'000's lines of code, I think it's around 100k now, all without semicolons. All these apps are heavily tested by test team and not a single bug could be traced back to the lack of semicolons. When a new developer joins and argues for the use of semicolons, I ask him to open a text editor and type 100k semicolons, each in new line, go to the CEO and explain to him what he achieved. Not a single one has done that and they all realized that typing useless stuff isn't productive.

[...]

Issue #394, opened 2015-07-03
Semicolons.  Sorta


laicosly:
The argument presented for using semicolons shows one of the few areas that semicolons are actually needed (anonymous self-enclosing functions, and for i loops).

Semicolons everywhere is clutter, use them only where required.

https://github.com/airbnb/javascript#20.1

justjake:
That's a perfectly reasonable stance on semicolons, and there are plenty of other styleguides out there that use that rule -- feross/standard comes to mind.

goatslacker:
As a fellow member of the no semicolons in my JS camp I couldn't agree more. But this is just the guide that we follow at Airbnb. I don't expect to see this change anytime soon.

goatslacker:
If you're interested in this guide sans semicolons then https://github.com/goatslacker/javascript

laicosly:
;very cool;
i just thought i'd suggest it;
thank you for the style-guide recommendations;
i am digging the feross/standard one so far;
carry on;

goatslacker:
Yeah two semicolons before and after each line is good practice, you know, for safety from the evil ASI monster that no one wants to bother to understand.

laicosly:
;; @goatslacker  :+1:  ;;

hshoff:
};

seb0zz:
Can anyone tell me why not using semicolons everywhere is a bad practice? I don't really find a satisfying argumentation anywhere in this repository.

ljharb:
@seb0zz that's a long argument, full of FUD on both sides, which really boils down to subjective preference. Personally I find the rules of ASI harder to remember and more likely to cause bugs, than the rules of semicolons - and ideologically I consider semicolons *so* required by the spec that it puts them in for you even if you forget :-) but it's really about what you and your team find consistently sensible.

lencioni:
@seb0zz Here's an example that I recently saw via Twitter: https://tonicdev.com/tolmasky/comparison

seb0zz:
@lencioni are you sure that the link is correct? Doesn't look like anything about semicolons ;)

lencioni:
@seb0zz Yes, it is demonstrating some cases where not having semicolons introduces ambiguity which causes unexpected behavior. I think the point that the author is making is that while the ES5 example is pretty unlikely to happen in the real world, the ES6 example is likely to happen in the real world because we now have array destructuring.

TL;DR:

This code with semicolons will assign `num` to `1`:

```js
[...]

PR #1573, opened 2017-10-01
The semicolon is YAGNI

thepeoplesbourgeois:
Since JavaScript is already interpreting newlines as the ends of developers' statements, and since developers shouldn't write more than one statement per line, semicolons are fundamentally unnecessary to a working JavaScript program. Much like [the cape on a superhero's back](https://www.youtube.com/watch?v=4R2aW03pwL0&t=1s), they are a holdover from a bygone era that serve only as a means for modern developers to trip over themselves.

If anyone has an example of any instance when the lack of a semicolon has caused unintended behavior in a modern browser, I _might_ reconsider my stance, _if_ Babel ever ceases to insert its own semicolons into its transpiled code

### DIFF
--- a/README.md
+++ b/README.md
@@ -2661,30 +2661,44 @@ Other Style Guides
 
 ## Semicolons
 
-  <a name="semicolons--required"></a><a name="20.1"></a>
-  - [21.1](#semicolons--required) **Yup.** eslint: [`semi`](http://eslint.org/docs/rules/semi.html) jscs: [`requireSemicolons`](http://jscs.info/rule/requireSemicolons)
+  <a name="semicolons--forbidden"></a><a name="20.1"></a>
+  - [21.1](#semicolons--forbidden) **Nope.** eslint: [`semi`](http://eslint.org/docs/rules/semi.html) jscs: [`disallowSemicolons`](http://jscs.info/rule/disallowSemicolons)
+  
+   > Why? semicolons help JavaScript transpilers minify your code, but since JavaScript is already interpreting your newline characters as statement breaks, and since you should never write more than one statement per line, semicolons are YAGNI from the developer's perspective. Also, since the semicolon stands out as the _only_ unary punctuation mark that _doesn't_ continue a statement to the next line, removing them from your code will immediately prevent confusion around whether or not a punctuation mark continues a statement to the next line.
 
-    ```javascript
-    // bad
-    (function () {
-      const name = 'Skywalker'
-      return name
-    })()
+   ```javascript
+   // bad
+   (function cuteAnimals() {
+     const animals = [new Puppy(), new Kitty(), new Snakey(), new Fishy(), new Rhinocery()];
+     return animals.
+              filter(animal => animal.hasCuteNose() && animal.isAnimal()).
+              map(animal => animal.giveCuddles({quota: false})).
+              reduce((animal, myFriends) => (myFriends[animal.name] = animal), {});
+   })();
 
-    // good
-    (function () {
-      const name = 'Skywalker';
-      return name;
-    }());
+   // bad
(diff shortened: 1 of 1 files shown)


kesne:
This has been brought up a few times in the past: https://github.com/airbnb/javascript/issues/394 and https://github.com/airbnb/javascript/issues/1567#issuecomment-331642928 come to mind.

To quote @ljharb, "semicolons are so necessary the language puts them in for you if you forget".

thepeoplesbourgeois:
> Relying on ASI has caused many bugs; anyone who omits semicolons and claims not to have run into bugs as a result just hasn't been programming long enough.

Respectfully, it would benefit not just your argument, but JavaScript developers too new to have encountered any of the dire errors you vaguely warn of, to have any examples of such errors. Don't just hand-wave someone you don't know as not knowing as much as you; guide them to where you are.

thepeoplesbourgeois:
Without any cases of misbehavior, to my years of perspective, this _appears to be_ a solved problem

goatslacker:
TBH I've wasted more time adding/removing unnecessary semicolons than I've ever wasted debugging ASI issues. 

Examples:

```js
class Foo {
}; // <-- does a semicolon go here (it doesn't)

const elements = path.node.elements.map(element => (
  wrapStatement(
    blockStatement([returnStatement(element)])
  ); // <-- oops!
));
```

More examples: https://gist.github.com/ryanflorence/61935031ff729f072d9b

There's FUD on both sides of the argument, and it's a matter of preference. Airbnb chooses to use semicolons. You're free to use our style guide and omit the semicolons though! [I regularly do it myself](https://github.com/goatslacker/alt/blob/master/.eslintrc#L14).

ljharb:
It's more than just that; ASI rules are so problematic that class fields are being held up on the question of whether or not to forbid ASI in field definitions.

thepeoplesbourgeois:
After reading a number of blogs defending the semicolon a voice I'd call "[blindingly furious](http://benalman.com/news/2012/04/semicolons-required-in-javascript/)", I finally stumbled upon the [language creator's opinion](https://brendaneich.com/2012/04/the-infernal-semicolon/) as well as a defense of letting your team [decide its own best approach](http://blog.izs.me/post/2353458699/an-open-letter-to-javascript-leaders-regarding) based on how comfortable they are with ASI.

Had I known there were such painful eccentricities to the way JavaScript regards what would clearly be expression closures and array/list literals in whitespace-terminated languages, I would probably not have written this pull request. But, since I am here now, given my new understanding of why Airbnb enforces semicolons, maybe it would be constructive of me to contribute a "**why**" explaining the gotchas and the "explicit > implicit" preference? Ideally, it would prevent similarly cavalier outsiders from suggesting the change again :P

lencioni:
@thepeoplesbourgeois that sounds like a good idea to me. It is always worthwhile to clarify reasoning.

ljharb:
That sounds great (please note, isaac's post there does not include new ASI hazards in ES6, nor upcoming ones in class fields).

I'd be happy to review a PR that explains rationales.

thepeoplesbourgeois:
(for my money, I think the class field problem could be solved without arbitary lookaheads, if infix operators could only appear _before_ line terminators while the most recently-entered closure was a class, but I am not here to fix ASI, and there are almost definitely still other complexities I didn't account for anyway)

I'll write up a paragraph later this afternoon!

ljharb:
If you have ideas about that for class fields, it's not too late - we're discussing it in November. Please feel free to comment on https://github.com/tc39/proposal-class-fields/issues/7.

Look forward to the prose PR!

thepeoplesbourgeois:
Likely because I rewrote history on my master branch, I needed to open a new PR for the updates: https://github.com/airbnb/javascript/pull/1578

PR #1578, opened 2017-10-04
Add explanation for semicolon enforcement rule

thepeoplesbourgeois:
Also update code samples to highlight ways in which ES6
currently mishandles code when relying on Automatic Semicolon
Insertion

### DIFF
--- a/README.md
+++ b/README.md
@@ -2767,24 +2767,45 @@ Other Style Guides
   <a name="semicolons--required"></a><a name="20.1"></a>
   - [21.1](#semicolons--required) **Yup.** eslint: [`semi`](https://eslint.org/docs/rules/semi.html) jscs: [`requireSemicolons`](http://jscs.info/rule/requireSemicolons)
 
+    > Why? When JavaScript encounters a line break without a semicolon, it uses a set of rules called [Automatic Semicolon Insertion](https://tc39.github.io/ecma262/#sec-automatic-semicolon-insertion) to determine whether or not it should regard that line break as the end of a statement, and (as the name implies) place a semicolon into your code before the line break if it thinks so. ASI contains a few eccentric behaviors, though, and your code will break if JavaScript misinterprets your line break. These rules will become more complicated as new features become a part of JavaScript. Explicitly terminating your statements and configuring your linter to catch missing semicolons will help prevent you from encountering issues.
+
     ```javascript
-    // bad
-    (function () {
-      const name = 'Skywalker'
-      return name
-    })()
+    // bad - raises exception
+    const luke = {}
+    const leia = {}
+    [luke, leia].forEach(jedi => jedi.father = 'vader')
+
+    // bad - raises exception
+    const reaction = "No! That's impossible!"
+    (async function meanwhileOnTheFalcon(){
+      // handle `leia`, `lando`, `chewie`, `r2`, `c3p0`
+      // ...
+    }())
+
+    // bad - returns `undefined` instead of the value on the next line - always happens when `return` is on a line by itself because of ASI!
+    function foo() {
+      return
+        'search your feelings, you know it to be foo'
+    }
 
     // good
-    (function () {
-      const name = 'Skywalker';
-      return name;
+    const luke = {};
+    const leia = {};
+    [luke, leia].forEach((jedi) => {
+      jedi.father = 'vader';
+    });
+
+    // good
+    const reaction = "No! That's impossible!";
+    (async function meanwhileOnTheFalcon(){
+      // handle `leia`, `lando`, `chewie`, `r2`, `c3p0`
+      // ...
     }());
 
-    // good, but legacy (guards against the function becoming an argument when two files with IIFEs are concatenated)
-    ;((() => {
-      const name = 'Skywalker';
-      return name;
-    })());
+    // good
+    function foo() {
+      return 'search your feelings, you know it to be foo';
+    }
     ```
 
     [Read more](https://stackoverflow.com/questions/7365172/semicolon-before-self-invoking-function/7365214#7365214).


README.md
```diff
@@ -2665,24 +2665,37 @@ Other Style Guides
...
+    }
 
     // good
-    (function () {
-      const name = 'Skywalker';
-      return name;
-    }());
+    const luke = {};
+    const leia = {};
+    [luke, leia].forEach(jedi => jedi.father = 'vader');
```
ljharb: this is a bad example; elsewhere in the guide we require that this arrow function be surrounded by curly braces

README.md
```diff
@@ -2665,24 +2665,37 @@ Other Style Guides
   <a name="semicolons--required"></a><a name="20.1"></a>
   - [21.1](#semicolons--required) **Yup.** eslint: [`semi`](http://eslint.org/docs/rules/semi.html) jscs: [`requireSemicolons`](http://jscs.info/rule/requireSemicolons)
 
+    > Why? When JavaScript encounters a line break without a semicolon, it uses a set of rules called [Automatic Semicolon Insertion](https://www.ecma-international.org/ecma-262/5.1/#sec-7.9) to determine whether or not it should regard that line break as the end of a statement, and (as the name implies) place a semicolon into your code before the line break if it thinks so. ASI contains a few eccentric behaviors, though, and your code will break if JavaScript misinterprets your line break. These rules will become more complicated as new features become a part of JavaScript. Explicitly terminating your statements and configuring your linter to catch missing semicolons will help prevent you from encountering issues.
```
ljharb: let's instead link to the ASI rules on [github](http://tc39.github.io/ecma262/), which is a more up to date source than ecma's website

README.md
thepeoplesbourgeois: Ah, right, because mutations are side effects 😅

README.md
```diff
@@ -2665,24 +2664,39 @@ Other Style Guides
...
+    }
 
     // good
-    (function () {
-      const name = 'Skywalker';
-      return name;
-    }());
+    const luke = {};
+    const leia = {};
+    [luke, leia].forEach(jedi => {
```
goatslacker: This example is fine for illustrative purposes but it's not typical nor great when a line is started by `[`. This array needs a name.

README.md
```diff
@@ -2665,24 +2664,39 @@ Other Style Guides
...
+    });
 
-    // good, but legacy (guards against the function becoming an argument when two files with IIFEs are concatenated)
-    ;((() => {
-      const name = 'Skywalker';
-      return name;
-    })());
+    // good
+    const foo = 2;
+    ("No! That's impossible!".repeat(foo));
```
goatslacker: Same here, there's no reason this needs to be wrapped in parens (other than for the illustration) it seems odd that this is in a "good" section.

README.md
```diff
@@ -2665,24 +2664,39 @@ Other Style Guides
...
+    // bad - raises exception
+    const luke = {}
+    const leia = {}
+    [luke, leia].forEach(jedi => jedi.father = 'vader')
+
+    // bad - raises exception
+    const foo = 2
+    ("No! That's impossible!".repeat(foo))
+
+    // bad - returns `undefined` instead of the value on the next line - always happens when `return` is on a line by itself because of ASI!
```
goatslacker: adding a semicolon here wouldn't have saved you:

```js
function foo() {
  return
    'search your feelings, you know it to be foo';
}
```

Also, single quotes everywhere instead of double plz :)

goatslacker: ^ what I mean by this is that this isn't ASI's fault or because of a lack of adding a semicolon. It's not a reason to use or not use a semicolon.

README.md
ljharb: assignments aren't mutations tho :-)

README.md
thepeoplesbourgeois: As far as whether or not it's a reason to use _semicolons,_ you are correct: the act of a developer adding this semicolon is not, itself, the thing that "would have saved you". However, a linter or IDE that noticed a `return` by itself would note that there was a semicolon missing on the line _containing_ the return. Semicolons would not _directly_ save you, but something warning you that these two lines appeared amiss would.

README.md
thepeoplesbourgeois: Indeed, it is difficult to mirror the real-world usages of code within example blocks. I would also be shunned for my terrible variable names, I think ;)

README.md
thepeoplesbourgeois: Oh? Hmm... [Is that a new rule?](https://github.com/airbnb/javascript#arrow-functions)

README.md
thepeoplesbourgeois: Thank you. Semantics are difficult but quite important ☝️

thepeoplesbourgeois: Although, it seems somewhat odd to consider an alteration to an object that will alter the return of both `obj.hasOwnProperty(key)` and `Object.keys(obj)` as **not** a mutation to the structure of the object. A very fast Google of the subject implies that [doing exactly this is, indeed, a mutation](https://slemgrim.com/mutate-or-not-to-mutate/). Perhaps I need more research.

thepeoplesbourgeois: Further research of [Mozilla's documentation](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Object/freeze) of how `Object.freeze` makes objects _im_-mutable implies that creating a new property on one, or changing a property already within one, is indeed a mutation

ljharb: @thepeoplesbourgeois ha, touché. `jedi.father =` is indeed a mutation :-)

however the reason this style isn't permitted isn't because it's a mutation - `jedi => Object.assign(jedi, { father: 'vader' })` would be fine - but because it's an assignment statement.

README.md
goatslacker: Oh yeah we should change both :P but that's for another PR I guess.

PR #1579, opened 2017-10-05
Fix a missing return statement in 8.2 rule's example.

pqkluan:
Nothing much to discuss, just a simple fix on [8.2](https://github.com/airbnb/javascript#arrows--implicit-return) last example.

### DIFF
--- a/README.md
+++ b/README.md
@@ -973,7 +973,7 @@ Other Style Guides
 
     // good
     foo(() => {
-      bool = true;
+      return bool = true;
     });
     ```
 


pqkluan:
Find out that the fix is incomplete, will create another pull.

ljharb:
@pqkluan please do not create another one; please reuse this one.

README.md
```diff
@@ -973,7 +973,7 @@ Other Style Guides
 
     // good
     foo(() => {
-      bool = true;
+      return bool = true;
```
ljharb: altho in this case, your change is bad - we should never conflate assignment with expressions.

PR #1581, opened 2017-10-09
amend `jsx-a11y/label-has-for` option

1pete:
`jsx-a11y` introduced a new option `allowChildren` to rule `label-has-for` which default to `false` which doesn't allow label tag to have text in it e.g.
```jsx
<label htmlFor="username">Username</label>
```
this pr is intended to allow such style

[rule reference](https://github.com/evcohen/eslint-plugin-jsx-a11y/blob/master/docs/rules/label-has-for.md)

### DIFF
--- a/packages/eslint-config-airbnb/rules/react-a11y.js
+++ b/packages/eslint-config-airbnb/rules/react-a11y.js
@@ -48,7 +48,7 @@ module.exports = {
 
     // require that JSX labels use "htmlFor"
     // https://github.com/evcohen/eslint-plugin-jsx-a11y/blob/master/docs/rules/label-has-for.md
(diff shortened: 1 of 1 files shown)


ljharb:
That's not what `allowChildren` is for; it's to allow you to bypass this rule if you're passing an input in as children.

This should remain `false`; label children should remain hardcoded inline, and use both nesting and for/ID linking for a11y.

1pete:
My mistake, you are correct. Surprisingly I got an error and after i added that option, it turned passed, so i misunderstood.

In fact, it's another option, i'll make another pr.

1pete:
My bad again, I was going make a pr to require nesting or for/ID or both, and I just read the part where you intend to require both nesting and for/ID, so no pr

PR #1583, opened 2017-10-10
Do not prefer destructuring for object assignment expression

frenzzy:
```js
let param = 'default value';

if (object)
  param = object.param; // <= prefer-destructuring
}

// ...
```
I think that `param = object.param;` is better than `({ param } = object);` because it is easier to read and tedious to write brackets by hand.

MDN: [Destructuring assignment](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Destructuring_assignment)

### DIFF
--- a/packages/eslint-config-airbnb-base/rules/es6.js
+++ b/packages/eslint-config-airbnb-base/rules/es6.js
@@ -118,7 +118,7 @@ module.exports = {
       },
       AssignmentExpression: {
         array: true,
-        object: true,
+        object: false,
       },
     }, {
       enforceForRenamedProperties: false,


frenzzy:
One more case:
```js
const fields = _.pick(req.body, allowedFieldNames);

if (!fields.senderEmail) {
  // AssignmentExpression { object: true }
  fields.senderEmail = siteContact.senderEmail; // <= error
  ({ fields.senderEmail } = siteContact); // <= ok
}

if (isValid(fields.recipientEmail)) {
  // AssignmentExpression { array: true }
  fields.recipientEmail = fields.recipientEmail.split(',')[0]; // <= error
  [fields.recipientEmail] = fields.recipientEmail.split(','); // <= ok
}
```
If you think that destructuring assignments are good, let's just close this.

ljharb:
@frenzzy to be fair, that example is mutation, which this guide effectively forbids. Try:
```js
const fields = {
  senderEmail: siteContact.senderEmail,
  recipientEmail: siteContact.recipientEmail,
  ...(_.pick(req.body, allowedFieldNames)),
};
```

Typically when you find awkward code based on our linter rules, it's because the pattern itself is already forbidden or discouraged elsewhere :-)

DanielGibbsNZ:
Some code that I'm working with uses a similar pattern to the original reporter in quite a few places:

```
let foo
switch (condition) {
  case SOMETHING.A:
    // other stuff
    foo = object1.foo
    break
  case SOMETHING.b:
    // other stuff
    foo = object2.foo
    break
  case SOMETHING.C:
    // other stuff
    foo = object3.foo
    break
  default:
    // other stuff
    foo = 'fallback value'
}
```

With the default rule, this forces me to replace the (arguably more readable) `foo = object1.foo` with `;({ foo } = object1)`.

PR #1585, opened 2017-10-11
Support unused vars

Runrioter:
Sometimes, I will have the following code:
```javascript
const express = require('express');

const router = express.Router();
router.get('/', (req, res, next) => {
  res.render('index', { title: 'Express' });
});
```
Currently, `next` is not used, so I want to support it like this:

```javascript
const express = require('express');

const router = express.Router();
router.get('/', (req, res, _) => {
  res.render('index', { title: 'Express' });
});
```
And underscore is natively for unused variable for programming language like Swift.
```swift
func f(p :String, _ :String) {
       // ...
}
```

I think that `_` is a good choice.

### DIFF
--- a/packages/eslint-config-airbnb-base/rules/variables.js
+++ b/packages/eslint-config-airbnb-base/rules/variables.js
@@ -36,7 +36,7 @@ module.exports = {
     'no-undefined': 'off',
 
     // disallow declaration of variables that are not used in the code
-    'no-unused-vars': ['error', { vars: 'all', args: 'after-used', ignoreRestSiblings: true }],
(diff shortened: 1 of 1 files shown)


ljharb:
If it's not used, why name it in the signature?

Runrioter:
Ok, You are right. 
But I think that keeping function signature(middleware from some web framework) consistent is better and readable.
And sometimes unused var can't be omitted. For example:
```javascript
try {
    const obj = JSON.parse(str);
    return obj;
} catch (_) {
    return null;
}
```

Runrioter:
@ljharb  Seems that you think that it is unnecessary, I maybe need to make a new repo or just override some eslint rule in config file. Thank you.

ljharb:
When the unused var *can't* be omitted, eslint will allow it.

PR #1587, opened 2017-10-12
change year

minimooooz:


package.json
```diff
@@ -25,7 +25,7 @@
     "lint",
     "airbnb",
     "es6",
-    "es2015",
+    "es2017",
```
ljharb: i'd be fine adding `es2016`, `es2017`, and `es2018`; but i don't want to remove `es2015`.

Also, we'd want to make these changes in the `packages/` dir's package.json, not in the root (that's not published anymore)

graingert: babel like to use `ES2015+` to define what eslint-preset-env does. Maybe add that as a kw?

PR #1595, opened 2017-10-16
breaking: disallow else if in no-else-return

graingert:


ljharb:
There's no need to make a PR to update things; v12.1.0 of eslint-config-airbnb-base is now released that requires eslint v4.9.

The breaking change to disallow `else if` doesn't need to happen in a hurry; please do rebase this (to simply remove the TODO and flip the boolean) and I'll rebase and merge it when it's time.

graingert:
@ljharb there's no mention of `else if` in the guide. So I'm pretty sure there's no guide changes needed.

ljharb:
@graingert that's exactly why it needs to be added; in general we shouldn't be enforcing rules that aren't documented in prose in the guide (there's a number of places this isn't yet the case, of course, but that doesn't change our best practice)

ljharb:
Turns out that there's no `no-else-return` section at all ¯\\\_(ツ)_/¯

Issue #794, opened 2016-03-21
7.1 says function expressions are always unnamed, but this changed in ES6


Arnavion:
[7.1](https://github.com/airbnb/javascript#functions--declarations) contains this statement:

>Function declarations are named, so they're easier to identify in call stacks.

Anonymous function expressions assigned to a binding are *also* named, per [ES6 semantics (1.e.iii).](http://www.ecma-international.org/ecma-262/6.0/#sec-assignment-operators-runtime-semantics-evaluation) That is, the "bad" example `const foo = function () { };` is the same as `const foo = function foo() { };`, which is equivalent to the "good" example `function foo() { }` in that respect.

Should the statement be qualified that it's only a distinction for ES5 and below ?

ljharb:
Function name inference is implemented in hardly any of the browsers we support. In addition, relying on function name inference hinders greppability, so even in a pure ES6 environment, we wouldn't recommend that.

(other subtler differences are that a function with an inferred name still doesn't have a lexical name binding in the function body, but that's not relevant here)

Arnavion:
Do you want to clarify that in the document itself? Via footnote, etc.

ljharb:
Sure, that's a great idea for a footnote!

tandav:
... frome Style Guide:
```js
// bad
const foo = function () {
  // ...
};

// good
const foo = function bar() {
  // ...
};
```

So, you say, there should be 2 different names for one function? 

For example, 

```js
// Is it worse
const sum = function(a, b) {
  return a + b;
};

// than this?
const my_sum = function sum(a, b) {
  return a + b;
};
```
I'm a begginer in JS/ES, can you tell, please, what are advantages of second method?

ljharb:
@tandav In your last example, the first function doesn't have a `.name` - which means in debugging, it's likely to show up as an anonymous function. In the latter example, it will show up with the name `sum` - and when you grep for it, all the uses in the file will use `my_sum`, so the actual function will be easier to locate.

JPeer264:
@ljharb I tested it using babel and `babel-preset-env`:
```js
const foo = () => {}
```
will be converted to
```js
var foo = function foo () {}
```

So it does have an `.name`. Also since Node v6 `const foo = () => {}` does have `foo` as `.name` (without babel). So it is not anonymous anymore.

ljharb:
Yes, that's function name inference. Which can't always be relied on, and isn't supported in older engines.

[...]

PR #1607, opened 2017-10-20
Clarify reasoning for named function expressions

jabacchetta:
For convenience, added brief explanations that are given in the linked
discussion — #794 — mostly eliminating the need to scroll through the comments.


jabacchetta:
The discussion link could also potentially be removed from the "Why?" block with these changes.

PR #1608, opened 2017-10-21
To Clarify the func-style description

liuchong:
As "foo" is still meaningless, but with "short", we can see that, we are intend to use a short alias name to make our coding happily.

refer to https://github.com/airbnb/javascript/issues/794

### DIFF
--- a/README.md
+++ b/README.md
@@ -648,7 +648,7 @@ Other Style Guides
 
     // good
     //  lexical name distinguished from the variable-referenced invocation(s)
-    const foo = function uniqueMoreDescriptiveLexicalFoo() {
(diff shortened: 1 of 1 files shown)


liuchong:
It's strange that the mark down style breaks, I've check it and found this:

```
diff README.md README.md.new | cat -A
651c651$
<     const foo = function uniqueMoreDescriptiveLexicalFoo() {$
---$
>  M-BM-  M-BM- const short = function longUniqueMoreDescriptiveLexicalFoo() {$
```

not sure where it came from, maybe my input software. I was just deleted them an input the blanks again, and it was fixed, sorry about it, this a the new pull request:
https://github.com/airbnb/javascript/pull/1610

ljharb:
Thanks

Issue #1071, opened 2016-09-13
Mixed operators with same precedence


aboyton:
The ESLint rule [no-mixed-operators](http://eslint.org/docs/rules/no-mixed-operators) is great for making things like

``` javascript
var foo = a && b < 0 || c > 0 || d + 1 === 0;
var bar = a && b || c;
```

clearer by adding parentheses (rather than having people remember the precedence of `&&` and `||`. That said, I'm unconvinced that

``` javascript
var foo = a + b - c
```

is clearer with parentheses, that is according to this style guide you'd need to write this as

``` javascript
var foo = (a + b) - c
```

The [no-mixed-operators](http://eslint.org/docs/rules/no-mixed-operators) rule has an option for this, `allowSamePrecedence` and the default is `true` meaning that `var foo = a + b - c` is valid code (which I think makes a lot more sense). Is there any reason for having this set to `false`?

ljharb:
Yes - `||` and `&&` also have the same precedence, and the parens are very necessary there to avoid ambiguity.

The change I've been considering here is removing some of the arithmetic operators entirely - ie, `+ - * /` but leaving the settings otherwise unchanged. Clearly with `+` and `-` though, there's some tweaking that would need to be done.

aboyton:
Thanks for the quick reply.

Actually `||` and `&&` do have different precedence (see [MDN operator precedence](https://developer.mozilla.org/en/docs/Web/JavaScript/Reference/Operators/Operator_Precedence) or [Microsoft operator precedence](https://msdn.microsoft.com/en-us/library/z3ks45k7(v=vs.94).aspx)). As an example `true || true && false` is `true` as `&&` has a higher precedence than `||`.

Because of this (with `"allowSamePrecedence": true`) the following are lint errors:

```javascript
var foo = a || b && c;
var foo = a + b * c;
```

but these are fine

```javascript
var foo = a || (b && c);
var foo = a + (b * c);
var foo = a + b - c;
```

I personally don't mind which way you go for `a + b * c` or `a + (b * c)` (I never get confused here but I get how some people do) but it would be nice to make `a + b - c` valid code.

blord-fullscreen:
Any reply on this? Many people work in an environment where this package is treated as holy writ, and changing the eslintrc is not politically possible. Somewhere out there an intern is rewriting addition and subtraction operations to add parens, because `a + b - c` is invalid.

ljharb:
[...]

PR #1611, opened 2017-10-23
Changes the eslint `no-mixed-operators` rule to only warn on mixing `**` and `%` with arithmetic operators

sharmilajesupaul:
Changes the eslint `no-mixed-operators` rule to no longer warn on mixing common arithmetic operators unless they are mixed with `**` or`%`.

(Related Comment: https://github.com/airbnb/javascript/issues/1071#issuecomment-337913745)

**Guide Description:**

Enclose operators in parentheses when they are mixed in a statement. When mixing arithmetic operators, do not mix `**` and `%` with themselves or with `+`, `-`, `*`, & `/`. eslint: [`no-mixed-operators`](http://eslint.org/docs/rules/no-mixed-operators.html)

> Why? This improves readability and clarifies the developer’s intention.

```javascript
    // bad
    const foo = a && b < 0 || c > 0 || d + 1 === 0;

    // bad
    const bar = a ** b - 5 % d;

    // bad
    if (a || b && c) {
      return d;
    }

    // good
    const foo = (a && b < 0) || c > 0 || (d + 1 === 0);

    // good
    const bar = (a ** b) - (5 % d);

    // good
    if ((a || b) && c) {
      return d;
    }

    // good
    const bar = a + b / c * d;
```

### DIFF
--- a/README.md
+++ b/README.md
@@ -1943,6 +1943,38 @@ Other Style Guides
     const baz = !c;
     ```
 
+  <a name="comparison--no-mixed-operators"></a><a name="15.8"></a>
+  - [15.7](#comparison--no-mixed-operators) Enclose operators in parentheses when they are mixed in a statement. When mixing arithmetic operators, do not mix `**` and `%` with themselves or with `+`, `-`, `*`, & `/`. eslint: [`no-mixed-operators`](http://eslint.org/docs/rules/no-mixed-operators.html)
+
+    > Why? This improves readability and clarifies the developer’s intention.
+
+    ```javascript
+    // bad
+    const foo = a && b < 0 || c > 0 || d + 1 === 0;
+
+    // bad
+    const bar = a ** b - 5 % d;
+
+    // bad
+    if (a || b && c) {
+      return d;
+    }
+
+    // good
+    const foo = (a && b < 0) || c > 0 || (d + 1 === 0);
+
+    // good
+    const bar = (a ** b) - (5 % d);
+
+    // good
+    if ((a || b) && c) {
+      return d;
+    }
+
+    // good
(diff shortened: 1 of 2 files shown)


README.md
```diff
@@ -1943,6 +1943,38 @@ Other Style Guides
     const baz = !c;
     ```
 
+  <a name="comparison--no-mixed-operators"></a><a name="15.8"></a>
+  - [15.7](#comparison--no-mixed-operators) Enclose operators in parenthese when they are mixed in a statement. When mixing arthimetic operators, do not mix `**` and `%` with themselves or with `+`, `-`, `*`, & `/`. eslint: [`no-mixed-operators`](http://eslint.org/docs/rules/no-mixed-operators.html)
```
ljharb: parenthese → parentheses

ljharb: arthimetic → arithmetic

sharmilajesupaul:
fixed typos @ljharb, ptal 🙂

bensaufley:
When will this be in a release? I'm currently getting errors for simple mixed PEMDAS operators, which this change appears to fix. Looks like [the last release was Oct 16](https://github.com/airbnb/javascript/releases/tag/eslint-config-airbnb-base-v12.1.0)?

ljharb:
@bensaufley the next release will be a breaking one, so i'm trying to include as many of those as possible. Hopefully in the next few weeks.

PaQ666:
I think someone misunderstood this setting configuration, as with current config it would trigger error when mixing + and - for example.

bensaufley:
@paq666 this is the PR that fixes that but it still hasn’t made it into a release.

PR #1618, opened 2017-10-25
Enables eslint rule: nonblock-statement-body-position

sharmilajesupaul:
Enables [nonblock-statement-body-position](https://eslint.org/docs/rules/nonblock-statement-body-position) eslint rule and adds the link to the related guide section.

### DIFF
--- a/packages/eslint-config-airbnb-base/rules/style.js
+++ b/packages/eslint-config-airbnb-base/rules/style.js
@@ -354,7 +354,7 @@ module.exports = {
 
     // enforce the location of single-line statements
     // http://eslint.org/docs/rules/nonblock-statement-body-position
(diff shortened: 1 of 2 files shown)


packages/eslint-config-airbnb-base/rules/style.js
```diff
@@ -354,7 +354,7 @@ module.exports = {
 
     // enforce the location of single-line statements
     // http://eslint.org/docs/rules/nonblock-statement-body-position
-    'nonblock-statement-body-position': 'off',
+    'nonblock-statement-body-position': 'error',
```
ljharb: we generally prefer to be explicit about what setting we're using, even if it's the default - could we change this to specify `['error', 'beside', { overrides: {} }]`?

PR #1629, opened 2017-10-31
Add section about not spreading props

### LINKED ISSUE #1443: Add section about not spreading props
Spread props should be avoided, except for exceptions like:
 - HOCs that proxy down props and hoist propTypes
 - spreading objects with known, explicit props - like if you construct an object literal (statically or via a "pick" pattern) and then promptly spread it
 - others?

Let's add a section to the React guide about this.

maggielove:
Fixes #1443

### DIFF
--- a/react/README.md
+++ b/react/README.md
@@ -380,6 +380,58 @@
   };
   ```
 
+  - Use spread props sparingly.
+  > Why? Otherwise you're more likely to pass unnecessary props down to components. And for React v15.6.1 and older, you could [pass invalid HTML attributes to the DOM](https://reactjs.org/blog/2017/09/08/dom-attributes-in-react-16.html).
+
+  Exceptions:
+
+  - HOCs that proxy down props and hoist propTypes
+
+  ```jsx
+  function HOC(WrappedComponent) {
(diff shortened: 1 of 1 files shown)


maggielove:
cc @ljharb

react/README.md
```diff
@@ -380,6 +380,58 @@
   };
   ```
 
+  - Use spread props sparingly.
+  > Why? Otherwise you're more likely to pass unnecessary props down to components. And for React 15 and older, you could [pass invalid HTML attributes to the DOM](https://reactjs.org/blog/2017/09/08/dom-attributes-in-react-16.html).
```
ljharb: I think it's React < 15.2 or 15.3? Let's be specific here :-)

maggielove:
Thanks @ljharb ! This is ready for re-review. :) I'm not sure why the Travis build is failing--looking into it.

ljharb:
Tests will now pass once this is rebased on master.

maggielove:
@ljharb Whenever you get a chance, I made your suggested changes. What do you think -- does it need more edits?

PR #1631, opened 2017-11-02
Fix confusing method name in Classes & Constructor

dubbha:
Why? The method in the current example is called `pop()` but it actually implements an [Array.prototype.shift()](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/shift) method which is confusing for no good reason. It just does it on the wrong side of an array.

I understand the change clutters the example code a bit, but at least it's less confusing.

Another approach--probably a better one--would be to change the name of the method from `pop` to `dequeue` leaving the implementation as is. The `peek` name and implementation could be left as is too then.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1066,8 +1066,8 @@ Other Style Guides
       this.queue = [...contents];
     }
     Queue.prototype.pop = function () {
-      const value = this.queue[0];
-      this.queue.splice(0, 1);
+      const value = this.queue[this.queue.length - 1];
+      this.queue.splice(-1, 1);
       return value;
     };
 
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -1066,8 +1066,8 @@ Other Style Guides
       this.queue = [...contents];
     }
     Queue.prototype.pop = function () {
-      const value = this.queue[0];
-      this.queue.splice(0, 1);
+      const value = this.queue[this.queue.length - 1];
```
ljharb: it's a queue tho - in a queue, you only ever pop off of the front - it's FIFO.

dubbha:
True. I'm just saying it's a bit confusing with the array methods naming, which are kind of stack-based (push/pop on th back, shift/unshift on the front). That's why I said we could alternatively just rename it to "dequeue".
Or we can leave it as is, and just close the PR, not a big deal really :)

ljharb:
The only confusion I can see is if you're under the impression that anything with a "pop" method is an array; since that's not the case, I think it's probably fine as-is. Thanks for the contribution tho!

Issue #1592, opened 2017-10-13
Collision with `import/extensions`


tunnckoCore:
input

```js
import utils from './lib/utils';
console.log(utils);
```

errors with `Missing file extension for "./lib/utils"  import/extensions`

When set the file extension

```js
import utils from './lib/utils.js';
console.log(utils);
```
errors with `Unexpected use of file extension "js" for "./lib/utils.js"  import/extensions`

charlike:
https://github.com/airbnb/javascript/blob/master/packages/eslint-config-airbnb-base/rules/imports.js#L132-L137
and https://github.com/airbnb/javascript/blob/master/packages/eslint-config-airbnb-base/rules/imports.js#L19-L22

ljharb:
All code should omit extensions always - this is a best practice, so that refactoring is enabled and "which extension" can be delegated to node, or a bundler.

https://github.com/airbnb/javascript/blob/master/packages/eslint-config-airbnb-base/rules/imports.js#L132-L137 says that extensions are always required, except for .js and .jsx files, for which it's always forbidden.

https://github.com/airbnb/javascript/blob/master/packages/eslint-config-airbnb-base/rules/imports.js#L19-L22 is a *setting*, not a rule - and it informs eslint-plugin-import that `jsx` is an extension it should look at.

I'm not sure what conflict you're talking about.

Re your example above, `<script type="module">` can't appear in a JS file - and regardless, in node, `.mjs` will be the file extension required for ES Modules - so the fact that browsers can accept a .js file under `type="module"` is irrelevant.

I'll be updating the rules soon to also include `mjs` in the list of supported extensions, and configure it to always forbid the extension from being specified.

As far as your original example: if `lib/utils.js` exists, I'm not sure why the conflict occurs. Do you also have any other files in `lib` named `utils`, with any other extension?

zidian257:
I had the same issue.

steveja42:
I think I'm having the same issue. Chrome browser doesn't allow import without the extension. Example:
import { log, assert } from './utils';
results in error:
Failed to load resource: the server responded with a status of 404 (Not Found) [http://localhost:9000/utils]
but import { log, assert } from './utils.js' 
works with no error.

tunnckoCore:
@ljharb 

> Do you also have any other files in lib named utils, with any other extension?

No.

[...]

PR #1635, opened 2017-11-05
Add .mjs extension support

### LINKED ISSUE #1634: Support .mjs extension
When using [ECMAScript Modules](https://nodejs.org/api/esm.html#esm_ecmascript_modules) in node, ESLint complains because the config doesn't yet support the `.mjs` extension.

olistic:
Fixes #1634 following what I think was @ljharb's intent, as per [his comment here](https://github.com/airbnb/javascript/issues/1592).

- [x] Add `.mjs` to the list of supported extensions
- [x] Forbid `.mjs` extension from being specified

### DIFF
--- a/packages/eslint-config-airbnb-base/rules/imports.js
+++ b/packages/eslint-config-airbnb-base/rules/imports.js
@@ -13,11 +13,12 @@ module.exports = {
   settings: {
     'import/resolver': {
       node: {
-        extensions: ['.js', '.json']
+        extensions: ['.mjs', '.js', '.json']
       }
     },
     'import/extensions': [
       '.js',
+      '.mjs',
       '.jsx',
     ],
     'import/core-modules': [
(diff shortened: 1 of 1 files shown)


packages/eslint-config-airbnb-base/rules/imports.js
```diff
@@ -13,11 +13,12 @@ module.exports = {
   settings: {
     'import/resolver': {
       node: {
-        extensions: ['.js', '.json']
```
ljharb: the order here needs to have `.mjs` first.

packages/eslint-config-airbnb-base/rules/imports.js
```diff
@@ -13,11 +13,12 @@ module.exports = {
...
+        extensions: ['.js', '.mjs', '.json']
       }
     },
     'import/extensions': [
       '.js',
+      '.mjs',
```
ljharb: I checked to be sure; order [doesn't matter here](https://github.com/benmosher/eslint-plugin-import/blob/dc3609f895f5ad390d88fd9925dce40960117930/utils/ignore.js#L22)

packages/eslint-config-airbnb-base/rules/imports.js
olistic: @ljharb is this so if there's a conflict, it picks `.mjs`?

ljharb: Yes, exactly; that's how it will work in node as well.

PR #1637, opened 2017-11-07
Ambiguous 'good' example

vorpus:
7.11 uses the 'bad' example from 7.1. Changed to arrow function to prevent ambiguity.

### DIFF
--- a/README.md
+++ b/README.md
@@ -806,7 +806,7 @@ Other Style Guides
     const h = function() {};
 
     // good
-    const x = function () {};
+    const x = () => {};
     const y = function a() {};
     ```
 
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -806,8 +806,9 @@ Other Style Guides
     const h = function() {};
 
     // good
-    const x = function () {};
+    const x = () => {};
```
ljharb: an arrow function doesn't have the spacing issue this section is concerned with; the fact that `function () {}` is bad doesn't mean it's not useful in this example.

I'm fine adding the arrow, but we need to keep the `function () {}`.

vorpus: @ljharb Yea I was thinking about that after I made the PR - added line 811 with a function that has a longer lexical name

vorpus: @ljharb You're probably right, it's too trivial of an inconsistency to not keep what is currently in the doc

PR #1641, opened 2017-11-20
enable import/order in eslint-config-airbnb-base

christophehurpeau:
Enable the rule "import/order"
This is a breaking change and if accepted will have to wait for a major


christophehurpeau:
You should probably remove the TODO then 😄

ljharb:
a) the todo says `// TODO: enable?`, and the question mark means we're not sure yet.
b) the todos are meant for airbnb devs to consider enabling; they're not actionable for contributors.

A great place to find things we need help on is [here](https://github.com/airbnb/javascript/issues?q=is%3Aopen+is%3Aissue+label%3A%22pull+request+wanted%22) :-)

Issue #1640, opened 2017-11-18
Explain UTF-8 BOM rule in readme


mk-pmb:
Most style decisions are explained in the readme, but I couldn't find the reasoning on why a BOM is considered bad. Where I've looked:

* [the patch](https://github.com/airbnb/javascript/commit/d9cb343b518349982f3b094f2ce48eda8347e6c7) that enabled it, found no commit message body.
* the current code of the rules file, https://github.com/airbnb/javascript/blob/8cf2c70a4164ba2dad9a79e7ac9021d32a406487/packages/eslint-config-airbnb-base/rules/style.js#L472-L474
* [the eslint rule page](https://eslint.org/docs/rules/unicode-bom) mentioned in the patch code comment, doesn't claim it's bad.
* searched the readme for "BOM", "unicode" and "byte order"
* searched issue tracker for "BOM", "unicode" and "byte order"

Could someone explain it, or add search keywords to make the explanation easier to find?

ljharb:
The rule page says "UTF-8 does not require a BOM because byte ordering does not matter when characters are a single byte. Since UTF-8 is the dominant encoding of the web, we make "never" the default option." - the reasoning for choosing "never" is because files should always only be in UTF-8.

Are you running into issues with this rule?

mk-pmb:
> Since UTF-8 is the dominant encoding of the web, we make "never" the default option.

Yes, that's what I found there as well. It's ok as a default for eslint. I thought the patch repeats this because there had been a stronger reason for "never".
In my projects I like BOMs because Firefox will trust the BOM more than it trusts the Content-Type header, thus it protects my files' encoding when used on webspace that announces another charset (e.g. legacy website). It's also helpful on webspace that doesn't send any charset info (e.g. `python -m SimpleHTTPServer`), or not even any Content-Type (`file://` access).

To reproduce, run Python 2.7.6 and Firefox 57 on Ubuntu trusty with system locale `en_US.UTF-8`. Python's SimpleHTTPServer sends just "text/plain" without a charset, probably because it doesn't consider itself authoritative to guess more details. Both this way and when loading via `file://`, without a BOM, Firefox guesses "windows-1252" and consequently garbles umlauts and emoji.

ljharb:
In general, this config repeats all the defaults explicitly.

I'd say that your solution should be to use a build process to auto-insert the BOM for you, rather than encoding that directly in the file.

[...]

PR #1643, opened 2017-11-22
[guide] Non-code issues: keep them out of your code files.

mk-pmb:
Summary of #1640. I extrapolated the reasoning to other transport layer packaging, hope it's the right conclusion.


mk-pmb:
rebased on current master.

README.md
```diff
@@ -3407,6 +3407,29 @@ Other Style Guides
...
 
+## Non-code issues
+
+Now that your code is covered, consider some aspects that are not/
+should not be about *your* code:
+
+  <a name="non-code-issues--transport-and-packaging"></a><a name="31.1"></a>
+  - [31.1](#non-code-issues--transport-and-packaging)
+    **Transport and packaging**:
+    Your code should not be concerned with how it's gonna be packaged or
```
ljharb: `s/gonna/going to`

README.md
```diff
@@ -3407,6 +3407,29 @@ Other Style Guides
...
+    transported. Caring about those aspects should be left to automated
+    tools that are part of your build stack.
```
ljharb: can you elaborate on this a bit?

mk-pmb: It's the essence I distilled from #1640 :
* Project code shouldn't carry artifacts meant to fix my editor's behavior. That's my editor's problem, not the project's.
* Project code shouldn't carry artifacts meant to fix problems with how browsers obtain the code file on an HTTP level (i.e. neither about parsing nor the code's behavior). If decoration is required to ensure safe transport of the file, some tool in the build chain should add it.

README.md
mk-pmb: fixed. :white_check_mark:

mk-pmb:
For clearer dos and don'ts about the BOM, we could use code blocks that contain just explanatory comments. Would at least keep the visual style consistent:

```javascript
/* <- bad:  Invisible BOM at start of code file, meant to
            fix your editor's charset.
      good: Set the charset in your editor config.
            If you need per-project config and your editor
            doesn't support it, get a better one.           */
```

```javascript
/* <- bad:  Invisible BOM at start of code file, meant to
            fix some browser's charset guessing.
      good: Use your webserver or some part of your toolchain
            to add the charset info.                        */
```

mk-pmb:
Draft of example for omitting potential UMD wrappers:

```javascript
// bad: bloat project code with AMD/UMD export
(function unifiedExport(e) {
  var d = ((typeof define === 'function') && define),
    m = ((typeof module === 'object') && module);
  if (d && d.amd) { d(function () { return e; }); }
  if (m && m.exports) { m.exports = e; }
}(api));

// good: 
// Just omit it. Use your toolchain to add compatibility
// mechanisms only in the dist version.
```

PR #1648, opened 2017-11-28
Add 'to' as a specialLink to the 'anchor-is-valid' a11y rule

### LINKED ISSUE #1647: ESLint 'anchor-is-valid' rule interfering with react-router-dom
```
    'jsx-a11y/anchor-is-valid': ['error', {
      components: ['Link'],
      specialLink: [],
      aspects: ['noHref', 'invalidHref', 'preferButton'],
    }],
```
The current configuration will cause `<Link to="/route" />` to be marked as invalid, as it does not have an `href` attribute. Internally, the `Link` component will transform the `to` prop and render an anchor element that does have an `href`.

`to` just needs to be added to the `specialLink` array. 

I'd be happy to open a PR if everyone is okay with this change.

kripod:
Support React Router's `<Link to="...">` attribute out of the box.

Fixes #1647.

### DIFF
--- a/packages/eslint-config-airbnb/rules/react-a11y.js
+++ b/packages/eslint-config-airbnb/rules/react-a11y.js
@@ -187,7 +187,7 @@ module.exports = {
     // https://github.com/evcohen/eslint-plugin-jsx-a11y/blob/0745af376cdc8686d85a361ce36952b1fb1ccf6e/docs/rules/anchor-is-valid.md
     'jsx-a11y/anchor-is-valid': ['error', {
       components: ['Link'],
-      specialLink: [],
+      specialLink: ['to'],
       aspects: ['noHref', 'invalidHref', 'preferButton'],
     }],
   },


ljharb:
@astorije one is certainly planned; i'm waiting on a few changes in eslint.

rdsedmundo:
Really wanting this to be released. I'm being forced to replicate the eslint's configs in all projects in order to add it as a `specialLink`.

ljharb:
The next major release.

kripod:
@ljharb When can we expect a new release? I would be grateful if there was a `@next` version available of the ESLint config npm packages.

ljharb:
We don’t publish prereleases; the next (semver-major) version will come when it’s ready :-)

rdsedmundo:
I'm commenting here because I'm one of the ones subscribed to this topic and waiting for this fix, for not having to overwrite in every project manually.

@ljharb 
It would be great to start publishing them so. If they aren't published the issues are simply not being resolved then, i.e if the code is merged on master but there's no way of getting it through NPM, then the issue is still happening on the library, i.e the issue is not resolved.

If it were a thing like weeks or 1 month it would be reasonable, but needing to wait more than 5 months to have something that has a fix already it doesn't make much sense.

ljharb:
@rdsedmundo in this case, the “fix” can be directly used in your eslintrc - you don’t have to wait for a release, so I’m not sure why it would matter if it was released in 5 minutes or 5 months.

PR #1652, opened 2017-12-04
Set import/extensions to ignorePackages

remcohaszing:
This allows to import non-JavaScript files through the main export of a dependency's package.json.

The following would trigger an error before, but is fine with the new configuration:

```js
import 'roboto-fontface';
```


remcohaszing:
I have rebased onto the latest master, in which `eslint-plugin-import` has already been bumped to `^2.9.0`.

anshgoel99:
It’s a nice one

PR #1662, opened 2017-12-16
Fix typo

tbroadley:
PR created using https://github.com/tbroadley/github-spellcheck-cli.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1754,8 +1754,8 @@ Other Style Guides
     }
     ```
 
-  <a name="hoisting--named-expresions"></a><a name="14.3"></a>
-  - [14.3](#hoisting--named-expresions) Named function expressions hoist the variable name, not the function name or the function body.
+  <a name="hoisting--named-expresions"></a><a name="hoisting--named-expressions"></a><a name="14.3"></a>
+  - [14.3](#hoisting--named-expressions) Named function expressions hoist the variable name, not the function name or the function body.
 
     ```javascript
     function example() {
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -1754,8 +1754,8 @@ Other Style Guides
     }
     ```
 
-  <a name="hoisting--named-expresions"></a><a name="14.3"></a>
-  - [14.3](#hoisting--named-expresions) Named function expressions hoist the variable name, not the function name or the function body.
+  <a name="hoisting--named-expressions"></a><a name="14.3"></a>
```
ljharb: while your spelling is indeed correct, links may never change, so this will have to be misspelled forever.

tbroadley: That makes sense. Maybe there's some way to change the link while preserving backwards compatibility with existing links. How about adding an empty anchor with the correct spelling, then fixing the spelling here? Something like:

```md
<a href="hoisting--named-expresions"></a>
- [14.3](#hoisting--named-expressions)
```

I've implemented that change here and it seems to work well: https://github.com/tbroadley/javascript/blob/fix-typo/README.md

ljharb: Sure, we could definitely add an empty anchor for the existing spelling and then correct it - great suggestion! If you restore your PR branch, I'll reopen this PR :-)

tbroadley: Done!

ljharb:
Thank you!

tbroadley:
You're welcome! Thank you for pointing out the issue with my original changeset.

PR #1668, opened 2017-12-24
[eslint config][base][patch] enable option `allowImplicit` in rule `array-callback-return`

1pete:
This allows return `undefined` implicitly.
```js
array.map((item) => {
  return;
});
```


ljharb:
Linking to https://github.com/eslint/eslint/issues/8539 / https://github.com/eslint/eslint/pull/9344

PR #1669, opened 2017-12-24
Docs: HTTP -> HTTPS

him2him2:
URL was updated to use HTTPS protocol


ghost:
> URL was updated to use HTTPS protocol

PR #1682, opened 2017-12-26
issue #1673 Method chaining

### LINKED ISSUE #1673: Method chaining lets the receiver nominate a next receiver
Re https://github.com/airbnb/javascript#constructors--chaining

Code like

```js
luke.jump().setHeight(20);
```
enables luke to provide something other than luke itself as the receiver of the `setHeight` message. When this delegation of control is intended, such method chaining is great. OTOH, if the caller assumes it is equivalent to

```js
luke.jump();
luke.setHeight(20);
```

then they depend on a contract that luke might not obey. That doesn't necessarily mean the style should not be used, but the hazard needs to be explained clearly whenever this style is encouraged.

Fatehsandhu:
@ljharb @erights Can you guys let me know if this correct?
Thanks!

Fixes #1673.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1111,6 +1111,8 @@ Other Style Guides
   <a name="constructors--chaining"></a><a name="9.3"></a>
   - [9.3](#constructors--chaining) Methods can return `this` to help with method chaining.
 
+    > When? Use chaining when functions manipulate the internals of an object instead of just returning data.
+
     ```javascript
     // bad
     Jedi.prototype.jump = function () {
@@ -1142,7 +1144,7 @@ Other Style Guides
     const luke = new Jedi();
 
     luke.jump()
-      .setHeight(20);
+      .setHeight(20); // same as const tmp = luke.jump(); tmp.setHeight(20);
     ```
 
   <a name="constructors--tostring"></a><a name="9.4"></a>


erights (commented):
I like the code rewrite in the comment. Unlike the naive rewrite, it makes the semantics clear.

Regarding the first change, I'd say "Only use" rather than "Use" but I defer to @ljharb on this.

Fatehsandhu:
@erights Thanks for the prompt reply. I will wait for @lijharb as well.

README.md
```diff
@@ -1111,6 +1111,8 @@ Other Style Guides
   <a name="constructors--chaining"></a><a name="9.3"></a>
   - [9.3](#constructors--chaining) Methods can return `this` to help with method chaining.
 
+    > When? Use chaining when functions manipulate the internals of an object instead of just returning data.
```
ljharb: Definitely not "only use", because (for example) all the non-mutating array methods are perfect chaining candidates but do not manipulate any internals.

In general, this section may need to be reworked to suggest making classes immutable (ie, to always return a new instance, allowing chaining and also avoiding mutation). However that shouldn't go in this PR.

README.md
```diff
@@ -1142,7 +1144,8 @@ Other Style Guides
     const luke = new Jedi();
 
     luke.jump()
-      .setHeight(20);
+      .setHeight(20); // same as var tmp = luke.jump(); tmp.setHeight(20);
+    // executes left to right. First jump() and then setHeight()
```
ljharb: is there any reason to believe this comment isn't self-evident?

Fatehsandhu: @ljharb I agree with you that line 1148 can be removed. Is that okay?

erights:
I'm ok with removing the comment. The reason I liked it is as an indirect reminder that it does not mean:

```js
luke.jump();
luke.setHeight(20);
```
but it's probably better just to be explicit about this.

README.md
```diff
@@ -1142,7 +1144,7 @@ Other Style Guides
     const luke = new Jedi();
 
     luke.jump()
-      .setHeight(20);
+      .setHeight(20); // same as var tmp = luke.jump(); tmp.setHeight(20);
```
gergoerdosi: If this comment stays then it should use `const` instead of `var`.

gergoerdosi (changes requested):

Fatehsandhu:
@ljharb @erights Any other changes you guys think are appropriate?

README.md
```diff
@@ -1177,11 +1144,11 @@ Other Style Guides
     const luke = new Jedi();
 
     luke.jump()
-      .setHeight(20);
+      .setHeight(20); // same as const tmp = luke.jump(); tmp.setHeight(20);
```
ljharb, line 1147: i'm not sure this adds much value.

PR #1685, opened 2017-12-31
Update license year

m1guelpf:
TO DO:

- [x] Wait to the next year (100% done, depends on timezone)
- [ ] Merge!

:octocat:

### DIFF
--- a/LICENSE.md
+++ b/LICENSE.md
@@ -1,6 +1,6 @@
 MIT License
 
-Copyright (c) 2017 Airbnb
+Copyright (c) 2012 Airbnb
 
 Permission is hereby granted, free of charge, to any person obtaining a copy
 of this software and associated documentation files (the "Software"), to deal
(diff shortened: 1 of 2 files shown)


LICENSE.md
```diff
@@ -1,6 +1,6 @@
 MIT License
 
-Copyright (c) 2017 Airbnb
+Copyright (c) 2018 Airbnb
```
ljharb: Technically these years don’t need to be updated at all, legally; I’d prefer to just set this back to the original year the guide was created and leave it be forever.

m1guelpf: What about {year}-present?

LICENSE.md
```diff
@@ -1,6 +1,6 @@
 MIT License
 
-Copyright (c) 2017 Airbnb
+Copyright (c) 2017-present Airbnb
```
ljharb: The copyright didn't start in 2017; this should just say "2012". The " - present" is always implied and thus is redundant.

m1guelpf:
@ljharb Changed to original date

PR #1688, opened 2018-01-03
Disable prefer-destructuring for array AssignmentExpressions

kripod:
For instance, the following code snippets were flagged as erroneous before this change:

```js
outputs[0] = inputs[0]; // "Fix": [outputs[0]] = inputs;
outputs[0] = inputs[100];
```

### DIFF
--- a/packages/eslint-config-airbnb-base/rules/es6.js
+++ b/packages/eslint-config-airbnb-base/rules/es6.js
@@ -119,7 +119,7 @@ module.exports = {
         object: true,
       },
       AssignmentExpression: {
-        array: true,
+        array: false,
         object: true,
       },
     }, {


ljharb:
Doing anything with array indexes is a code smell, including assigning directly to an index in `outputs`.

This change is intentionally enabled, since the only way anything should be extracted out of an array is with destructuring, or array iteration methods.

ericblade:
... because it isn't common to need to get at the 100th element of an array without also accessing the first 99? o.O

ljharb:
@ericblade no, it's decidedly not common. An array is conceptually an atomic list; you should always be operating on every item at once. You shouldn't need the 100th item, but you might use `.find` to extract an item that has the characteristics you want, which might *happen* to be the 100th item.

ericblade:
... which is exactly the same argument I would make (had i thought of how to phrase it first :) ) that destructuring an array seems like something that would hardly ever be used, as usually I would expect to use an array to hold a large volume of things . . I'd probably expect to know that I need the Xth item in an array way more frequently (perhaps by caching the results of .findIndex) than knowing I need the first, second, and third elements of an array every time.  Either way, you're right, it's not nearly as common as needing to iterate all or potentially all of a list.

ljharb:
Totally fair. In those rare cases where you’re using an index from findIndex (for performance, presumably, as opposed to doing another `.find`) i think an eslint override comment is appropriate.

PR #1690, opened 2018-01-05
Update README.md

liujuntao123:
when an Object is created by Object.create(null), it's prototype object should be null, not itself.

### DIFF
--- a/README.md
+++ b/README.md
@@ -285,7 +285,7 @@ Other Style Guides
   <a name="objects--prototype-builtins"></a>
   - [3.7](#objects--prototype-builtins) Do not call `Object.prototype` methods directly, such as `hasOwnProperty`, `propertyIsEnumerable`, and `isPrototypeOf`.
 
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -285,7 +285,7 @@ Other Style Guides
   <a name="objects--prototype-builtins"></a>
   - [3.7](#objects--prototype-builtins) Do not call `Object.prototype` methods directly, such as `hasOwnProperty`, `propertyIsEnumerable`, and `isPrototypeOf`.
 
-    > Why? These methods may be shadowed by properties on the object in question - consider `{ hasOwnProperty: false }` - or, the object may be a null object (`Object.create(null)`).
+    > Why? These methods may be shadowed by properties on the object in question - consider `{ hasOwnProperty: false }` - or, the object's prototype may be a null object (`Object.create(null)`).
```
ljharb: `Object.create(null)` is a null object; its prototype is `null`, not a null object.

This is correct as-is.

PR #1694, opened 2018-01-10
Use npx to use install-peerdeps without actually globally installing install-peerdeps.

itaditya:
As node v8 is in LTS, many people have upgraded it to. With node8, there comes an awesome utility `npx` which allows us to run cli tools without installing them.

### DIFF
--- a/packages/eslint-config-airbnb/README.md
+++ b/packages/eslint-config-airbnb/README.md
@@ -20,7 +20,13 @@ If you use yarn, run `npm info "eslint-config-airbnb@latest" peerDependencies` t
   npm info "eslint-config-airbnb@latest" peerDependencies
   ```
 
-  Linux/OSX users can run
+  If using **npm 5+**, use this shortcut
+
+  ```sh
+  npx install-peerdeps --dev eslint-config-airbnb
+  ```
+
+  If using **npm < 5**, Linux/OSX users can run
 
   ```sh
   (
@@ -35,13 +41,12 @@ If you use yarn, run `npm info "eslint-config-airbnb@latest" peerDependencies` t
   npm install --save-dev eslint-config-airbnb eslint@^#.#.# eslint-plugin-jsx-a11y@^#.#.# eslint-plugin-import@^#.#.# eslint-plugin-react@^#.#.#
   ```
 
-  Windows users can either install all the peer dependencies manually, or use the [install-peerdeps](https://github.com/nathanhleung/install-peerdeps) cli tool.
+  If using **npm < 5**, Windows users can either install all the peer dependencies manually, or use the [install-peerdeps](https://github.com/nathanhleung/install-peerdeps) cli tool.
 
   ```sh
   npm install -g install-peerdeps
   install-peerdeps --dev eslint-config-airbnb
   ```
-
   The cli will produce and run a command like:
 
   ```sh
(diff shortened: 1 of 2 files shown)


packages/eslint-config-airbnb/README.md
```diff
@@ -40,6 +40,10 @@ If you use yarn, run `npm info "eslint-config-airbnb@latest" peerDependencies` t
   ```sh
   npm install -g install-peerdeps
   install-peerdeps --dev eslint-config-airbnb
+  
+  or if using node8, use this shortcut
```
ljharb: specifically, this is using npm 5, which just happens to be in node 8.

packages/eslint-config-airbnb/README.md
```diff
@@ -40,6 +40,10 @@ If you use yarn, run `npm info "eslint-config-airbnb@latest" peerDependencies` t
   ```sh
   npm install -g install-peerdeps
   install-peerdeps --dev eslint-config-airbnb
+
```
ljharb: let's close the highlighted section here, and start it anew below, since "or if" isn't valid shell syntax

packages/eslint-config-airbnb/README.md
```diff
@@ -41,6 +41,11 @@ If you use yarn, run `npm info "eslint-config-airbnb@latest" peerDependencies` t
   npm install -g install-peerdeps
   install-peerdeps --dev eslint-config-airbnb
   ```
+  
+  or if using node8, use this shortcut
```
ljharb: let's change "node8" to "npm 5+", since that's more accurate.

itaditya: Ok :+1:

yairEO:
When I run:

    npx install-peerdeps eslint-config-airbnb-base -g

it installs locally and ignores the `-g` flag. is that normal? 
I want to install this package globally (to use with sublime) 

Thanks

ljharb:
@yairEO yes, nothing should be installed globally, including eslint. sublime should be configured to run the local eslint.

yairEO:
why? I want global, not local. I want all the projects I'm working on to use ESLINT, but not to pollute the `package.json` of those projects because I work on clients' projects and they don't like changes to `package.json` that aren't absolutely necessary, and this is obviously only for myself and not the client (which is another developer most of the time which I'm helping with).

So you see, this solves me two issues:
1. Sublime ESLINT can only be configured with a single permanent `.eslintrc`   
2. I get to not keep separation of concerns where my global style-guide is for me only and not others, since I honestly don't care if they follow it or not, but for myself, I find it helpful, showing hints and stuff, you know. 

A reasonable use-case wouldn't you agree?

yairEO:
I do not agree, since most have no style at all. honestly, so many out there care nothing for style-guide and write a messy code and I cannot simply force my standards on them, but what I *can* do is to at least lint the code that I write on these shared projects, regardless of the errors I will see, It's better than nothing and a good reminder for best-practice coding.

Also, I sometimes want linting small script files that aren't projects (but standalone files, without `package.json`) and only want the possibility to use sublime + eslint to help a bit with linting. again for  good-practices reminder. 

I could make my own global  `.eslintrc` file, but I like yours and would like to be synced with the project by installing it globally. 

If this is illogical than I will reconsider my line of thoughts regarding linting. maybe it is something that **aught** to be per-project and never a global thing (even if all projects should follow the same rules)

PR #1698, opened 2018-01-16
4.7 layout problem, be better

ZweiZhao:
have not change any code, only add a tab.


ljharb:
What's the benefit of adding this tab?

ZweiZhao:
Markdown rules that the contents of the sequence should have a tab indentation.
README.md other parts have tab indentation, only there is not.

ljharb:
Hmm. Is there some kind of linting tool we could use that would prevent this from regressing in the future?

ZweiZhao:
Sadly, no.
Just a layout problem, not a grammar issue.

PR #1699, opened 2018-01-16
fix example to meet standard listingId v listingID

BenFortner:
fixed code example to meet the standard. `listingId` changed to `listingID` 
See - Naming Conventions: Acronyms and initialisms should always be all capitalized, or all lowercased.


BenFortner:
No problem! Thank you all for the standards!

PR #1707, opened 2018-01-25
Ignore JSX curly brace presence for props to improve CSS-in-JS support

kripod:
This fix should only be in place until a better solution is found. _(See https://github.com/yannickcr/eslint-plugin-react/issues/1592)_


ljharb:
I don't think this is an appropriate fix; the examples in that thread aren't valid in React (I assume you're using something else to transform those).

The better approach would be to submit a PR to eslint-plugin-react that fixes that bug.

PR #1708, opened 2018-01-26
Use `acc` as `reduce` accumulator instead of `memo` to make it valid

astorije:
In [this part of the `no-param-reassign` configuration](https://github.com/airbnb/javascript/blob/53b2d7d245ba4abefc0429bfda4a46f099b9ace5/packages/eslint-config-airbnb-base/rules/best-practices.js#L174-L175), `acc` is allowed to be mutated (rule [7.12](https://github.com/airbnb/javascript#functions--mutate-params), but `memo` is not.

This causes the [rule 4.6 example](https://github.com/airbnb/javascript#arrays--callback-return) in the README to actually be invalid.

**Alternative proposal: Add `memo` to the whitelist along with `acc`, as it is a very common accumulator variable name for `reduce`.**

### DIFF
--- a/README.md
+++ b/README.md
@@ -404,16 +404,16 @@ Other Style Guides
     // good
     [1, 2, 3].map(x => x + 1);
 
-    // bad - no returned value means `memo` becomes undefined after the first iteration
-    [[0, 1], [2, 3], [4, 5]].reduce((memo, item, index) => {
-      const flatten = memo.concat(item);
-      memo[index] = flatten;
+    // bad - no returned value means `acc` becomes undefined after the first iteration
+    [[0, 1], [2, 3], [4, 5]].reduce((acc, item, index) => {
+      const flatten = acc.concat(item);
+      acc[index] = flatten;
     });
 
     // good
-    [[0, 1], [2, 3], [4, 5]].reduce((memo, item, index) => {
-      const flatten = memo.concat(item);
-      memo[index] = flatten;
+    [[0, 1], [2, 3], [4, 5]].reduce((acc, item, index) => {
+      const flatten = acc.concat(item);
+      acc[index] = flatten;
       return flatten;
     });
 


astorije:
Hey @ljharb, I'm wondering if I didn't explain myself poorly, because it seems to me that you misunderstood my intent in this PR, so let me try again 🙂

I agree with you that keeping `acc` is probably best. In fact, we are using that in our project, so there is nothing we need to configure on our end. I was just suggesting that to comply with the example in your README, not our project :)

When it comes to this PR however, if you look at the diff, I'm not making the `bad` example pass: it still fails because my change doesn't affect application of rule 4.6 (`array-callback-return`). I am simply fixing the documentation.

Consider the `good` case, before/after these changes:

**Before**, this example actually causes a linter failure because, even though it complies with rule 4.6, `memo` is being mutated:

```js
// good
[[0, 1], [2, 3], [4, 5]].reduce((memo, item, index) => {
  const flatten = memo.concat(item);
  memo[index] = flatten;
  return flatten;
});
```

**After**, it passes because mutation on `acc` is allowed:

```js
// good
[[0, 1], [2, 3], [4, 5]].reduce((acc, item, index) => {
  const flatten = acc.concat(item);
  acc[index] = flatten;
  return flatten;
});
```

If you look at the `bad` case before/after, it's the same thing: after these changes, the `bad` case still fails because `acc` is not returned, which is indeed against rule 4.6.

Sure, if we take this example in isolation, in a situation where `no-param-reassign` is not enforced, then yes, the `good` case of this example passes, but it's confusing that this `good` case doesn't pass the linter with the config of this guide.
Does that clarify the intent? Would you mind reconsidering this PR?

ljharb:
Thanks for clarifying :-) Sounds reasonable.

astorije:
While I'm/you're here, would you be open to a PR that adds a note or an example [here](https://github.com/airbnb/javascript#functions--mutate-params) about allowed exceptions such as `acc` as accumulator variable for `reduce`?

ljharb:
I'd prefer not mentioning the exemptions in the guide.

astorije:
In this case, it would be great if there was a rule to force the first variable of the callback of `reduce` to be called `acc`. I don't believe ESLint supports this.

I have to admit, I was confused when I saw that `memo.push()` was forbidden, then digged into the code of the package before realizing `acc` was whitelisted but not `memo` :)

PR #1710, opened 2018-01-26
[eslint config] [eslint config base] [guide] Prevent line breaks before and after `=`

sharmilajesupaul:
I'm wasn't too sure where to place this addition in our guide, so I added it to the variables section. Let me know if you think it belongs elsewhere. 

----

Guide update:

- [13.7](#variables--linebreak) Avoid linebreaks before or after `=` in an assignment. If your assignment violates [`max-len`](https://eslint.org/docs/rules/max-len.html), surround the value in parens. eslint [`operator-linebreak`](https://eslint.org/docs/rules/operator-linebreak.html).

> Why? Linebreaks surrounding `=` can obfuscate the value of an assignment.

```jsx
    // bad
    const foo =
      superLongLongLongLongLongLongLongLongFunctionName();

    // bad
    const foo
      = 'superLongLongLongLongLongLongLongLongString';

    // good
    const foo = (
      superLongLongLongLongLongLongLongLongFunctionName()
    );

    // good
    const foo = 'superLongLongLongLongLongLongLongLongString';
```

### DIFF
--- a/README.md
+++ b/README.md
@@ -1699,6 +1699,28 @@ Other Style Guides
     const truthyCount = array.filter(Boolean).length;
     ```
 
+  - [13.7](#variables--linebreak) Avoid linebreaks before or after `=` in an assignment. If your assignment violates [`max-len`](https://eslint.org/docs/rules/max-len.html), surround the value in parens. eslint [`operator-linebreak`](https://eslint.org/docs/rules/operator-linebreak.html).
+
+    > Why? Linebreaks surrounding `=` can obfuscate the value of an assignment.
+
+    ```javascript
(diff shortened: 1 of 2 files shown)


README.md
```diff
@@ -1698,6 +1698,28 @@ Other Style Guides
     const sum = array.reduce((a, b) => a + b, 0);
     const truthyCount = array.filter(Boolean).length;
     ```
+  <a name="variables--linebreak"></a><a name="13.7"></a>
```
ljharb: new sections don't get numbered links; only legacy ones do

PR #1722, opened 2018-02-04
Suggest using WeakMaps when hidden properties are needed

not-an-aardvark:
This updates the "trailing or leading underscores" guideline to suggest an better way to make properties hidden, as an alternative to just removing the underscore and making the property public.

This is motivated by a discussion with @ljharb today in the `eslint/eslint` Gitter channel.

### DIFF
--- a/README.md
+++ b/README.md
@@ -2993,6 +2993,11 @@ Other Style Guides
 
     // good
     this.firstName = 'Panda';
+
+    // good, in environments where WeakMaps are available
+    // see https://kangax.github.io/compat-table/es6/#test-WeakMap
+    const firstNames = new WeakMap();
+    firstNames.set(this, 'Panda');
     ```
 
   <a name="naming--self-this"></a><a name="22.5"></a>


README.md
```diff
@@ -2993,6 +2993,10 @@ Other Style Guides
 
     // good
     this.firstName = 'Panda';
+    
+    // good
```
ljharb: i think this is useful, but we should include a note warning about browser support that links to http://kangax.github.io/compat-table/es6/#test-WeakMap

ericblade:
May I use this topic as a learning experience? :-)

Why WeakMap, over any other particular thing?

README.md
not-an-aardvark: Updated, let me know if that was what you had in mind.

not-an-aardvark:
@ericblade Using a `WeakMap` is useful for two reasons:

* If you don't pass the `WeakMap` instance around to consumers of your object, then they can't access the value in the `WeakMap` associated with your object. This means that you can effectively store private properties on the object which are only accessible to your code.
* If the consumers of your object stop it using the object and it gets garbage-collected, the private properties will also be garbage-collected.

ericblade:
.... OK, it took me a minute of thinking this through, perhaps a more in-depth explanation of the situation, or slightly more in-depth code example would be helpful  (or perhaps I missed it, just reading that specific updated section)..

In this particular case, you have a

````
const privateData = new WeakMap();

class thingWithPrivateData {
....
}
````

.. and the WeakMap is a Map, so that multiple instances of the class can use the same data store for all of them, using "this" as the key.  That was the part that wasn't obviously clear to me -- at a quick glance, I failed to notice that we were talking about class instances, rather than like a module, in which case, basically any data structure would also work (so long as it wasn't containing references to external objects)

PR #1726, opened 2018-02-08
[testing] Add section: "testing is documentation"

drpicox:
It is great to suggest people to write tests, it is really a good idea, but I think that it is not enough.

From my experience working with other coders I have discovered that they consider hard to write tests. Not because it is difficult (you just use the test API like any other library), but because they think that tests are pointless and a waste of time.

I have done a lot of reviews of PR, including those containing tests, and I have discovered very bad testing practices which usually occurs when people are forced to write tests (usually big companies and some test obsessed people). The tests resulting of these practices are almost useless (usually testing code copies production code) and lacks of many of the expected properties of a test.

> But: program testing can be a very effective way to show the presence of bugs, but is hopelessly inadequate for showing their absence. — Edsger W. Dijkstra 1972

So tests objective should not be the only one of creating a system bugs free. It is an impossible task from the point of view of that test can effectively achieve, but there are many other objectives which are very valuables.

Main tests objective should be to make _legacy code_ disappear.

I consider _legacy code_, not just old unsupported code, but any code difficult to change because we may break our system. And as I far as I know it happens in two different scenarios:

1. We want to change some code, but we cannot guarantee that we will break some functionality because we need also to refactor the test, or there is not test,
2. We want to change or add some code, but we cannot guarantee to break any functionality because we do not know which are all supported functionalities of our system,

Both points can be summarized in good maintainability and good documentation, and we can achieve both with well written tests.

And here comes the good news, there is a simple trick to achieve these two properties: "testing is documentation".

When coders forgot about it, coders usually focus in testing other points. They being obsessed in test function by function all possible cases, all input combinations, all ranges, they try to get the 100% code coverage regardless its usefulness. But all this effort just creates a rigid codes and poorly covered. I think that you tried to make people think better about it when you said: "Be cautious about stubs and mocks", but I it is not enough.

An example of 100% code coverage and almost useless testing code:

```javascript
describe('addListener', () => {
  it('should add a callback to the queue', () => {
    dispatcher.addListener(cb);
    expect(dispatcher.queue).toContain(cb);
  });
});

describe('deliver', () => {
  it('should invoke queue callbacks with the received argument', () => {
    dispatcher.queue.push(cb);
    dispatcher.deliver('message');
    expect(cb).toHaveBeenCalledWith('message');
  });
});
```

These tests achieve 100% of code coverage, even if you suppress expectations in both tests, or replace them by checking anything else, like they both methods return undefined.

If you look carefully these tests tells almost nothing about what to do with the dispatcher, it even exposes some data structures that meant to be internal ( we could implement them with:  https://github.com/airbnb/javascript#naming--leading-underscore ). It does not matter how many other test you add (what happens if the queue is empty, if there are many callbacks, if there is no message, ...), you will have hard time trying understand what dispatcher and it will be almost impossible to refactor it without break all tests first (refactor usually implies changing internal representations).

Now check this other test:

```javascript
it('should deliver messages to listeners', () => {
  dispatcher.addListener(cb);
  dispatcher.deliver('message');

  expect(cb).toHaveBeenCalledWith('message')
});
```

It is simple, easy to understand, does not use internal representation, and it has 100% of code coverage.

It should be our objective, and if you look carefully, it is a wonderful documentation. You understand involved steps and see the whole picture. If you want can do copypaste quickly, and build new code from here. And of course, you can refactor the dispatcher code as much as you want without changing a simple line of a test as long as you do not break the dispatcher api compatibility.

Because all exposed here that is the reason why I suggest the addition this section, just to help to create better and more effective testing code. It will mitigate the frustration of many programmers that do not know how to start writing a test and it will give them a direction to follow.

With these few indications I hope that coders will feel good with their own test and give them some deeper meaning that helps them to be proud of their work.

---

Note: I was tempted to add something about AAA (Arrange Act Assert) or GWT (Give When Then), and force people to split tests that test something more than one thing, but I think that as far as they understand that each test case is an example of usage for the documentation it should not be necessary.

### DIFF
--- a/README.md
+++ b/README.md
@@ -3403,6 +3403,16 @@ Other Style Guides
     - We primarily use [`mocha`](https://www.npmjs.com/package/mocha) at Airbnb. [`tape`](https://www.npmjs.com/package/tape) is also used occasionally for small, separate modules.
     - 100% test coverage is a good goal to strive for, even if it’s not always practical to reach it.
     - Whenever you fix a bug, _write a regression test_. A bug fixed without a regression test is almost certainly going to break again in the future.
+    
+  <a name="testing--is-documentation"></a>
+  - [30.3](#testing--is-documentation) **Is your documentation**:
+    - Tests should explain what your code does.
+    - First write a case for each main functionality, and then their exceptions.
+    - Refactor tests until they are easy to read and understand.
+    - Never use _"private"_ properties to test code.
+    - Code inside tests should be copy-pasteable code to make new developments.
+    - If some functionality works but is not tested, do not use it. It may break your code in the future.
+    - Test functionalities, not functions.
 
 **[⬆ back to top](#table-of-contents)**
 


README.md
```diff
@@ -3390,6 +3390,16 @@ Other Style Guides
     - We primarily use [`mocha`](https://www.npmjs.com/package/mocha) at Airbnb. [`tape`](https://www.npmjs.com/package/tape) is also used occasionally for small, separate modules.
     - 100% test coverage is a good goal to strive for, even if it’s not always practical to reach it.
     - Whenever you fix a bug, _write a regression test_. A bug fixed without a regression test is almost certainly going to break again in the future.
+    
+  <a name="testing--is-documentation"></a><a name="30.3"></a>
```
ljharb: please remove this 30.3 link - new numbered sections should not be added; they're only here for legacy reasons.

README.md
```diff
@@ -3390,6 +3390,16 @@ Other Style Guides
...
+  - [30.3](#testing--is-documentation) **Is your documentation**:
+    - Tests should explain what your code does.
+    - First write a case for each main functionality, and then their exceptions.
+    - Refactor tests until they are easy to read and understand.
+    - Never use object internal representation from test code, use objects as other code will.
+    - Tests should be copy-paste code to make new developments.
```
ljharb: can you elaborate on what this means?

README.md
```diff
@@ -3390,6 +3390,16 @@ Other Style Guides
     - We primarily use [`mocha`](https://www.npmjs.com/package/mocha) at Airbnb. [`tape`](https://www.npmjs.com/package/tape) is also used occasionally for small, separate modules.
     - 100% test coverage is a good goal to strive for, even if it’s not always practical to reach it.
     - Whenever you fix a bug, _write a regression test_. A bug fixed without a regression test is almost certainly going to break again in the future.
+    
+  <a name="testing--is-documentation"></a><a name="30.3"></a>
+  - [30.3](#testing--is-documentation) **Is your documentation**:
+    - Tests should explain what your code does.
+    - First write a case for each main functionality, and then their exceptions.
+    - Refactor tests until they are easy to read and understand.
+    - Never use object internal representation from test code, use objects as other code will.
```
ljharb: Can you elaborate on this one?

ljharb (changes requested):

README.md
drpicox: Ok!

README.md
drpicox: I rewrite it.

README.md
drpicox: I rewrite it.

drpicox: > Never use _"private"_ properties to test code.

yes... I elaborate making it shorter O: ) 
I do not want to break the _"rhythm"_ of the all testing section.

I think that it is enough telling to anyone do not use "private" stuff, because they already know what means "private". And if not, and they search for "private" they will find more things later.
--


"_ of all testing section making something large
I do not want to break the _"rythm of all testing section making something largeI do not want to break the _"rythm"_ of all testing section making something large

README.md
drpicox: Ops! Nice found, copypaste ≠ copy-pasteable, it seems that I force people to make copy pastes from here, which it is not my objective.

> Code inside tests should be copy-pasteable code to make new developments.

README.md
```diff
@@ -3403,6 +3403,16 @@ Other Style Guides
     - We primarily use [`mocha`](https://www.npmjs.com/package/mocha) at Airbnb. [`tape`](https://www.npmjs.com/package/tape) is also used occasionally for small, separate modules.
     - 100% test coverage is a good goal to strive for, even if it’s not always practical to reach it.
     - Whenever you fix a bug, _write a regression test_. A bug fixed without a regression test is almost certainly going to break again in the future.
+    
+  <a name="testing--is-documentation"></a>
+  - [30.3](#testing--is-documentation) **Is your documentation**:
+    - Tests should explain what your code does.
+    - First write a case for each main functionality, and then their exceptions.
+    - Refactor tests until they are easy to read and understand.
+    - Never use _"private"_ properties to test code.
```
ljharb: i'm not sure this really belongs; if something is truly private, it's impossible to test directly; and if it's possible to test directly, it's not actually private.

README.md
```diff
@@ -3403,6 +3403,16 @@ Other Style Guides
...
+    - Code inside tests should be copy-pasteable code to make new developments.
```
ljharb: i'm afraid i'm still not clear on what this one means :-/

README.md
```diff
@@ -3403,6 +3403,16 @@ Other Style Guides
...
+    - If some functionality works but is not tested, do not use it. It may break your code in the future.
+    - Test functionalities, not functions.
```
ljharb: Can you elaborate more on what this means?

README.md
```diff
@@ -3403,6 +3403,16 @@ Other Style Guides
...
     - Whenever you fix a bug, _write a regression test_. A bug fixed without a regression test is almost certainly going to break again in the future.
+    
+  <a name="testing--is-documentation"></a>
+  - [30.3](#testing--is-documentation) **Is your documentation**:
+    - Tests should explain what your code does.
+    - First write a case for each main functionality, and then their exceptions.
+    - Refactor tests until they are easy to read and understand.
+    - Never use _"private"_ properties to test code.
+    - Code inside tests should be copy-pasteable code to make new developments.
+    - If some functionality works but is not tested, do not use it. It may break your code in the future.
```
ljharb: I'd say if it's not tested, then it objectively doesn't work, because you don't have a test proving it does :-)

README.md
drpicox: Uhmmm, I understand your point, I guess that it comes from here:
- https://github.com/airbnb/javascript#naming--leading-underscore
> This convention might lead developers to wrongly think that a change won’t count as breaking, or that tests aren’t needed.

What about changing it for the following statement:

    - JavaScript does not have the concept of privacy in terms of properties or methods,
      if you want something to be “private”, it must not be observably present,
     and they won't be used in tests, see [here](#naming--leading-underscore) for details.

README.md
drpicox: The idea is the following: because testing code is documentation, when you start writing new code, you have a very high probability of having that code already implemented in a test. So when someone starts coding and needs to use an existing code, he just have to go the specs, and make copy-paste. 
I liked the copy-pasteable expression because it implies that the testing code should be as close as possible to the production code, so each test initialisation, operation and check should be closer to other code initialisation, operation, and querying.
Although this idea, may be it should be clearer something like:

    - When someone starts writing code will 
       tests are the first place to know how to use existing code,
       make them easy to reuse code

README.md
drpicox: Uhmm, yeah!

What I have found that some times there are some behaviours that emerge from the current code structure, for example:

```javascript
getCar(carIndex) {
   return carList[carIndex];
}
```

It turns out that if you call to `getCar(undefined)` it returns undefined.
Some people think that to return undefined if argument undefined it is a desired functionality, 
and they start to put inputs to `getCar` that may have an undefined.

Some time later, you may decide to do something else with carIndex that will break if it is undefined. 
And then you will break previous code.
And... if they did a lot of mocking, you will not discover it until it is too late (or difficult to trace).

In this scenario coders can say that the function is tested. But if there was no case in which it returns undefined, they never should expected and undefined.

Of course you can add 1.000 cases for each combination, but 1: it is not feasible, 2: when you define an API you define which cases are accepted. 

So at the end, specially you should look test code always before using something in some way that was not designed for.

What about:

    - If it is not tested, than it objectively does not work, 
      because you do not have a test proving it does.
      Including those cases that are not present.

drpicox:
I want to review comments, give me some days to review them.
(although any comment is welcome meanwhile)

README.md
ljharb: I think this is kind of self-evident; and would be more clear with a concrete example of what to avoid.

README.md
ljharb: I'm still not clear on this. Are you saying that individual tests should be good starting points for adding new tests? I'm not sure this is even broadly true.

picoluna1969:
Cool

picoluna1969:
Cool

README.md
ljharb: Sounds good to me.

PR #1732, opened 2018-02-15
Allow devDependencies for foo_spec.js naming style

lencioni:
Some projects name test files like `foo_spec.js` instead of `foo.spec.js`.


fredbuluku:
Thanks

PR #1733, opened 2018-02-18
Line breaks for multi-entry array

jfdnc:
Description states, "Use line breaks after open and before close array brackets if an array has multiple lines," which would imply the form reflected in the edit. If the intention was actually to list the array as 'const arr = [[0, 1], [2, 3], [4, 5]];' a caveat should be added to 4.7 description, since it doesn't involve a line break after the opening or before the closing brackets.

### DIFF
--- a/README.md
+++ b/README.md
@@ -460,7 +460,11 @@ Other Style Guides
     ];
 
     // good
-    const arr = [[0, 1], [2, 3], [4, 5]];
+    const arr = [
+      [0, 1], 
+      [2, 3], 
+      [4, 5],
+    ];
 
     const objectInArray = [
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -460,7 +460,11 @@ Other Style Guides
     ];
 
     // good
-    const arr = [[0, 1], [2, 3], [4, 5]];
+    const arr = [
```
ljharb: this is intentional; this array *doesn't* have multiple lines, so there's no need to use line breaks - that's why it's a "good" example.

ljharb (changes requested):

PR #1737, opened 2018-02-19
Add react/no-this-in-sfc

caub:


ljharb:
Thanks; this is a change I already have locally as I prepare for the next major bump. I'll leave this open in the interim.

PR #1740, opened 2018-02-20
[guide] [react] add comment about what JS standards are followed

raunofreiberg:
Seen a few pull requests for the React style guide that try to introduce features that aren't yet in stage >= 3 (static class fields for i.e), so I've decided to add a comment to the beginning of the guide. Not sure if this is the right place to put it, but here it is.

This disclaimer might help:
1) give the community an idea what standards this guide is based on
2) prevent redundant pull requests

### DIFF
--- a/react/README.md
+++ b/react/README.md
@@ -2,6 +2,8 @@
 
 *A mostly reasonable approach to React and JSX*
 
+This style guide is mostly based on the standards that are currently prevalent in JavaScript, although some conventions (i.e async/await or static class fields) may still be included or prohibited on a case-by-case basis. Currently, anything prior to stage 3 is not included nor recommended in this guide.
+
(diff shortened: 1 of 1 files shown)


react/README.md
```diff
@@ -2,6 +2,9 @@
 
 *A mostly reasonable approach to React and JSX*
 
+This style guide is based on the standards that are currently prevalent in JavaScript or proposals in stage 3 or greater.
```
ljharb: Let's qualify this by saying it's case-by-case; async/await is still prohibited by this guide which is stage 4, as are class instance fields which are stage 3.

raunofreiberg: Fair point. I've went ahead and improved the general idea of this comment by amending the last commit. Let me know how it looks like now.

PR #1742, opened 2018-02-21
Add an anchor for rule 13.7

aifreedom:
As the title. It fixes the anchor link.

@ljharb

### DIFF
--- a/README.md
+++ b/README.md
@@ -1701,6 +1701,7 @@ Other Style Guides
     const truthyCount = array.filter(Boolean).length;
     ```
 
+<a name="variables--linebreak"></a>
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -1701,6 +1701,7 @@ Other Style Guides
     const truthyCount = array.filter(Boolean).length;
     ```
 
+<a name="variables--linebreak"></a><a name="13.7"></a>
```
ljharb: The first one of these would be great; but we don't add new numbered links, they're only there for legacy reasons.

aifreedom:
@ljharb removed the legacy one. PTAL

PR #1743, opened 2018-02-22
enable no useless path segments

graingert:


ljharb:
Thanks; dependency updates (and the resulting rules they bring in) are something I need to do myself; I'll adapt this PR to what's needed.

Issue #1744, opened 2018-02-22
The npm packages get distributed without licenses


honzajavorek:
There is a root license file, but the individual npm packages get distributed without full license text. In such case whatever is uploaded to npm is in fact not licensed, as the license itself mentions it needs to be attached:

> The above copyright notice and this permission notice shall be included in all copies or substantial portions of the Software.

Let me know if I can help with fixing the issue.

PR #1745, opened 2018-02-22
Ensure licenses in sub packages

honzajavorek:
Fix https://github.com/airbnb/javascript/issues/1744. Not sure how to trigger patch release, couldn't find any contributing docs.


honzajavorek:
This won't work, npm apparently doesn't follow symlinks when packing the package.

PR #1746, opened 2018-02-22
Add licenses to sub packages

honzajavorek:
Fix https://github.com/airbnb/javascript/issues/1744. Not sure how to trigger patch release, couldn't find any contributing docs. Please advise.

Replaces #1745

### DIFF
--- a/packages/eslint-config-airbnb-base/LICENSE.md
+++ b/packages/eslint-config-airbnb-base/LICENSE.md
@@ -1,0 +1,21 @@
+MIT License
+
+Copyright (c) 2012 Airbnb
+
+Permission is hereby granted, free of charge, to any person obtaining a copy
+of this software and associated documentation files (the "Software"), to deal
+in the Software without restriction, including without limitation the rights
+to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
+copies of the Software, and to permit persons to whom the Software is
+furnished to do so, subject to the following conditions:
+
+The above copyright notice and this permission notice shall be included in all
+copies or substantial portions of the Software.
+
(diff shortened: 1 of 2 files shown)


packages/eslint-config-airbnb-base/LICENSE.md
```diff
@@ -0,0 +1,21 @@
+MIT License
+
+Copyright (c) 2012 Airbnb
```
raunofreiberg: 2012?

packages/eslint-config-airbnb/LICENSE.md
```diff
@@ -0,0 +1,21 @@
+MIT License
+
+Copyright (c) 2012 Airbnb
```
raunofreiberg: Ditto.

honzajavorek:
@ljharb Thank you! While I'm no lawyer myself, I think although the package is a product of the repository with a license text, it is still necessary to accompany the package with a full license text, as it is distributed separately. And the MIT license requires the full license text to be alongside the code. Will link this as well: https://github.com/babel/babel/pull/7308#issuecomment-367615152

honzajavorek:
@ljharb btw, do you plan a new release?

ljharb:
The next release will be breaking (due to other merged changes); I’m not going to backport a patch release just for this.

ljharb:
Separately, the point of a SPDX-compliant identifier, as i understand it, is that the license text is no longer relevant or needed because it is uniquely identified by the license identifier (which implies license text). Maybe @kemitchell can confirm, or correct me if I’m wrong?

honzajavorek:
> I’m not going to backport a patch release just for this.

I didn't think of it like that. It was more like whether you have an idea when the next release might happen.

ljharb:
No idea, but hopefully soon.

PR #1747, opened 2018-02-24
Arrow function  will make it better

apoorvtomar2222:


README.md
```diff
@@ -212,6 +212,15 @@ Other Style Guides
         return atom.value + value;
       },
     };
+    
+    //even better
```
ljharb: For one, this isn’t actually better; but for another, this isn’t using concise method shorthand, which is the purpose of this section.

PR #1749, opened 2018-02-27
Adds eslint config extends with only whitespace rules enabled

sharmilajesupaul:
Adds a change to pull in the eslint config from the project root, turn off all rules except the whitespace rules explicitly listed in the array.

This list is hardcoded for now, if there are any that I'm missing please let me know: 
```jsx
[
  'array-bracket-newline',
  'array-bracket-spacing',
  'array-element-newline',
  'arrow-spacing',
  'block-spacing',
  'comma-spacing',
  'computed-property-spacing',
  'dot-location',
  'eol-last',
  'func-call-spacing',
  'function-paren-newline',
  'generator-star-spacing',
  'implicit-arrow-linebreak',
  'indent',
  'key-spacing',
  'keyword-spacing',
  'line-comment-position',
  'linebreak-style',
  'multiline-ternary',
  'newline-per-chained-call',
  'no-irregular-whitespace',
  'no-mixed-spaces-and-tabs',
  'no-multi-spaces',
  'no-regex-spaces',
  'no-spaced-func',
  'no-trailing-spaces',
  'no-whitespace-before-property',
  'nonblock-statement-body-position',
  'object-curly-newline',
  'object-curly-spacing',
  'object-property-newline',
  'one-var-declaration-per-line',
  'operator-linebreak',
  'padded-blocks',
  'padding-line-between-statements',
  'rest-spread-spacing',
  'semi-spacing',
  'semi-style',
  'space-before-blocks',
  'space-before-function-paren',
  'space-in-parens',
  'space-infix-ops',
  'space-unary-ops',
  'spaced-comment',
  'switch-colon-spacing',
  'template-tag-spacing',
  'import/newline-after-import',
]
```

### DIFF
--- a/packages/eslint-config-airbnb-base/whitespace.js
+++ b/packages/eslint-config-airbnb-base/whitespace.js
@@ -1,0 +1,73 @@
+const baseConfig = require('.');
+const entries = require('object.entries');
+const { CLIEngine } = require('eslint');
+
+function onlyErrorOnRules(rulesToError, config) {
+  const errorsOnly = { ...config };
+  const cli = new CLIEngine({ baseConfig: config, useEslintrc: false });
+  const baseRules = cli.getConfigForFile('./').rules;
+
+  entries(baseRules).forEach(([key, value]) => {
+    if (rulesToError.indexOf(key) === -1) {
+      if (Array.isArray(value)) {
+        errorsOnly.rules[key] = ['warn'].concat(value.slice(1));
+      } else if (typeof value === 'number') {
+        errorsOnly.rules[key] = 1;
+      } else {
+        errorsOnly.rules[key] = 'warn';
+      }
+    }
+  });
+
+  return errorsOnly;
+}
+
+module.exports = onlyErrorOnRules([
+  'array-bracket-newline',
+  'array-bracket-spacing',
+  'array-element-newline',
+  'arrow-spacing',
+  'block-spacing',
+  'comma-spacing',
+  'computed-property-spacing',
+  'dot-location',
+  'eol-last',
+  'func-call-spacing',
+  'function-paren-newline',
+  'generator-star-spacing',
+  'implicit-arrow-linebreak',
+  'indent',
+  'key-spacing',
+  'keyword-spacing',
+  'line-comment-position',
+  'linebreak-style',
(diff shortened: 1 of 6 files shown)


packages/eslint-config-airbnb/whitespace.js
```diff
@@ -0,0 +1 @@
+require('../eslint-config-airbnb-base/whitespace');
```
ljharb: I think this will need extra react-specific whitespace rules?

sharmilajesupaul:
added info to readme @ljharb, here's what it looks like: 

### eslint-config-airbnb-base/whitespace

This entry point that only warns on whitespace rules and sets all other rules to warnings. View the list of whitespace rules [here](https://github.com/airbnb/javascript/blob/master/packages/eslint-config-airbnb-base/whitespace.js).

packages/eslint-config-airbnb-base/whitespace.js
```diff
@@ -0,0 +1,73 @@
+const baseConfig = require('.');
+const entries = require('object.entries');
+const { CLIEngine } = require('eslint');
+
+function onlyErrorOnRules(rulesToError, config) {
+  const errorsOnly = { ...config };
```
1pete: This object spread property in nodejs is introduced at version 8+. If this file is meant to be run with nodejs 8+, then we don't need to depend on more lib e.g. `object.entries`, `Object.entries` already exists there.

I can make a PR to clean this. What do you think, @ljharb?

1pete: Even if we convert this file to be compatible with nodejs 4 which is declared in package.json, we still don't need this extra dependency, since eslint already depends on lodash, then we can add it too and use _.forEach.

ljharb: Whoops, this should use `object.assign` instead of spread.

There’s no cost to an extra dev dep, and I’d prefer not to use lodash.

1pete: I see, then we need to change array destructuring at line 10 too.

I'll make a PR, ok?

ljharb: Sure.

PR #1751, opened 2018-03-01
ensure `whitespace` entry point  compatible with node 4

1pete:


packages/eslint-config-airbnb-base/whitespace.js
```diff
@@ -1,20 +1,24 @@
 const baseConfig = require('.');
+const assign = require('object.assign');
 const entries = require('object.entries');
 const { CLIEngine } = require('eslint');
```
ljharb: turns out this destructuring also doesn't work on node 4; i'll update your PR.

ljharb:
I've also added some tests (failing on master [here](https://travis-ci.org/ljharb/javascript/builds/347634969)) to prove that your changes fix the problem.

Thanks! Once this passes, I'll merge it.

PR #1757, opened 2018-03-07
Turkish translation

eraycetinay:
I started to translate javascript style guide to turkish language.


eraycetinay:
I just created a pull request to draw attention to the issue and in order to avoid the translation conflict with other translators.

PR #1770, opened 2018-03-23
Change import/no-self-import from "off" to "error"

lencioni:
We recently had an incident where this rule would have prevented
infinite loops in Hypernova and in the browser.


gdborton:
Any idea why this was left off with a todo? Seems intentional.

lencioni:
Not sure. It was added here but the commit message gives us no insight: https://github.com/airbnb/javascript/commit/1fbeabd0c4b5d6ece9b7b966aa594854c6da9930

PR #1771, opened 2018-03-27
feat: support vuex state

Shyam-Chen:
```js
// mutations.js
// @flow

import { ICounter } from './constants';

export default {
  increment(state: ICounter): void {
    state.value += 1;
  },
  decrement(state: ICounter): void {
    state.value -= 1;
  },
};
```


ljharb:
Is Vuex different from Vue?

Shyam-Chen:
Vuex is a state management for Vue.

ljharb:
“state” seems like a very generic argument name :-/ I’m a bit concerned about the continued expansion of this list.

PR #1779, opened 2018-04-04
warn on cyclical imports

sharmilajesupaul:
Adds [`import/no-cycle`](https://github.com/benmosher/eslint-plugin-import/blob/master/docs/rules/no-cycle.md) rule to `airbnb-config-base`

### DIFF
--- a/packages/eslint-config-airbnb-base/rules/imports.js
+++ b/packages/eslint-config-airbnb-base/rules/imports.js
@@ -228,8 +228,7 @@ module.exports = {
 
     // Forbid cyclical dependencies between modules
     // https://github.com/benmosher/eslint-plugin-import/blob/d81f48a2506182738409805f5272eff4d77c9348/docs/rules/no-cycle.md
-    // TODO: enable, semver-major
-    'import/no-cycle': ['off', { maxDepth: Infinity }],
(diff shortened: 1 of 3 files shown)


lencioni:
Do we want to set maxDepth?

sharmilajesupaul:
@lencioni seems like by default, the `maxDepth` is set to `Infinity`. Is there a certain depth we want to allow? or should I explicitly set the `maxDepth` to the default in our config, so the behavior doesn't change if the default is changed?

packages/eslint-config-airbnb-base/rules/imports.js
```diff
@@ -226,6 +226,10 @@ module.exports = {
     // https://github.com/benmosher/eslint-plugin-import/blob/44a038c06487964394b1e15b64f3bd34e5d40cde/docs/rules/no-self-import.md
     'import/no-self-import': 'error',
 
+    // Forbid cyclical dependencies between modules
+    // https://github.com/benmosher/eslint-plugin-import/blob/d81f48a2506182738409805f5272eff4d77c9348/docs/rules/no-cycle.md
+    'import/no-cycle': 'error',
```
ljharb: please do always hardcode the defaults (Infinity is fine)

sharmilajesupaul:
ok updated @ljharb

PR #1780, opened 2018-04-06
fix: bad comment already had a space at the beginning

HenrYxZ:
For section 18.3 of the README, the example of a bad comment shouldn't have a space at the beginning.


HenrYxZ:
Then what is wrong with that comment to be classified as a bad comment?

HenrYxZ:
hahaha no problem xD

PR #1787, opened 2018-04-18
Replace flag images with emoji

lencioni:
We can use emoji for most of these flags, which should make
them look better on higher resolution displays and improve the
loading performance of the style guide a little.

The only flag we have here that does not have a good emoji yet
is the flag of Catalonia, which I left as-is for now.

https://emojipedia.org/flag-for-catalonia-esct/

### DIFF
--- a/README.md
+++ b/README.md
@@ -3640,22 +3640,22 @@ Other Style Guides
 
   This style guide is also available in other languages:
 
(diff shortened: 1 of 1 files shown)


ljharb:
Does Github handle rendering these properly on browsers and devices that lack proper emoji support?

lencioni:
Looking at the source code in my inspector, it looks like it probably does. Here's what it renders for the Brazil flag:

```html
<g-emoji class="g-emoji" alias="brazil" fallback-src="https://assets-cdn.github.com/images/icons/emoji/unicode/1f1e7-1f1f7.png">🇧🇷</g-emoji>
```

PR #1794, opened 2018-04-30
only one newline at the end

graingert:
it seems 'no-multiple-empty-lines'.maxEOF is out by one


graingert:
Good:

`export default 'foo';\n`

Bad:

`export default 'foo';\n\n`

ljharb:
This seems like an eslint issue then - has one been filed upstream?

graingert:
No it seems intentional https://github.com/eslint/eslint/blob/fc7f404365e87fb6ee4f30149e486229cda4eb2e/lib/rules/no-multiple-empty-lines.js#L61

PR #1802, opened 2018-05-06
Remove redundant component from label-has-for

taion:
Also make the rule configuration more explicit.

### DIFF
--- a/packages/eslint-config-airbnb/rules/react-a11y.js
+++ b/packages/eslint-config-airbnb/rules/react-a11y.js
@@ -48,7 +48,13 @@ module.exports = {
 
     // require that JSX labels use "htmlFor"
     // https://github.com/evcohen/eslint-plugin-jsx-a11y/blob/master/docs/rules/label-has-for.md
-    'jsx-a11y/label-has-for': ['error', { components: ['label'] }],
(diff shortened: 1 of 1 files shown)


packages/eslint-config-airbnb/rules/react-a11y.js
```diff
@@ -48,7 +48,13 @@ module.exports = {
 
     // require that JSX labels use "htmlFor"
     // https://github.com/evcohen/eslint-plugin-jsx-a11y/blob/master/docs/rules/label-has-for.md
-    'jsx-a11y/label-has-for': ['error', { components: ['label'] }],
+    'jsx-a11y/label-has-for': ['error', {
+      components: [],
+      required: {
+        every: ['nesting', 'id'],
```
taion: The default here of "every" rather than "some" is sort of weird, BTW. While in principle some screen readers only support one or the other, in practice every modern screen reader supports both. We always override this to "some".

ljharb: This is incorrect; the latest version of Dragon, for example, only supports for-id linking, and there are definitely a few that only support nesting.

The proper default here is `every`, if one wants to be truly accessible.

PR #1819, opened 2018-05-29
Trailing comma's as //bad?????

csanda87:
Sounds like an opinion


ljharb:
It's a style guide, almost everything is opinion.

PR #1824, opened 2018-06-03
Add peer-version-check to CI step

stefanbuck:
I noticed that you're using the same workaround to deal with `peerDependencies` for development like we do by duplicating `peerDependencies` as `devDependencies`. 

Obviously duplicating is not the best solution, but it works. However there is a risk that they can get out of sync easily. To solve this issue, I wrote this CLI which checks if every peerDependency is defined as a devDependency with the same semver range.  Exits with code 1 and an error message if the check fails and 0 if not.

https://github.com/stefanbuck/peer-version-check

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -12,6 +12,9 @@ install:
   - 'if [ -n "${PACKAGE-}" ]; then cd "packages/${PACKAGE}"; fi'
   - 'if [ "${TRAVIS_NODE_VERSION}" = "0.6" ] || [ "${TRAVIS_NODE_VERSION}" = "0.9" ]; then nvm install --latest-npm 0.8 && npm install && nvm use "${TRAVIS_NODE_VERSION}"; else npm install; fi;'
   - 'if [ -n "${ESLINT}" ]; then npm install --no-save "eslint@${ESLINT}"; fi'
+before_script:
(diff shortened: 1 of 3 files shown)


stefanbuck:
Ohh just realised that your support quite a big range of different node versions. Have to check if my script works with node 4 and beyond.

stefanbuck:
I'm using https://github.com/sindresorhus/meow which requires node 6 along with some es6 sugar. Let me know if you're interested in a compatible version.

ljharb:
`npm ls` already does a peer version check, there is ZERO reason for a separate package for that to ever exist - we could add `npm ls` to `before_script`, i suppose.

Duplicating peerDeps into deps or devDeps isn't a workaround, it's a best practice.

(unless i'm misunderstanding, and the package explicitly compares the ranges to make sure they're identical - in which case that seems great! but it'd have to support super old nodes, and i'd prefer not to depend on any of that author's packages)

stefanbuck:
Ohh ... I didn't know that `npm ls` supports this already. I'm using a simple `isEqual` to compare like `npm ls` also does. Initial I was considering to compare the ranges, but wasn't sure about side effects, so I went with the simplest solution for now.

PR #1825, opened 2018-06-05
Add DEV/CI scripts to no-extraneous-dependencies exception list

dietergeerts:
Several of our projects have script files that are executed in our CI build, like for example to check the licenses of npm packages used. These are not used for production, and thus the dependencies used are devDependencies. The current rule settings report errors now for those imports. Right now there is only testing and build config stuff in the exception list, but other non-production scripts is also common to many projects. So that's the reason for this pull-request. If such scripts are put in a scripts folder, I think that is clear, but if there are other opinions about this, please share.


ljharb:
At some point, it's easier to use a glob override to reconfigure the rule for specific folders. I think "scripts" is generic enough that it's best not to put it in the overall config.

PR #1827, opened 2018-06-07
Max 7799 Apply sort-keys ESLint Rule

shinxi:
* Update readme to add sort-keys.
* Add docs for invalid sort-keys cases.


shinxi:
Wrong place to send...

PR #1828, opened 2018-06-08
Add no-unused-vars to README

sangaman:
This documents the enforced but currently undocumented `no-unused-vars` rule with a brief description, examples, and noting the `"ignoreRestSiblings": true` option.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1743,6 +1743,45 @@ Other Style Guides
     const foo = 'superLongLongLongLongLongLongLongLongString';
     ```
 
+  - [13.8](#variables--no-unused-vars) Disallow unused variables. eslint: [`no-unused-vars`](https://eslint.org/docs/rules/no-unused-vars)
+
+    > Why? Variables that are declared and not used anywhere in the code are most likely an error due to incomplete refactoring. Such variables take up space in the code and can lead to confusion by readers.
+
+    ```javascript
+    // bad
+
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -1743,6 +1743,43 @@ Other Style Guides
...
+    // Write-only variables are not considered as used.
+    var y = 10;
+    y = 5;
+
+    // A read for a modification of itself is not considered as used.
+    var z = 0;
+    z = z + 1;
+
+    // Unused function arguments.
+    function getY(x, y) {
```
ljharb: in this case, `x` won't be reported because `y` is used. if you `return x`, however, then `y` would be reported.

README.md
```diff
@@ -1743,6 +1743,43 @@ Other Style Guides
...
+    function getXPlusY(x, y) {
+      return x + y;
+    }
+
+    var x = 1;
+    var y = a + 2;
+
+    alert(getXPlusY(x, y));
+
+    // 'type' is ignored even if unused because it has a rest property sibling.
```
ljharb: let's elaborate on why: this is a form of extracting an object that omits specific keys.

README.md
sangaman: Interesting, I did not know that. Will fix.

sangaman:
Thanks, I amended the commit with your feedback.

PR #1831, opened 2018-06-20
Add new methods introduced in react@16.3

### LINKED ISSUE #1805: Add `getSnapshotBeforeUpdate` to `exceptMethods` of `class-methods-use-this`
The new [`getSnapshotBeforeUpdate`][react-docs] `React.Component` lifecycle method, introduced in React 16.3, is currently [not listed in the `exceptMethods` array of the `class-methods-use-this` rule][code] (package `eslint-config-airbnb`).

This results in a error/warning when no `this` keyword is used within this method.

gh-1805-preview

Besides, I think that the now deprecated lifecycles prefixed with `UNSAFE_` should be included too. If you agree the title of the issue should be adjusted.

I can submit a PR if this change is accepted.

[react-docs]: https://reactjs.org/docs/react-component.html#getsnapshotbeforeupdate
[code]: https://github.com/airbnb/javascript/blob/master/packages/eslint-config-airbnb/rules/react.js#L20-L33

paliwalgaurav:
React have introduced new methods and marked few old methods as Legacy methods and adviced not to use in new code. I've added all that methods in this commit for eslint rules. 

Fixes #1805.

### DIFF
--- a/packages/eslint-config-airbnb/rules/react.js
+++ b/packages/eslint-config-airbnb/rules/react.js
@@ -23,13 +23,17 @@ module.exports = {
         'getDefaultProps',
         'getChildContext',
         'componentWillMount',
+        'UNSAFE_componentWillMount',
         'componentDidMount',
         'componentWillReceiveProps',
+        'UNSAFE_componentWillReceiveProps',
         'shouldComponentUpdate',
         'componentWillUpdate',
+        'UNSAFE_componentWillUpdate',
         'componentDidUpdate',
         'componentWillUnmount',
         'componentDidCatch',
+        'getSnapshotBeforeUpdate'
       ],
     }],
 


ljharb:
(linking to #1805)

paliwalgaurav:
Okay @ljharb

packages/eslint-config-airbnb/rules/react.js
```diff
@@ -23,13 +23,18 @@ module.exports = {
...
         'componentDidMount',
         'componentWillReceiveProps',
+        'UNSAFE_componentWillReceiveProps',
         'shouldComponentUpdate',
         'componentWillUpdate',
+        'UNSAFE_componentWillUpdate',
         'componentDidUpdate',
         'componentWillUnmount',
         'componentDidCatch',
+        'getDerivedStateFromProps',
```
ljharb: This shouldn't be needed, because `getDerivedStateFromProps` is a static method - static methods aren't checked by this rule.

PR #1852, opened 2018-07-02
Fixed issue by adding return in map

combizs:


README.md
```diff
@@ -945,7 +945,7 @@ Other Style Guides
     // bad
     [1, 2, 3].map(number => {
       const nextNumber = number + 1;
-      `A string containing the ${nextNumber}.`;
+      return `A string containing the ${nextNumber}.`;
```
ljharb: This is a "bad" example - the missing `return` is the whole reason it's bad, so it needs to stay missing.

PR #1854, opened 2018-07-03
Updated Arrays Section

### LINKED ISSUE #1851: [Arrays] To convert an array-like object to an array, use spreads ... instead of Array.from
I do not agree with what you can read in paragraph [4.4](https://github.com/airbnb/javascript#arrays--from).

In fact, my understanding is that _spread syntax_ can only be applied to _iterable objects_ and not to _array-like objects_.

References:

- [MDN](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Spread_syntax#Only_for_iterables)
- [Javscript.info](https://javascript.info/rest-parameters-spread-operator#spread-operator)

Before changes introduced with issue #1084, you could read:

> To convert an _array-like object_ to an array, use **Array.from**

which I consider being correct.

Furthermore, I would personally expand it by saying:

> To convert an _iterable object_ or an _array-like object_ to an array, use **Array.from**

That is because I agree with the authors of [Javscript.info](https://javascript.info/rest-parameters-spread-operator#spread-operator):

> So, for the task of turning something into an array, Array.from tends to be more universal.

maurozenoni:
- corrected paragraph 4.4 (conversion from iterable to array)
- added paragraph 4.5 (conversion from array-like to array)

Closes #1851


README.md
```diff
@@ -369,8 +369,8 @@ Other Style Guides
     const itemsCopy = [...items];
     ```
 
-  <a name="arrays--from"></a><a name="4.4"></a>
-  - [4.4](#arrays--from) To convert an array-like object to an array, use spreads `...` instead of [`Array.from`](https://developer.mozilla.org/en/docs/Web/JavaScript/Reference/Global_Objects/Array/from).
+  <a name="arrays--from-iterable"></a><a name="4.4"></a>
```
ljharb: we'll need to keep an `<a name="arrays--from"></a>` so that existing links don't break.

README.md
```diff
@@ -382,8 +382,21 @@ Other Style Guides
     const nodes = [...foo];
     ```
 
-  <a name="arrays--mapping"></a>
-  - [4.5](#arrays--mapping) Use [Array.from](https://developer.mozilla.org/en/docs/Web/JavaScript/Reference/Global_Objects/Array/from) instead of spread `...` for mapping over iterables, because it avoids creating an intermediate array.
+  <a name="arrays--from-array-like"></a><a name="4.5"></a>
```
ljharb: there's no need to add this 4.5 anchor here; numbered links are legacy

README.md
```diff
@@ -393,8 +406,8 @@ Other Style Guides
     const baz = Array.from(foo, bar);
     ```
 
-  <a name="arrays--callback-return"></a><a name="4.5"></a>
-  - [4.6](#arrays--callback-return) Use return statements in array method callbacks. It’s ok to omit the return if the function body consists of a single statement returning an expression without side effects, following [8.2](#arrows--implicit-return). eslint: [`array-callback-return`](https://eslint.org/docs/rules/array-callback-return)
+  <a name="arrays--callback-return"></a><a name="4.7"></a>
```
ljharb: and none of the others section's links should be renumbered.

Issue #1834, opened 2018-06-21
eslint v5 support


ljharb:
Just like #1447, I'm filing this in advance.

v17.0.0 of `eslint-config-airbnb` and v13.0.0 of `eslint-config-airbnb-base` have just been released.

eslint v5 is planned to be released in the next 24-48 hours, and the plan for our configs will be to either support v3 || v4 || v5, or, if necessary, to drop support for v3 and/or v4 as part of this release.

The timeframe is "probably not as soon as you would like, but as soon as we can".
Please accept this, and please add no noise to this thread beyond emoji reactions - distractions won't help anyone get the work done any faster.

### **Please** do not file additional issues and do not file pull requests attempting to add support - please, instead, continue to use eslint v4 until such time as our configs support v5.

Peer Dep eslint combined v3 and v4 and v5 support released, or v5 support released (depending on which path we need to take):
 - [x] `eslint-plugin-import`
    - [x] [PR](https://github.com/benmosher/eslint-plugin-import/pull/1122)
    - [x] Released as v2.13.0
 - [x] `eslint-find-rules`
   - [x] [PR](https://github.com/sarbbottam/eslint-find-rules/pull/296)
   - [x] [Release](https://github.com/sarbbottam/eslint-find-rules/pull/300) as v3.3.1
 - [x] update the base config
 - [x] publish the base config (v13.1.0)
 - [x] `eslint-plugin-react`
   - [x] [PR](https://github.com/yannickcr/eslint-plugin-react/pull/1843)
   - [x] Released as v7.10.0
 - [x] `eslint-plugin-jsx-a11y`
   - [x] [PR](https://github.com/evcohen/eslint-plugin-jsx-a11y/pull/451)
   - [x] [Released](https://github.com/evcohen/eslint-plugin-jsx-a11y/pull/452) as v6.1.0
 - [x] eslint-plugin-react
   - [x] enable new option in (jsx-one-expression-per-line](https://github.com/yannickcr/eslint-plugin-react/issues/1775)
   - [x] released as v7.11.0
   - [x] updated in the main config
 - [x] a11y plugin: fix [idlist bug](https://github.com/evcohen/eslint-plugin-jsx-a11y/issues/454) ([PR](https://github.com/evcohen/eslint-plugin-jsx-a11y/pull/456)) and release
 - [x] a11y plugin: [label-has-for bug](https://github.com/evcohen/eslint-plugin-jsx-a11y/issues/455)
 - [x] update and publish the main config

[...]

PR #1857, opened 2018-07-04
break(base): upgrade eslint v5 & drop support node 4, 5

Aladdin-ADD:


ljharb:
Please see #1834 where i explicitly asked for no pull requests to be filed. (See also #1856)

PR #1865, opened 2018-07-11
feat: enable no-negated-condition for readability

Aladdin-ADD:
is there any discussion to disable it?


ljharb:
There’s not; and an issue would be appropriate to do so before a pull request.

I don’t find that this always increases readability, personally i find that it can decrease it. Either way, there’s no interest internally in legislating this.

PR #1879, opened 2018-08-09
Update README.md

schalivendri:
Removing incorrect bad function


ljharb:
As you may have realized, the "bad" examples are intentionally bad so as to demonstrate what "good" means :-)

PR #1891, opened 2018-08-21
FDH-591: set "prefer-default-export": "off"

GregorSondermeier:
I've set the rule option to "off" in the monolith and removed the paragraph from this README.


GregorSondermeier:
ignore this. I have set the wrong target.

PR #1903, opened 2018-08-31
add   - [3.9](#objects--attributes-attributes) Use javascript declara…

odykyi:
…tion attributes mapping without values firstly in object

### DIFF
--- a/README.md
+++ b/README.md
@@ -324,6 +324,37 @@ Other Style Guides
     const { a, ...noA } = copy; // noA => { b: 2, c: 3 }
     ```
 
+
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -324,6 +324,35 @@ Other Style Guides
     const { a, ...noA } = copy; // noA => { b: 2, c: 3 }
     ```
 
+  <a name="objects--attributes-attributes"></a>
+  - [3.9](#objects--attributes-attributes) Use javascript declaration attributes mapping without values firstly in object
```
ljharb: This isn't a prescription we want to make (object key ordering).

PR #1904, opened 2018-08-31
Update README.md

odykyi:


README.md
```diff
@@ -195,6 +195,15 @@ Other Style Guides
   - [3.3](#es6-object-shorthand) Use object method shorthand. eslint: [`object-shorthand`](https://eslint.org/docs/rules/object-shorthand.html)
 
     ```javascript
+    // bad
+    const atom = {
+      value: 1,
+
+      addValue: function (value) {
+        return this.value + value;
```
ljharb: this bad example only differentiates by using `this`, and that's not what "use object method shorthand" is trying to discuss. I don't think this example belongs here.

PR #1916, opened 2018-09-14
Examples more consistent with the guidline

ernestodebesto:
comparison operator <= shows better the confusion when using arrow function, than just operator <

### DIFF
--- a/README.md
+++ b/README.md
@@ -1052,18 +1052,18 @@ Other Style Guides
 
     ```javascript
     // bad
-    const itemHeight = item => item.height > 256 ? item.largeSize : item.smallSize;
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -1052,18 +1052,18 @@ Other Style Guides
 
     ```javascript
     // bad
-    const itemHeight = item => item.height > 256 ? item.largeSize : item.smallSize;
+    const itemHeight = item => item.height <= 256 ? item.largeSize : item.smallSize;
```
ljharb: I don’t agree; i think the reverse of the operator shows it better than half of the tokens of the operator. Do you have more rationale for this change than that?

PR #1917, opened 2018-09-15
Allow modification of draft and draftState for "immer"

brummelte:
The immer library requires the modification of parameter properties.

https://github.com/mweststrate/immer/issues/189

### DIFF
--- a/packages/eslint-config-airbnb-base/rules/best-practices.js
+++ b/packages/eslint-config-airbnb-base/rules/best-practices.js
@@ -179,13 +179,15 @@ module.exports = {
       ignorePropertyModificationsFor: [
         'acc', // for reduce accumulators
         'accumulator', // for reduce accumulators
-        'e', // for e.returnvalue
+        'e', // for e.returnValue
         'ctx', // for Koa routing
         'req', // for Express requests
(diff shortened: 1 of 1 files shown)


packages/eslint-config-airbnb-base/rules/best-practices.js
```diff
@@ -179,13 +179,15 @@ module.exports = {
...
-        'e', // for e.returnvalue
+        'e', // for e.returnValue
         'ctx', // for Koa routing
         'req', // for Express requests
         'request', // for Express requests
         'res', // for Express responses
         'response', // for Express responses
         '$scope', // for Angular 1 scopes
+        'draft', // for immer
+        'draftState', // for immer
```
ljharb: i don't think we need to add specific exceptions for every mutating pattern, and i'm not yet convinced that immer's pattern of encouraging mutation in code is a beneficial one.

spawnia:
As @mweststrate summarized quite nicely, this exception is really the only way for the `no-param-reassign` rule to play nicely with `immer`.

@ljharb i think the popularity of `immer` is evidence enough that it is useful, and those 2 extra lines of config should be well worth it.

PR #1936, opened 2018-10-17
Update to babel 7

vikr01:
Not much to do here, just updating the dependencies should be enough. (Although I would drop use of [`babel-tape-runner`](https://github.com/wavded/babel-tape-runner) for `tape --require @babel/register`.)

With `babel` v7, we can also use `babel.config.js` to run all tests at the same time. If interested I can PR that as well.


vikr01:
True, forgot it doesn't work on `node@<=4`

PR #1941, opened 2018-10-23
Add node 11 for CI

noelyoo:
## Summary

This PR adds `node` since Node.js `11` has been released.

https://github.com/nodejs/node/blob/master/doc/changelogs/CHANGELOG_V11.md#11.0.0

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,5 +1,6 @@
 language: node_js
 node_js:
+  - "11"
   - "10"
   - "9"
   - "8"


.travis.yml
```diff
@@ -1,5 +1,6 @@
 language: node_js
 node_js:
+  - "node"
```
ljharb: we don't want to default to just "the latest", this should be `11`.

noelyoo:
@ljharb I agree. Thanks

PR #1946, opened 2018-10-30
[19.16] changed Enforce to avoid

theBstar:
Good practice is to avoid spaces between functions and their invocations. But by mistake (I think so) you used the word Enforce instead of Avoid

### DIFF
--- a/README.md
+++ b/README.md
@@ -2781,7 +2781,7 @@ Other Style Guides
     ```
 
   <a name="whitespace--func-call-spacing"></a>
-  - [19.16](#whitespace--func-call-spacing) Enforce spacing between functions and their invocations. eslint: [`func-call-spacing`](https://eslint.org/docs/rules/func-call-spacing)
+  - [19.16](#whitespace--func-call-spacing) Avoid spaces between functions and their invocations. eslint: [`func-call-spacing`](https://eslint.org/docs/rules/func-call-spacing)
 
     ```javascript
     // bad


README.md
```diff
@@ -2781,7 +2781,7 @@ Other Style Guides
     ```
 
   <a name="whitespace--func-call-spacing"></a>
-  - [19.16](#whitespace--func-call-spacing) Enforce spacing between functions and their invocations. eslint: [`func-call-spacing`](https://eslint.org/docs/rules/func-call-spacing)
+  - [19.16](#whitespace--func-call-spacing) Avoid spaces between functions and their invocations. eslint: [`func-call-spacing`](https://eslint.org/docs/rules/func-call-spacing)
```
ljharb: Enforce is correct; we’re enforcing “no spaces”.

theBstar: 'Enforce spacing' is little confusing, As It's not clear whether to add spaces or not. Although Example definitely makes this clear or one have to check the eslint doc.

ljharb: The examples and the linter config should make it clear; "enforce no spaces" or "enforce spaces" would both be correct.

theBstar: " Enforce no spaces" or "Avoid spaces" makes it more cleaner than "Enforce spaces". "Enforce no spaces" is good ( If we really want to stick with 'enforce' ). The guide is mostly self explanatory and this change will help to keep it intact.

theBstar:
It's fine now !

PR #1951, opened 2018-11-05
jsx-a11y/label-has-associated-control: assert either not both

dlwalsh:
The current rules require an input within a label to be marked up with `id`/`htmlFor`. This seems bizarre and unnecessary.

That is, the following markup is perfectly accessible and the rules should reflect that:
```js
<label>
  <input type="checkbox" />
  Accept terms and conditions 
</label>
```

Similar examples are shown on the rule's readme doc. https://github.com/evcohen/eslint-plugin-jsx-a11y/blob/master/docs/rules/label-has-associated-control.md

### DIFF
--- a/packages/eslint-config-airbnb/rules/react-a11y.js
+++ b/packages/eslint-config-airbnb/rules/react-a11y.js
@@ -51,7 +51,7 @@ module.exports = {
     'jsx-a11y/label-has-for': ['error', {
       components: [],
       required: {
-        every: ['nesting', 'id'],
+        some: ['nesting', 'id'],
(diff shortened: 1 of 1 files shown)


pbarbiero:
I found this odd as well, nesting should satisfy the need for a for attribute: `When absent, the label being defined is associated with the element's contents.` (https://www.w3.org/TR/html401/interact/forms.html#h-17.9.1)

Are there assistive devices that violate this rule? Or is it because the nesting rule is insufficient and could let cases slip through where the label cant match up with a field?

PR #1957, opened 2018-11-14
Use correct github url for eslint-config-airbnb package

danwenzel:
When clicking the github link from the eslint-config-airbnb npm page, it should go to the github page for this specific package, not the parent repo

### DIFF
--- a/packages/eslint-config-airbnb/package.json
+++ b/packages/eslint-config-airbnb/package.json
@@ -18,7 +18,7 @@
   },
   "repository": {
     "type": "git",
-    "url": "https://github.com/airbnb/javascript"
+    "url": "https://github.com/airbnb/javascript/tree/master/packages/eslint-config-airbnb"
   },
   "keywords": [
     "eslint",


packages/eslint-config-airbnb/package.json
```diff
@@ -18,7 +18,7 @@
   },
   "repository": {
     "type": "git",
-    "url": "https://github.com/airbnb/javascript"
+    "url": "https://github.com/airbnb/javascript/tree/master/packages/eslint-config-airbnb"
```
ljharb: this link would break if we restructured the repo, or changed the default branch. It might be better to instead link to the package readmes at the top of the main repo, and continue to link there.

(Separately, the config isn't meant to be used independently of the guide)

danwenzel: Good point about the link possibly breaking down the road.

> It might be better to instead link to the package readmes at the top of the main repo

Are you saying just keep it as-is, or is there some other specific link that you're referring to?

Just to give my use case:  A google search for the config led me to the npm page, but I always like to click over to the github page from the npm page.  I already know about the guide - I just wanted to see the instructions for installing the config.

ljharb: I'm saying to keep the link in package.json as-is, but to add prominent links to the individual package readmes at the top of the guide itself.

danwenzel: Yeah, I know... it's just a workflow I'm used to, because I much prefer the formatting of readmes in github.

I'm just going to close this, as I may be the only one in this boat and it's probably not worth any more effort.  

Thanks!

PR #1960, opened 2018-11-19
[17.2] Fixed linking

NLDev:
Missing 's' prevented the anchor link from working

### DIFF
--- a/README.md
+++ b/README.md
@@ -2249,7 +2249,7 @@ Other Style Guides
     }
     ```
 
-  <a name="control-statement--value-selection"></a>
+  <a name="control-statements--value-selection"></a>
   - [17.2](#control-statements--value-selection) Don't use selection operators in place of control statements.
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -2249,7 +2249,7 @@ Other Style Guides
     }
     ```
 
-  <a name="control-statement--value-selection"></a>
+  <a name="control-statements--value-selection"></a>
```
ljharb: While this appears to be a typo from the [original commit](c8b116410), to prevent links from breaking, we need to keep the singular one forever.

Either we need to fix the links to point to the plural one, or, we should add an anchor tag here for both.

NLDev:
The approach with the second anchor is probably the best. Made a mistake by deleting the fork, so I'll close this PR here. 
New one: https://github.com/airbnb/javascript/pull/1963

PR #1962, opened 2018-11-20
Consolidate `strict` rule definition between airbnb and airbnb-base

### LINKED ISSUE #1582: Consolidating `strict` rule definition between airbnb and airbnb-base
Hi!

I've spotted some inconsistency/duplication with the `strict` rule and wondered if this was intentional?

Currently:
* There's the [eslint-config-airbnb-base/rules/strict](https://github.com/airbnb/javascript/blob/dfbeaaea4bf09be28463eea15979dd8c862fc10a/packages/eslint-config-airbnb-base/rules/strict.js#L4) file that declares `strict: ['error', 'never']`.
* That file is [only used in](https://github.com/airbnb/javascript/blob/dfbeaaea4bf09be28463eea15979dd8c862fc10a/packages/eslint-config-airbnb/index.js#L4) `eslint-config-airbnb` and not `eslint-config-airbnb-base`.
* Instead `eslint-config-airbnb-base` sets `strict: 'error'` manually [in its rules directive](https://github.com/airbnb/javascript/blob/dfbeaaea4bf09be28463eea15979dd8c862fc10a/packages/eslint-config-airbnb-base/index.js#L19) (which results in the ESLint default of `strict: ['error', 'safe']`).
* However both `eslint-config-airbnb-base` and `eslint-config-airbnb`  set `sourceType: 'module'`, which causes the ESLint strict rule to "_disallow[s] strict mode directives, no matter which option is specified_" regardless of those options - ie: `strict: ['error', 'never']` ([docs](https://eslint.org/docs/rules/strict#rule-details)).

Issues:
* It was my understanding that for the most part, the generic (ie non react, non-AirBnb-as-as-company specific) rules/configuration belong in `eslint-config-airbnb-base` rather than `eslint-config-airbnb`, but yet the `eslint-config-airbnb-base/rules/strict` `extends` is contrary to that, and confusing to the end user **if** it actually resulted in different strict mode behaviour.
* However unless people override `sourceType: 'module'`, both `eslint-config-airbnb-base` and `eslint-config-airbnb` actually result in the same strict mode behaviour (ie: equivalent of `strict: ['error', 'never']`) anyway, so in most cases the rule being listed twice is just duplication rather than having any effect.

I think this situation might just be an accidental leftover from the split to the airbnb-base package ([see here](https://github.com/airbnb/javascript/commit/52d3401f953afcf09ca64dbbb177c6284832aae6#diff-3452a03d02d221986fbcc32f695263d2)) and then #637. 

As such, to increase consistency I would like to propose:
* Moving `eslint-config-airbnb-base/rules/strict` from `eslint-config-airbnb`'s `extends` to `eslint-config-airbnb-base`'s (this won't affect legacy).
* Deleting the manual `strict` rule definition from `eslint-config-airbnb-base`'s `rules`.
* Probably bumping the `eslint-config-airbnb-base` major version in case there's anyone using it who sets `sourceType: 'script'` manually rather than using legacy, since they would see a difference in behaviour.

If this sounds acceptable, I'm happy to open a PR :-)

edmorley:
See individual commit messages for more details.

Fixes #1582.

### DIFF
--- a/packages/eslint-config-airbnb-base/index.js
+++ b/packages/eslint-config-airbnb-base/index.js
@@ -7,12 +7,11 @@ module.exports = {
     './rules/variables',
     './rules/es6',
     './rules/imports',
+    './rules/strict',
   ].map(require.resolve),
   parserOptions: {
     ecmaVersion: 2018,
     sourceType: 'module',
   },
-  rules: {
-    strict: 'error',
-  },
+  rules: {},
 };
--- a/packages/eslint-config-airbnb-base/legacy.js
+++ b/packages/eslint-config-airbnb-base/legacy.js
@@ -27,6 +27,7 @@ module.exports = {
       property: '__defineSetter__',
       message: 'Please use Object.defineProperty instead.',
     }],
-    'no-var': 'off'
+    'no-var': 'off',
+    strict: ['error', 'safe'],
   }
 };
--- a/packages/eslint-config-airbnb/index.js
+++ b/packages/eslint-config-airbnb/index.js
@@ -1,7 +1,6 @@
 module.exports = {
   extends: [
     'eslint-config-airbnb-base',
-    'eslint-config-airbnb-base/rules/strict',
     './rules/react',
     './rules/react-a11y',
   ].map(require.resolve),


edmorley:
I've rebased on master to resolve the conflicts, given the next major release is likely to be not too far out :-)

PR #1863, opened 2018-07-10
[eslint config][major][guide] Always require parens around arrow function arguments

sharmilajesupaul:
guide + eslint config change to always require parens around arrow function arguments.

PR #1965, opened 2018-11-27
:) conflicting with rule 8.2 :D

michaelKaefer:


tunnckoCore:
Hm. Didn't we moved to "always use parens" recently (last 2-3 months)? Hm, why I am thinking about that.

ljharb:
@tunnckoCore you're thinking of #1863 which has not been merged yet. I've updated it to include these changes.

tunnckoCore:
Oooh yea, exactly : ) Thanks.

PR #1966, opened 2018-11-29
Add markers to spaced-comment block for Flow types

jeffvandyke:
Tested in one of my projects - Flow allows a syntax for typing in comments, but these markers have to be added to not throw errors. Flow could do it with a space in between, but it seems like an easy enough style to allow, and a good one to follow.

### DIFF
--- a/packages/eslint-config-airbnb-base/rules/style.js
+++ b/packages/eslint-config-airbnb-base/rules/style.js
@@ -498,7 +498,7 @@ module.exports = {
       },
       block: {
         exceptions: ['-', '+'],
-        markers: ['=', '!'], // space here to support sprockets directives
+        markers: ['=', '!', ':', '::'], // space here to support sprockets directives and flow comment types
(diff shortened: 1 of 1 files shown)


jeffvandyke:
To bump this, and to make it easier to view the original motivation for this PR, I'll link to the page that describes how Flow can work with type declarations inside these comments: <https://flow.org/en/docs/types/comments/>

packages/eslint-config-airbnb-base/rules/style.js
```diff
@@ -491,11 +491,13 @@ module.exports = {
     'spaced-comment': ['error', 'always', {
       line: {
         exceptions: ['-', '+'],
-        markers: ['=', '!'], // space here to support sprockets directives
+        // support sprockets directives
```
ljharb: please move the comments back to being inline; line length isn't worth worrying about.

PR #1972, opened 2018-12-18
Replace 'capitalize' with 'uppercase'

honzajavorek:
The examples and the reasoning indicates the author of the rule meant uppercase instead of capitalization. [According to Wikipedia](https://en.wikipedia.org/wiki/Capitalization), capitalization is 

> writing a word with its first letter as a capital letter (uppercase letter) and the remaining letters in lower case

while the rule apparently tries to prevent exactly that.

### DIFF
--- a/README.md
+++ b/README.md
@@ -3268,7 +3268,7 @@ Other Style Guides
     ```
 
   <a name="naming--Acronyms-and-Initialisms"></a>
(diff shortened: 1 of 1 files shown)


honzajavorek:
Thanks! Yup, the explanation makes sense, but I hope this is clearer for everyone.

PR #1991, opened 2019-01-12
Update 8.2 [arrow-body-style] improve good/bad example readability

nakhodkiin:
There is an inconsistency of behavior between functions in a `good and `bad` examples:

```javascript
// bad
[1, 2, 3].map(number => {
});

// good
[1, 2, 3].map(number => `A string containing the ${number}.`);
```

Actually, the first one will generate an array of 3 `undefined` since we're not returning anything from the `map`-function.

Assuming that `good` and `bad` examples, which are compared side-by-side should only consist of code-style differences, not different logic, which in this case might cause losing the focus from the main point - code style.

### DIFF
--- a/README.md
+++ b/README.md
@@ -958,8 +958,7 @@ Other Style Guides
     ```javascript
     // bad
     [1, 2, 3].map(number => {
-      const nextNumber = number + 1;
-      `A string containing the ${nextNumber}.`;
+      return `A string containing the ${number}.`;
     });
 
     // good


README.md
```diff
@@ -958,8 +958,7 @@ Other Style Guides
     ```javascript
     // bad
     [1, 2, 3].map(number => {
+      return `A string containing the ${number}.`;
```
ljharb: Unfortunately this is exactly why it’s a bad example, and it’s correct as is.

ljharb:
(To elaborate, the main point is always correctness - style is an essential part of ensuring correctness)

README.md
nakhodkiin: Oh, ok then, got it. Thanks for the clarification.

PR #1993, opened 2019-01-16
[guide] [react] add note about deprecated lifecycles

meetzaveri:
There were still deprecated lifecycle being mentioned in two places. Thus by removing them, it will make clear sense of latest React API after 16.2.

### DIFF
--- a/react/README.md
+++ b/react/README.md
@@ -627,8 +627,11 @@ We don’t recommend using indexes for keys if the order of items may change.
   1. optional `static` methods
   1. `constructor`
   1. `getChildContext`
+  1. `componentWillMount`
   1. `componentDidMount`
+  1. `componentWillReceiveProps`
   1. `shouldComponentUpdate`
+  1. `componentWillUpdate`
   1. `componentDidUpdate`
   1. `componentWillUnmount`
   1. *clickHandlers or eventHandlers* like `onClickSubmit()` or `onChangeDescription()`
@@ -680,8 +683,11 @@ We don’t recommend using indexes for keys if the order of items may change.
   1. `getDefaultProps`
   1. `getInitialState`
   1. `getChildContext`
+  1. `componentWillMount`
   1. `componentDidMount`
+  1. `componentWillReceiveProps`
   1. `shouldComponentUpdate`
+  1. `componentWillUpdate`
   1. `componentDidUpdate`
   1. `componentWillUnmount`
   1. *clickHandlers or eventHandlers* like `onClickSubmit()` or `onChangeDescription()`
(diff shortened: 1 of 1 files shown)


ljharb (changes requested):
These are not actually "deprecated" in the sense that they'll only be "unsafe with async rendering". Either way, we should leave these here, and let React itself (and eslint-plugin-react rules) warn on these methods as needed.

meetzaveri:
This makes sense. But can we like address developer that this lifecycles are now no more in use like by adding a label in parenthesis or making seperate section for it ?

ljharb:
I think a section for it that calls out that only in React 16.3+, certain methods should be avoided in favor of others, would be useful, yes.

meetzaveri:
Okay. So should I make changes now in readme file acc. to it?

ljharb:
Updating this PR to do that would be great!

meetzaveri:
Sure. So should I finally add a section about avoiding lifecycles?

ljharb:
Not avoiding; but using the ones appropriate for your React version. Not everyone is on the latest version.

react/README.md
```diff
@@ -701,6 +702,15 @@ We don’t recommend using indexes for keys if the order of items may change.
   > Why? [`isMounted` is an anti-pattern][anti-pattern], is not available when using ES6 classes, and is on its way to being officially deprecated.
 
   [anti-pattern]: https://facebook.github.io/react/blog/2015/12/16/ismounted-antipattern.html
+  
+## Things to avoid
```
ljharb: This isn't the right phrasing here; someone using React 16.2 shouldn't be avoiding anything.

react/README.md
```diff
@@ -701,6 +702,15 @@ We don’t recommend using indexes for keys if the order of items may change.
...
+
+Using deprecated APIs will create bad legacy code. Hence it's good to know that which APIs have been deprecated.
```
ljharb: All of these APIs are not "bad" or "legacy", they're solely not safe to use with async rendering. Can you rephrase this to have less of a judgemental connotation?

react/README.md
```diff
@@ -701,6 +702,15 @@ We don’t recommend using indexes for keys if the order of items may change.
...
+
+These Component lifecycles are been deprecated since React version 16.3+ :
```
ljharb: Could we add, for each one, which methods should be used instead (with links to their documentation)?

ljharb (changes requested):

MDASI:
There were still deprecated lifecycle being mentioned in two places. so by removing them, it will make clear sense of latest React API after 16.2.

PR #1994, opened 2019-01-16
Adjust lines-around-comment to enforce empty lines before comments

fnwbr:
Enabling this rule only enforces what's been in the styleguide for a while already: https://github.com/airbnb/javascript#comments


ljharb:
Although it's in the guide, we've avoided enabling this rule so far because there's concern it'd be too noisy.

fnwbr:
Gotcha, thanks for letting us know.

Going ahead and closing this!

PR #1996, opened 2019-01-21
Allow use of __REDUX_DEVTOOLS_EXTENSION_COMPOSE__

remcohaszing:
This style guide disallows the use of underscores in properties. However, the Redux browser extension requires the use of a variable named `__REDUX_DEVTOOLS_EXTENSION_COMPOSE__`. Since Redux is so popular with React, it makes sense to allow this.

This saves an `eslint-disable` comment in pretty much all of my React projects. I’m probably not the only one.


ljharb:
What's the use of this property? It doesn't appear anywhere in the airbnb codebase, and we use redux extensively.

ljharb:
(i'm going to tweak this a bit so it applies when using the main config, and not the base)

PR #1998, opened 2019-01-30
Add jest.setup.js to import/no-extraneous-dependencies devDependencies

tclindner:
Hi team - Thank you for all of your excellent work on this project. 🙌 I use Jest on a number of projects. A pattern Jest users utilize is moving setup to a separate `jest.setup.js` file. I would like to use devDependencies in this file without receiving an error from `import/no-extraneous-dependencies`. Would you be open to adding a new pattern to the `import/no-extraneous-dependencies` rule? Please let me know if you have questions or would like me to make changes. Thank you in advance!


tclindner:
Thank you!

PR #2000, opened 2019-02-06
Fix releated to arrow function

ziishaned:


arcticicestudio:
This would print the numbers `2`, `3` and `4`instead of `1`, `2` and `3`. The `number` variable contains the actual value if the index but not the index number itself so there's no need to increment it by one. It doesn't start with `0` only because it processes an array. I think that' the reason you submitted this PR.

To get the index of the currently processed array value use the second return value:

```js
[1, 2, 3].map((number, index) => `A string containing the ${number} at index ${index}.`);
```

Take a look at the [documentation of the `Array.map()` function][mdn] for more details.

[mdn]: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/map

README.md
```diff
@@ -963,7 +963,7 @@ Other Style Guides
     });
 
     // good
-    [1, 2, 3].map(number => `A string containing the ${number}.`);
+    [1, 2, 3].map(number => `A string containing the ${number + 1}.`);
```
ljharb: This remains correct, but why is this “fixing” anything? How does output of 2/3/4 make this example better than it being 1/2/3?

ziishaned: The fix is to make the "good example" output the same value as bad example.

PR #2001, opened 2019-02-07
Improve ESLint config setup instructions for Yarn

plbrault:
It is now possible to use the `npx install-peerdeps` command when using Yarn, since it will detect that the user is using Yarn and will act accordingly:

screenshot from 2019-02-07 11-54-56

Therefore, this pull request improves the setup instructions to mention that it is possible to use this shortcut while using Yarn as well.

### DIFF
--- a/packages/eslint-config-airbnb/README.md
+++ b/packages/eslint-config-airbnb/README.md
@@ -12,8 +12,6 @@ We export three ESLint configurations for your usage.
 
 Our default export contains all of our ESLint rules, including ECMAScript 6+ and React. It requires `eslint`, `eslint-plugin-import`, `eslint-plugin-react`, and `eslint-plugin-jsx-a11y`. If you don't need React, see [eslint-config-airbnb-base](https://npmjs.com/eslint-config-airbnb-base).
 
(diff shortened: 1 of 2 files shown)


packages/eslint-config-airbnb/README.md
```diff
@@ -26,6 +24,9 @@ If you use yarn, run `npm info "eslint-config-airbnb@latest" peerDependencies` t
   npx install-peerdeps --dev eslint-config-airbnb
   ```
 
+  If using **Yarn**, you can also use the shortcut described above if you have npm 5+ installed on your machine, as the command will detect that you are using Yarn and will act accordingly.
+  Otherwise, run `npm info "eslint-config-airbnb@latest"   peerDependencies` to list the peer dependencies and versions, then run `yarn add --dev <dependency>@<version>` for each listed peer dependency.
```
ljharb:
```suggestion
  Otherwise, run `npm info "eslint-config-airbnb@latest" peerDependencies` to list the peer dependencies and versions, then run `yarn add --dev <dependency>@<version>` for each listed peer dependency.
```

PR #2003, opened 2019-02-13
9.3 fix wrong example

biggieman:


README.md
```diff
@@ -1158,8 +1158,8 @@ Other Style Guides
     };
 
     const luke = new Jedi();
-    luke.jump(); // => true
-    luke.setHeight(20); // => undefined
+    luke.jump() // => true
+      .setHeight(20); // => undefined
```
ljharb: this won't work as-is; which does make it "bad", but the point was to show a working example that doesn't use chaining.

biggieman: Yeap, you right!

Issue #1980, opened 2019-01-02
autofix "import/order" will produce new problems?


hbshun:
env:
os: macos 10.14.2
node: 10.15.0
npm: 6.4.1
eslint: local 5.11.1

package.json
```json
{
  "name": "testeslint",
  "version": "0.0.1",
  "scripts": {
    "eslint": "eslint --ext .js .",
  },
  "devDependencies": {
    "eslint": "^5.11.1",
    "eslint-config-airbnb-base": "^13.1.0",
    "eslint-plugin-import": "^2.14.0",
    "express": "^4.16.4"
  }
}
```
.eslintrc.js
```js
module.exports = {
    "extends": "airbnb-base"
};
```

a.js
```js
const env = require('./config');

Object.keys(env);

const http = require('http');

function express() {}

http.createServer(express());
```

b.js
```js
const env = require('./config');

Object.keys(env);

const http = require('http');
const express = require('express');

http.createServer(express());
```

```bash
$ npm run eslint

> testeslint@0.0.1 eslint /Users/hbshun/testeslint
> eslint --ext .js .

/Users/hbshun/testeslint/a.js
  5:14  error  `http` import should occur before import of `./config`  import/order

/Users/hbshun/testeslint/b.js
  1:13  error  `./config` import should occur after import of `express`                       import/order
  6:17  error  'express' should be listed in the project's dependencies, not devDependencies  import/no-extraneous-dependencies

✖ 3 problems (3 errors, 0 warnings)
  1 error and 0 warnings potentially fixable with the `--fix` option.
```

```bash
$ npm run eslint -- --fix

> testeslint@0.0.1 eslint /Users/hbshun/testeslint
> eslint --ext .js . "--fix"

/Users/hbshun/testeslint/a.js
  5:14  error  `http` import should occur before import of `./config`  import/order

/Users/hbshun/testeslint/b.js
  2:13  error  'env' was used before it was defined                                           no-use-before-define
  5:17  error  'express' should be listed in the project's dependencies, not devDependencies  import/no-extraneous-dependencies

✖ 3 problems (3 errors, 0 warnings)
```

Now, b.js:

```js

Object.keys(env);

const http = require('http');
const express = require('express');
const env = require('./config');

http.createServer(express());
```

[repo for test](https://github.com/hbshun/testeslint)

ljharb:
This should be filed on the eslint-plugin-import repo, since it’s a bug in its autofixer.

hbshun:
[...]

PR #2005, opened 2019-02-14
Bump version of eslint-plugin-import

### LINKED ISSUE #2002: Why is the definition of a variable is moved after using?
The require of the `package.json` is moved later — the resulting code will not work

Why such code:
```js
const pkg = require('../package.json')

if (!process.version !== pkg.engines.node) { /* .. */ }

const app = require('yargs')
const updateNotifier = require('update-notifier')
```

will be turned into such:
```js
 
if (!process.version !== pkg.engines.node) { /* .. */ }

const app = require('yargs');
const updateNotifier = require('update-notifier');
const pkg = require('../package.json');
```

I use MacOS, `node@10.10.0`, `npm@6.4.1`, `eslint@5.13.0`, `eslint-config-airbnb-base@13.1.0`

aboyton:
This fixes a bug with moving around variable definitions.

Fixes #2002 and  #1980

### DIFF
--- a/packages/eslint-config-airbnb-base/package.json
+++ b/packages/eslint-config-airbnb-base/package.json
@@ -58,14 +58,14 @@
     "editorconfig-tools": "^0.1.1",
     "eslint": "^4.19.1 || ^5.3.0",
     "eslint-find-rules": "^3.3.1",
-    "eslint-plugin-import": "^2.15.0",
+    "eslint-plugin-import": "^2.16.0",
     "in-publish": "^2.0.0",
     "safe-publish-latest": "^1.1.2",
-    "tape": "^4.9.1"
+    "tape": "^4.10.1"
   },
   "peerDependencies": {
     "eslint": "^4.19.1 || ^5.3.0",
-    "eslint-plugin-import": "^2.15.0"
+    "eslint-plugin-import": "^2.16.0"
   },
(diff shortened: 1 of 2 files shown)


ljharb:
Thanks, but `^2.15.0` already includes `v2.16` via the semver range, so this isn't necessary.

aboyton:
That is true, people can always bump the version themselves and it's consistent with the semantic range that you have set (as you recommended in #2002).

But correct me if I'm wrong if people are using `package-lock.json` (or `yarn.lock`) and they've installed a while ago then they will pin to `2.15.0` and `npm ls` won't let them know that there's any issues even though they have a buggy version of `eslint-plugin-import`, where as if you bump the version and people upgrade `eslint-config-airbnb` then it will be more obvious to them. Honest question, I'm curious what's better here from a library standpoint.

ljharb:
By using a lockfile, you've signed yourself up for running `npm outdated` (or using greenkeeper, or any other means of finding out when there's updates), and manually doing those updates.

I agree that before the next release we'll want to bump the version, but bumping deps is usually done by maintainers as prep for the release process.

aboyton:
Thanks for the explanation. Feel free to either leave this around as a reminder or close it, whatever is most helpful for you.

PR #2012, opened 2019-02-22
Update README.md

yovasx2:
According to the above rule parenthesis must be together the name

### DIFF
--- a/README.md
+++ b/README.md
@@ -675,7 +675,7 @@ Other Style Guides
 
     ```javascript
     // immediately-invoked function expression (IIFE)
-    (function () {
+    (function() {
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -675,7 +675,7 @@ Other Style Guides
 
     ```javascript
     // immediately-invoked function expression (IIFE)
-    (function () {
+    (function() {
```
ljharb: I'm not sure what rule you mean; there must *always* be a single space after the `function` keyword, no exceptions.

Ensive:
@ljharb 
I think you might want to close this PR due to the fact, that whitespace was the only change intended by author

PR #2014, opened 2019-02-24
Add explanations and examples for .bind and arrow functions in render path in react components

### LINKED ISSUE #1995: React Style Guide question: bind and arrow functions in render
https://github.com/airbnb/javascript/tree/master/react#methods

Using `bind ` is bad. Why? `A bind call in the render path creates a brand new function on every single render.` And we should use arrow functions. But, AFAIK, arrow functions also  creates a brand new function on every single render. What I do not understand?

Ensive:
I've deeply researched the topic regarding arrow functions and `.bind` usage within React components (specifically render path). This PR is my proposal to update the javascript styleguide (React page).
Let me know your thoughts.
Shall we split the example with appropriate usage of arrow functions into two:
- within event handler (with focus on the need to pass `event` object to another function)
- within `.map` or other cases (but in this case creating a separate function outside of render path is a better practice)

Any suggestions on editing are welcome.

Closes #1995

### DIFF
--- a/react/README.md
+++ b/react/README.md
@@ -531,7 +531,7 @@ We don’t recommend using indexes for keys if the order of items may change.
 
 ## Methods
 
-  - Use arrow functions to close over local variables.
+  - Use arrow functions to close over local variables. It is handy when you need to pass additional data to an event handler. Although, make sure they [do not massively hurt performance](https://www.bignerdranch.com/blog/choosing-the-best-approach-for-react-event-handlers/), in particular when passed to custom components that might be PureComponents, because they will trigger a possibly needless rerender every time.
 
     ```jsx
     function ItemList(props) {
@@ -540,7 +540,7 @@ We don’t recommend using indexes for keys if the order of items may change.
           {props.items.map((item, index) => (
             <Item
               key={item.key}
-              onClick={() => doSomethingWith(item.name, index)}
+              onClick={(event) => doSomethingWith(event, item.name, index)}
             />
           ))}
         </ul>
@@ -550,7 +550,7 @@ We don’t recommend using indexes for keys if the order of items may change.
 
   - Bind event handlers for the render method in the constructor. eslint: [`react/jsx-no-bind`](https://github.com/yannickcr/eslint-plugin-react/blob/master/docs/rules/jsx-no-bind.md)
 
-    > Why? A bind call in the render path creates a brand new function on every single render.
+    > Why? A bind call in the render path creates a brand new function on every single render. Do not use arrow functions in class fields, because it makes them [challenging to test and debug, and can negatively impact performance](https://medium.com/@charpeni/arrow-functions-in-class-properties-might-not-be-as-great-as-we-think-3b3551c440b1), and because conceptually, class fields are for data, not logic.
 
     ```jsx
     // bad
@@ -561,6 +561,17 @@ We don’t recommend using indexes for keys if the order of items may change.
 
       render() {
         return <div onClick={this.onClickDiv.bind(this)} />;
+      }
+    }
+
+    // very bad
+    class extends React.Component {
+      onClickDiv = () => {
+        // do stuff
+      }
+
+      render() {
+        return <div onClick={this.onClickDiv} />
       }
     }
 


react/README.md
```diff
@@ -624,19 +635,19 @@ We don’t recommend using indexes for keys if the order of items may change.
...
-  1. `componentWillReceiveProps`
-  1. `shouldComponentUpdate`
-  1. `componentWillUpdate`
-  1. `componentDidUpdate`
-  1. `componentWillUnmount`
-  1. *clickHandlers or eventHandlers* like `onClickSubmit()` or `onChangeDescription()`
-  1. *getter methods for `render`* like `getSelectReason()` or `getFooterContent()`
-  1. *optional render methods* like `renderNavigation()` or `renderProfilePicture()`
-  1. `render`
+  2. `constructor`
```
Ensive: This change "prettier" does automatically. It looks good to me.
Shall we keep it with this PR? Or delete? Or create a separate PR?

ljharb: Please do not run prettier on any project that doesn't already have it set up.

react/README.md
```diff
@@ -550,7 +550,7 @@ We don’t recommend using indexes for keys if the order of items may change.
 
   - Bind event handlers for the render method in the constructor. eslint: [`react/jsx-no-bind`](https://github.com/yannickcr/eslint-plugin-react/blob/master/docs/rules/jsx-no-bind.md)
 
-    > Why? A bind call in the render path creates a brand new function on every single render.
+    > Why? A bind call in the render path creates a brand new function on every single render. Do not use arrow functions in class fields, because it makes them [challenging to test, debug and negatively impact a performance](https://medium.com/@charpeni/arrow-functions-in-class-properties-might-not-be-as-great-as-we-think-3b3551c440b1).
```
ljharb:
```suggestion
    > Why? A bind call in the render path creates a brand new function on every single render. Do not use arrow functions in class fields, because it makes them [challenging to test, debug, and can negatively impact performance](https://medium.com/@charpeni/arrow-functions-in-class-properties-might-not-be-as-great-as-we-think-3b3551c440b1).
```

react/README.md
ljharb: so yes, please revert all unrelated changes, and please do not run prettier on anything.

Ensive: Got it

react/README.md
Ensive: @ljharb Thanks for the review. Applied a suggested change

react/README.md
```diff
@@ -531,7 +531,7 @@ We don’t recommend using indexes for keys if the order of items may change.
 
 ## Methods
 
-  - Use arrow functions to close over local variables.
+  - Use arrow functions to close over local variables. It is handy when you need to pass additional data to an event handler. Although, make sure they [do not massively hurt performance](https://www.bignerdranch.com/blog/choosing-the-best-approach-for-react-event-handlers/).
```
ljharb:
```suggestion
  - Use arrow functions to close over local variables. It is handy when you need to pass additional data to an event handler. Although, make sure they [do not massively hurt performance](https://www.bignerdranch.com/blog/choosing-the-best-approach-for-react-event-handlers/), in particular when passed to custom components that might be PureComponents, because they will trigger a possibly needless rerender every time.
```

react/README.md
```diff
@@ -550,7 +550,7 @@ We don’t recommend using indexes for keys if the order of items may change.
 
   - Bind event handlers for the render method in the constructor. eslint: [`react/jsx-no-bind`](https://github.com/yannickcr/eslint-plugin-react/blob/master/docs/rules/jsx-no-bind.md)
 
-    > Why? A bind call in the render path creates a brand new function on every single render.
+    > Why? A bind call in the render path creates a brand new function on every single render. Do not use arrow functions in class fields, because it makes them [challenging to test, debug, and can negatively impact performance](https://medium.com/@charpeni/arrow-functions-in-class-properties-might-not-be-as-great-as-we-think-3b3551c440b1).
```
ljharb: One more refinement:
```suggestion
    > Why? A bind call in the render path creates a brand new function on every single render. Do not use arrow functions in class fields, because it makes them [challenging to test and debug, and can negatively impact performance](https://medium.com/@charpeni/arrow-functions-in-class-properties-might-not-be-as-great-as-we-think-3b3551c440b1), and because conceptually, class fields are for data, not logic.
```

Ensive:
@ljharb 
The recent review comments are addressed. Good addition 👍

PR #2022, opened 2019-03-13
[eslint config] [react] Add support for React Hooks

### LINKED ISSUE #1944: Add support for React Hooks
I would suggest to enforce the [Rules of Hooks](https://reactjs.org/docs/hooks-rules.html) for React v16.7.0-alpha and beyond. A linter plugin is already available under the package [eslint-plugin-react-hooks](https://www.npmjs.com/package/eslint-plugin-react-hooks).

wsmd:
Fixes #1944 

Looks like everyone (via #1944) is in favor of adding support for React Hooks!

`eslint-plugin-react-hooks` is stable now and could be included in `eslint-config-airbnb`.

### DIFF
--- a/packages/eslint-config-airbnb/README.md
+++ b/packages/eslint-config-airbnb/README.md
@@ -10,7 +10,7 @@ We export three ESLint configurations for your usage.
 
 ### eslint-config-airbnb
 
-Our default export contains all of our ESLint rules, including ECMAScript 6+ and React. It requires `eslint`, `eslint-plugin-import`, `eslint-plugin-react`, and `eslint-plugin-jsx-a11y`. If you don't need React, see [eslint-config-airbnb-base](https://npmjs.com/eslint-config-airbnb-base).
+Our default export contains all of our ESLint rules, including ECMAScript 6+ and React. It requires `eslint`, `eslint-plugin-import`, `eslint-plugin-react`, `eslint-plugin-react-hooks`, and `eslint-plugin-jsx-a11y`. If you don't need React, see [eslint-config-airbnb-base](https://npmjs.com/eslint-config-airbnb-base).
 
 1. Install the correct versions of each package, which are listed by the command:
 
@@ -39,7 +39,7 @@ Our default export contains all of our ESLint rules, including ECMAScript 6+ and
   Which produces and runs a command like:
 
   ```sh
-  npm install --save-dev eslint-config-airbnb eslint@^#.#.# eslint-plugin-jsx-a11y@^#.#.# eslint-plugin-import@^#.#.# eslint-plugin-react@^#.#.#
+  npm install --save-dev eslint-config-airbnb eslint@^#.#.# eslint-plugin-jsx-a11y@^#.#.# eslint-plugin-import@^#.#.# eslint-plugin-react@^#.#.# eslint-plugin-react-hooks@^#.#.#
   ```
 
   If using **npm < 5**, Windows users can either install all the peer dependencies manually, or use the [install-peerdeps](https://github.com/nathanhleung/install-peerdeps) cli tool.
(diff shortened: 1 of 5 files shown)


packages/eslint-config-airbnb/rules/react.js
```diff
@@ -465,6 +466,14 @@ module.exports = {
     // https://github.com/yannickcr/eslint-plugin-react/blob/bc976b837abeab1dffd90ac6168b746a83fc83cc/docs/rules/jsx-fragments.md
     // TODO: enable, semver-major
     'react/jsx-fragments': ['off', 'syntax'],
+
+    // Enforce Rules of Hooks
+    // https://github.com/facebook/react/blob/f99fca3cb28e1fbd87e0b647ac43959add290be7/packages/eslint-plugin-react-hooks/README.md
+    'react-hooks/rules-of-hooks': 'error',
+
+    // Verify the list of the dependencies for Hooks like useEffect and similar
+    // https://github.com/facebook/react/blob/f99fca3cb28e1fbd87e0b647ac43959add290be7/packages/eslint-plugin-react-hooks/README.md
```
wsmd: The official repo doesn't have a separate README file for each rule, so I linked to the main one. There's also this link: https://reactjs.org/docs/hooks-rules.html

ljharb:
The addition of a new peer dep is always a breaking change, so there's no avoiding that.

packages/eslint-config-airbnb/rules/react.js
```diff
@@ -465,6 +466,14 @@ module.exports = {
     // https://github.com/yannickcr/eslint-plugin-react/blob/bc976b837abeab1dffd90ac6168b746a83fc83cc/docs/rules/jsx-fragments.md
     // TODO: enable, semver-major
     'react/jsx-fragments': ['off', 'syntax'],
+
+    // Enforce Rules of Hooks
+    // https://github.com/facebook/react/blob/f99fca3cb28e1fbd87e0b647ac43959add290be7/packages/eslint-plugin-react-hooks/README.md
+    'react-hooks/rules-of-hooks': 'error',
```
ljharb: let's add this in a separate entry point; i'm thinking `hooks.js`, so that you can extend `['airbnb', 'airbnb/hooks']`. That way, folks that aren't using hooks or don't want to, don't have to.

packages/eslint-config-airbnb/rules/react.js
ljharb: At the least, each rule should have a permalink - if not, can you file an upstream issue for them to add one?

packages/eslint-config-airbnb/rules/react.js
QuentinRoy: Won't having two entry points add unnecessary complexity? If you do not use hooks, these rules won't affect you. But folks may start using hooks thinking they are covered while they're not.

ljharb: They very well might, since component detection is hard, and functions whose names start with “use” sent something hooks invented.

QuentinRoy: Is it to avoid introducing an additional peer dependency? Or just that airbnb's position is that hooks are bad so you would rather ignore them by default?

ljharb: No, we'd still require the additional peer dependency regardless - there's only one package.json per package.

Airbnb definitely won't be recommending against using hooks; but that doesn't mean there's no caveats with them.

QuentinRoy: That's what I figured. I suppose you could still not add the peer and document that it is required in this particular condition. But it looked like a terrible idea.
To be honest, I still don't quite understand why it is best to keep these rules as a separated config in this case. But thanks for the answer.

wsmd: > To be honest, I still don't quite understand why it is best to keep these rules as a separated config in this case.

I can see both sides of the argument, but I also see no harm in making them an opt-in. If that's not the case, the individual rules can still be disabled if people choose not to use hooks, but... that's just unnecessary noise and it requires people to maintain their config if more hook rules were added in the future.

>  If you do not use hooks, these rules won't affect you. But folks may start using hooks thinking they are covered while they're not.

I don't necessarily disagree, but worth mentioning that I found a couple of places where methods prefixed with `use` (that weren't hooks) in a codebase I'm working on. It's possibly an edge case, but it could happen! The rules only use the function name to decide if it is a hook.

For example, with `eslint-plugin-react-hooks` added, the following file will throw an error even though I'm not using React at all:

```js
function useSomething() {}

export function doSomething() {
  if (window.totallyNotAReactComponent) {
    useSomething();
    ~~~~~~~~~~~~
  }
}
```

packages/eslint-config-airbnb/rules/react.js
wsmd: I can get a permalink to the actual rule implementation. Does that work?

There aren't any dedicated READMEs for each one.

KamuelaFranco:
What's still outstanding for this PR?

PR #2031, opened 2019-05-01
[guide] Fix description of rule 23.8 to specify `export-default`

DannyNemer:
Fixes the description of rule 23.8 to specify naming files in "PascalCase when you export-**default** a constructor", not when you merely export a constructor (i.e., among multiple exports).

### DIFF
--- a/README.md
+++ b/README.md
@@ -3308,7 +3308,7 @@ Other Style Guides
     ```
 
   <a name="naming--PascalCase-singleton"></a><a name="22.8"></a>
-  - [23.8](#naming--PascalCase-singleton) Use PascalCase when you export a constructor / class / singleton / function library / bare object.
+  - [23.8](#naming--PascalCase-singleton) Use PascalCase when you export-default a constructor / class / singleton / function library / bare object.
 
     ```javascript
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -3308,7 +3308,7 @@ Other Style Guides
     ```
 
   <a name="naming--PascalCase-singleton"></a><a name="22.8"></a>
-  - [23.8](#naming--PascalCase-singleton) Use PascalCase when you export a constructor / class / singleton / function library / bare object.
+  - [23.8](#naming--PascalCase-singleton) Use PascalCase when you export-default a constructor / class / singleton / function library / bare object.
```
ljharb: This is an incorrect change; PascalCase *always* applies to these cases, whether named or default exported.

DannyNemer: If a file has multiple exports, one of which is one of these cases (e.g., class), then that file's name should still match that single export? What if you export multiple classes?

DannyNemer: Pardon me! I thought this rule was referring to naming the file when you export-default a class. The two rules immediately before refer to naming the file in such cases, and the example uses `export-default`. My mistake!

image

PR #2056, opened 2019-07-06
Update ref section

uxitten:
I think we don't need ```{ }``` because it's a single line 
Updated method name from ```onClick``` to ```handleClick```
Removed unnecessary ```(event) =>``` when we have one argument ```event =>```
Importing ```import React from 'react'``` to ```import React, { Component } from 'react'```

### DIFF
--- a/react/README.md
+++ b/react/README.md
@@ -540,7 +540,7 @@ We don’t recommend using indexes for keys if the order of items may change.
           {props.items.map((item, index) => (
             <Item
               key={item.key}
-              onClick={(event) => doSomethingWith(event, item.name, index)}
+              onClick={(event) => { doSomethingWith(event, item.name, index); }}
             />
           ))}
         </ul>


react/README.md
```diff
@@ -470,7 +470,7 @@ We don’t recommend using indexes for keys if the order of items may change.
 
     // good
     <Foo
-      ref={(ref) => { this.myRef = ref; }}
+      ref={ref => this.myRef = ref}
```
ljharb: in fact, we definitely still need it - this guide strictly forbids using assignments in expression position.

react/README.md
```diff
@@ -644,15 +644,15 @@ We don’t recommend using indexes for keys if the order of items may change.
...
+  1. *clickHandlers or eventHandlers* like `handleClickSubmit()` or `handleChangeDescription()`
   1. *getter methods for `render`* like `getSelectReason()` or `getFooterContent()`
   1. *optional render methods* like `renderNavigation()` or `renderProfilePicture()`
   1. `render`
 
   - How to define `propTypes`, `defaultProps`, `contextTypes`, etc...
 
     ```jsx
-    import React from 'react';
+    import React, { Component } from 'react';
```
ljharb: please revert this; `React.Component` is the preferred way to reference it.

uxitten: Yes, I agree with you

PR #2067, opened 2019-08-12
Minor change from condition z to y

utsavkarkee:
Minor change from condition if(z) return y to
if (y) return y
just for better semantics, in my opinion.

Never mind if that was intended.


ljharb:
Since both the before and the after example don't convey any semantics, I'm not sure why this change is useful?

PR #2083, opened 2019-08-30
eslint-config-airbnb: allow eslint-plugin-react-hooks 2

christophehurpeau:
This allows to install eslint-plugin-react-hooks 2 by adding it in peerDependencies


ljharb:
What are the rule changes that warranted a v2?

PR #2086, opened 2019-09-05
adding jumpcut to In The Wild section

julianjca11:


ljharb:
Permissions seemed a bit messed up here. Looks like you replaced this with #2089?

PR #2106, opened 2019-10-03
Enforce rule 7.9 in eslint (default-param-last)

vieira:
Enforces the existing [airbnb style guide rule about default parameters coming last](https://github.com/airbnb/javascript#functions--defaults-last) in eslint by using [default-param-last](https://eslint.org/docs/rules/default-param-last).

As this rule was introduced in ESLint 6.4.0, I bumped the eslint version in `package.json` but I am not sure if that is the right thing to do? This is my first PR in this repository. 😨

### DIFF
--- a/packages/eslint-config-airbnb-base/package.json
+++ b/packages/eslint-config-airbnb-base/package.json
@@ -57,7 +57,7 @@
     "babel-preset-airbnb": "^4.0.1",
     "babel-tape-runner": "^3.0.0",
     "eclint": "^2.8.1",
-    "eslint": "^5.16.0 || ^6.1.0",
+    "eslint": "^5.16.0 || ^6.4.0",
     "eslint-find-rules": "^3.4.0",
     "eslint-plugin-import": "^2.18.2",
     "in-publish": "^2.0.0",
--- a/packages/eslint-config-airbnb-base/package.json
+++ b/packages/eslint-config-airbnb-base/package.json
@@ -65,7 +65,7 @@
     "tape": "^4.11.0"
   },
   "peerDependencies": {
-    "eslint": "^5.16.0 || ^6.1.0",
+    "eslint": "^5.16.0 || ^6.4.0",
     "eslint-plugin-import": "^2.18.2"
   },
   "engines": {
--- a/packages/eslint-config-airbnb-base/rules/style.js
+++ b/packages/eslint-config-airbnb-base/rules/style.js
@@ -73,6 +73,9 @@ module.exports = {
     // enforces consistent naming when capturing the current execution context
     'consistent-this': 'off',
 
+    // enforce that default parameters should come last
+    'default-param-last': ['error'],
+
     // enforce newline at the end of file, with no multiple empty lines
     'eol-last': ['error', 'always'],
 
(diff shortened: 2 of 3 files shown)


packages/eslint-config-airbnb-base/package.json
```diff
@@ -57,7 +57,7 @@
     "babel-preset-airbnb": "^4.0.1",
     "babel-tape-runner": "^3.0.0",
     "eclint": "^2.8.1",
-    "eslint": "^5.16.0 || ^6.1.0",
+    "eslint": "^5.16.0 || ^6.4.0",
```
ljharb: this change would need to be made in both dev deps and peerDeps; it also is arguably a breaking change, but I've typically treated this as semver-minor in the past.

packages/eslint-config-airbnb-base/rules/style.js
```diff
@@ -73,6 +73,9 @@ module.exports = {
     // enforces consistent naming when capturing the current execution context
     'consistent-this': 'off',
 
+    // enforce that default parameters should come last
+    'default-param-last': ['error'],
```
ljharb: please update the [guide section](https://github.com/airbnb/javascript#functions--defaults-last) to reference the eslint rule.

This change is semver-minor, since it's enforcing something already in the guide.

ljharb (changes requested):

packages/eslint-config-airbnb-base/rules/style.js
vieira: Hello @ljharb: are you referring to something in addition to the change in [README.md](https://github.com/airbnb/javascript/pull/2106/files#diff-04c6e90faac2675aa89e2176d2eec7d8) I made?

ljharb: nope, that covers it! (sorry if i missed it the first time)

packages/eslint-config-airbnb-base/package.json
vieira: @ljharb Done, thanks for reviewing!

ljharb:
I just realized that since this rule isn't in eslint 5, this would in fact be a breaking change - so it can't come in enabled just yet.

vieira:
Ahh I see, sorry, should I drop eslint 5 from the dependencies and you will merge this sometime in the future or is there other way of doing things?

ljharb:
Typically I add all new rules as "off" as I update eslint versions, and leave a TODO comment to enable them - but they are not enabled until the next major release.

vieira:
@ljharb Understood, let me know if you need anything else from me!

ljharb:
Ideally this can stay open until we’re ready to do a semver-major.

vieira:
Can this be merged or closed now?

vieira:
Can this be merged or closed now?

ljharb:
@vieira we're still not ready to do a semver-major, so i'd prefer to leave it open.

ljharb:
Oops, good catch. I should have included this in that release. you're right; i'll merge this later today.

ljharb:
Done in f3d3a075c

PR #2111, opened 2019-10-05
remove redundant dependencies `object.assign`

1pete:
We are now not supporting nodejs v4.
So we can now use builtin `Object.assign` and module `object.assign` is no longer needed.

PS. I have to disable `prefer-object-spread` because object spreading is not supported in nodejs v6 which we have supported

### DIFF
--- a/packages/eslint-config-airbnb-base/.eslintrc
+++ b/packages/eslint-config-airbnb-base/.eslintrc
@@ -6,5 +6,7 @@
     "comma-dangle": 0,
     // we support node 4
     "prefer-destructuring": 0,
+    // we support node 6
+    "prefer-object-spread": 0,
   },
 }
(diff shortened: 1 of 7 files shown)


ljharb:
What’s the point of this change? This is a dev dep.

Also, although i haven’t tested in awhile, it’s possible that node 4 still has a broken Object.assign. Using the package avoids that.

1pete:
It is a direct dependency of config packages.

And we no longer support nodejs v4, do we?
https://github.com/airbnb/javascript/blob/ab72ab9e90f403e90590f8815e0563248af10470/packages/eslint-config-airbnb-base/package.json#L71-L73

ljharb:
That is true, but airbnb's babel preset currently transpiles object spreads to use this dep, and I don't think there's enough benefit to outweigh the risk of breakage.

PR #2125, opened 2019-10-29
Enforce modern usage of Promises

kripod:
As `async`/`await` syntax is spreading across the ecosystem with native support from [Node.js](https://node.green/), [browsers](https://caniuse.com/#feat=async-functions) and transpilers, they may be preferred over Promises.


ljharb:
Separately, no, the syntax is largely not preferred over Promises. await is very easy to misuse.

PR #2133, opened 2019-11-09
update with master

roygopal:


ljharb:
This seems like it was meant for your fork, but would you be interested in a fuller review and inclusion here?

PR #2136, opened 2019-11-19
Change jsx-a11y/label-has-associated-control from both to either

guoyunhe:
According to the [documentation](https://github.com/evcohen/eslint-plugin-jsx-a11y/blob/master/docs/rules/label-has-associated-control.md), **either** `htmlFor` or `nesting` should work. Enforcing `both` will make developers' life much harder. In many designs, you need detached label and input elements.

### DIFF
--- a/packages/eslint-config-airbnb/rules/react-a11y.js
+++ b/packages/eslint-config-airbnb/rules/react-a11y.js
@@ -58,12 +58,11 @@ module.exports = {
     }],
 
     // Enforce that a label tag has a text label and an associated control.
-    // https://github.com/evcohen/eslint-plugin-jsx-a11y/blob/b800f40a2a69ad48015ae9226fbe879f946757ed/docs/rules/label-has-associated-control.md
+    // https://github.com/evcohen/eslint-plugin-jsx-a11y/blob/master/docs/rules/label-has-associated-control.md
     'jsx-a11y/label-has-associated-control': ['error', {
       labelComponents: [],
       labelAttributes: [],
       controlComponents: [],
-      assert: 'both',
       depth: 25
     }],
 


ljharb:
Both is necessary, as there are devices that only recognize one or the other - Dragon for example requires linking, many other devices nesting.

To be truly accessible - to every single device, no exceptions - you must use both.

guoyunhe:
But you need to provide evidence: screencast or some other reports. I have searched through the internet and here is nothing support this opinion.

ljharb:
a) no, even the merest *possibility* that something won't be accessible makes it worth the restriction
b) https://www.levelaccess.com/dragon-13-aria-support-is-here/ See "Dragon 13 still does not appear to support implicit labels. "

guoyunhe:
So here is evidence that some device/software only support `htmlFor`. But is here any devices/software **only** working with `nesting`? If no, then `htmlFor` is enough.

For any components, applying an `htmlFor` attribute is easy. But `nesting` is impossible in many cases.

ljharb:
Yes, quite a number of devices only support nesting (since that is the modern practice).

In which cases is nesting impossible?

guoyunhe:
[This example](http://toolkit.adl.3stripes.net/adidas/components/forms/inputs/design#default) uses a floating label over the input, when focusing on the input, it will raise to top.

image

image

ljharb:
That example is abusing placeholders, which are only for example input, not for labels - do you have any examples that aren’t bad practices (like infield labels) for which nesting is impossible?

guoyunhe:
Here are two most starred React UI library on GitHub:

## [Material UI](https://material-ui.com/components/text-fields/)

The label is a shared React component. In the component itself, it cannot nest any inputs. Like this:

```jsx
const MyLabel = (props) => <lable htmlFor={props.htmlFor}>{props.children} {props.required && '*'}</label>
```

## [Ant Design](https://ant.design/components/form/#components-form-demo-register)

This example is a form layout that requires labels and inputs to be in different columns.

----

I still didn't see any evidence that any device/software doesn't support `htmlFor`. It is the most used way to link labels and inputs. Can you provide any links?

ljharb:
(also, https://dequeuniversity.com/rules/axe/3.0/label-title-only talks about some more devices that don't support nesting properly by itself, such as JAWS; and https://webaim.org/discussion/mail_thread?thread=7458 and https://www.aaron-gustafson.com/notebook/labeled-with-love/ also discuss many devices and browsers that struggle with nesting alone)

There's UX reasons as well to always do nesting; for checkboxes and radio buttons in particular, it increases the clickable/tappable area, making your page easier to use, especially on mobile.

At the moment, I'm not coming up with links that explain why for *every* kind of form input, for-ID linking is not sufficient, but even your examples don't hold up. (See https://www.aaron-gustafson.com/notebook/labeled-with-love/#placeholders-arent-labels for how to do the infield label, properly, without any JS; the material design one absolutely can take an input as a child, it's just that the linter rule would have to be disabled for that case; and the ant design one imo is a flaw in the design itself)

guoyunhe:
So you cannot provide evidence of why `htmlFor` is insufficient. Then why you insist `both`?

ljharb:
That I can't provide evidence with some brief googling now doesn't mean none exists - lack of evidence is not evidence of absence. Historically, both were required. For *accessibility* - for which what is paramount is "no user is excluded", developer pain is *always* a worthy tradeoff for including humans. In this case, doing both is not difficult, and I believe is always possible when avoiding bad practices, so *no evidence is required* because "both" is the proper ethical choice.

guoyunhe:
You claim "many devices" early. But now you say "no evidence is required"...

ljharb:
@guoyunhe historically, many devices. i can't find supporting evidence with the brief time i've spent googling for it - but my overarching point remains.

samuraijane:
I agree with @guoyunhe. One use case is hiding the inputs for a group of radio buttons and using the adjacent sibling selector to make the labels clickable.

That being said, I think this error message is misleading if both are truly required.

`A form label must be associated with a control` clearly does not communicate the need for both and adds confusion in cases where the _for-id_ relationship does indeed exist.

hades200082:
One of the most common UI frameworks out there, Bootstrap, uses the `for=""` method and does not use nesting in any of their documented examples.

https://getbootstrap.com/docs/4.5/components/forms/

From their documentation... https://getbootstrap.com/docs/4.5/getting-started/accessibility/

> The overall accessibility of any project built with Bootstrap depends in large part on the author’s markup, additional styling, and scripting they’ve included. However, provided that these have been implemented correctly, it should be perfectly possible to create websites and applications with Bootstrap that fulfill WCAG 2.0 (A/AA/AAA), Section 508 and similar accessibility standards and requirements.

This seems to indicate that both are indeed NOT required when using the `for=""` form and that you can still achieve a highly accessible website without nesting.

ljharb:
@hades200082 on the contrary; that just indicates that bootstrap isn't concerned with the devices and human beings (hearing-impaired or otherwise) that excludes - including the latest version of Dragon Naturally Speaking, a commonly used tool.

joshferrell:
@ljharb Ran into this issue for my inputs. An example for when implicit doesn't work is if you put the description and error messages above inputs. This is recommended in the [UK Gov Design System](https://design-system.service.gov.uk/components/text-input/number-input/index.html). See [Adrian Roselli's article](https://adrianroselli.com/2017/01/avoid-messages-under-fields.html) concerning why placing text under inputs is problematic.

The html for this looks something like this:
```html
<div class="govuk-form-group">
  <label class="govuk-label" for="account-number">
    Account number
  </label>
  <span id="account-number-hint" class="govuk-hint">
    Must be between 6 and 8 digits long
  </span>
  <input class="govuk-input govuk-input--width-10" id="account-number" name="account-number" type="text" aria-describedby="account-number-hint" pattern="[0-9]*" inputmode="numeric" spellcheck="false">
</div>
```

It would be invalid semantically to wrap the description in the label as it's not the label, but a description of the input.

TeChn4K:
> @hades200082 on the contrary; that just indicates that bootstrap isn't concerned with the devices and human beings (hearing-impaired or otherwise) that excludes - including the latest version of Dragon Naturally Speaking, a commonly used tool.

Would it be possible to think that Dragon Naturally Speaking does not support this very old specification?
Every modern browser on this planet support either `for` or `nest`. But all developers have to comply to Dragon?

ljharb:
@TeChn4K yes, all developers have to comply with the lowest common denominator. accessibility means *everyone* - no user should be left behind.

guoyunhe:
However, here is no any evidence that Dragon Naturally Speaking doesn't support `for` on the internet, except for @ljharb 's personal knowledge.

I think nobody should force themselves to use `both`. As the W3 standard says, `either` is the right way to go. We won't waste time for something that nobody ever see existing.

ljharb:
You’re welcome to try the latest version yourself and verify it.

The standard is irrelevant; for accessibility, only actual assistive devices matter.

PR #2137, opened 2019-11-20
feat(coercion): add coercion math

wenzi0github:
When rounding data, it is recommended to use the method under "math" for operation of pure numeric data, and "parseInt" for string type numbers, which is more semantic and more in line with the definition of function parameters

### DIFF
--- a/README.md
+++ b/README.md
@@ -3226,6 +3226,20 @@ Other Style Guides
     const hasAge = !!age;
     ```
 
+  <a name="coercion--math"></a>
+  - [22.7](#coercion--math) When rounding data, it is recommended to use methods under `Math` for operation of pure numeric data, and "parseInt" for string type numbers, which is more semantic and more in line with the definition of function parameters
+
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -3205,6 +3205,20 @@ Other Style Guides
...
+
+    ```javascript
+    const num = 12.6;
+    const strNum = '12.6';
+
+    const val = Math.floor(num); // 12
+
+    const val = Math.ceil(num); // 13
+
+    const val = parseInt(strNum, 10); // 12
```
```suggestion
    const val = Number.parseInt(strNum, 10); // 12
```

wenzi0github: yes, Number.parseInt seems more obvious contrast.

wenzi0github:
@ljharb this pull could be merged? thank you!

wenzi0github:
difference `Math.floor` and `Number.parseInt`, Math always manage number type, parseInt always mange string type to number.

PR #2138, opened 2019-11-23
normalize multiline word according to merriam-webster

geo4orce:
"multiline" = 44 matches in 9 files
"multi-line" = 5 matches in 3 files

It seems that the former is preferred.

Also merriam-webster confirms it:
```
$ curl -I https://www.merriam-webster.com/dictionary/multi-line
HTTP/2 301
location: /dictionary/multiline
```

### DIFF
--- a/packages/eslint-config-airbnb-base/rules/best-practices.js
+++ b/packages/eslint-config-airbnb-base/rules/best-practices.js
@@ -43,6 +43,11 @@ module.exports = {
     // https://eslint.org/docs/rules/eqeqeq
     eqeqeq: ['error', 'always', { null: 'ignore' }],
 
+    // Require grouped accessor pairs in object literals and classes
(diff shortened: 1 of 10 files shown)


packages/eslint-config-airbnb-base/rules/best-practices.js
```diff
@@ -23,7 +23,7 @@ module.exports = {
     'consistent-return': 'error',
 
     // specify curly brace conventions for all control statements
-    curly: ['error', 'multi-line'],
+    curly: ['error', 'multi-line'], // multiline
```
geo4orce: this is the only spot where I didn't make the actual change because not sure if it would break anything 😨

ljharb: indeed, it would :-)

PR #2168, opened 2020-02-20
Fix typo in no-multiple-empty-lines rule

### LINKED ISSUE #2140: bug in no-multiple-empty-lines
The following comment:

https://github.com/airbnb/javascript/blob/master/packages/eslint-config-airbnb-base/rules/style.js#L314-L316

implies that there should be no new lines at the beginning of a file. But that is not what the rule is configured to do. Currently, the rule allows one empty line at the beginning of a file, which seems strange. For example:

Dog1.js
```
// This is how my dog class should be, with no empty lines at the beginning of the file
class Dog {}
```

Dog2.js
```

// I accidentally put a new line at the beginning of the file,
// but the Airbnb eslint rules don't care about this and give no errors
class Dog {}
```

Dog3.js
```


// I accidentally put two new lines at the beginning of the file,
// the Airbnb eslint rules care about this and throw an error
class Dog {}
```

It looks like `maxBOF: 1` should be changed to `maxBOF: 0`, correct me if I am wrong.

Luis729:
## Why is the change being made?

This change is made because the Airbnb documentation states to "avoid a
newline at the beginning of files", yet the code does not follow this.

 ## What has changed to address the problem?

This change fixes the `no-multiple-empty-lines` rule by setting max
beginning of file (`maxBOF`) to from 1 to 0.

 ## How was this change tested?

This change was tested with `npm test`.

 ## Related docs

https://github.com/airbnb/javascript#whitespace--no-multiple-empty-lines

Fixes #2140

### DIFF
--- a/packages/eslint-config-airbnb/test/test-react-order.js
+++ b/packages/eslint-config-airbnb/test/test-react-order.js
@@ -24,7 +24,7 @@ function lint(text) {
 }
 
 function wrapComponent(body) {
-  return `
+  return `\
 import React from 'react';
 
 export default class MyComponent extends React.Component {
--- a/packages/eslint-config-airbnb-base/rules/style.js
+++ b/packages/eslint-config-airbnb-base/rules/style.js
@@ -313,7 +313,7 @@ module.exports = {
 
     // disallow multiple empty lines, only one newline at the end, and no new lines at the beginning
     // https://eslint.org/docs/rules/no-multiple-empty-lines
-    'no-multiple-empty-lines': ['error', { max: 2, maxBOF: 1, maxEOF: 0 }],
+    'no-multiple-empty-lines': ['error', { max: 2, maxBOF: 0, maxEOF: 0 }],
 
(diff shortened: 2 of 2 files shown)


Luis729:
I will actually set `maxEOF` back to 0, apparently that's the value its supposed to be (see https://github.com/airbnb/javascript/pull/1794)

Will update PR and commit to reflect on this, as well as fix https://github.com/airbnb/javascript/issues/2140

Zamiell:
What is the status on this PR?

packages/eslint-config-airbnb/test/test-react-order.js
```diff
@@ -24,7 +24,7 @@ function lint(text) {
 }
 
 function wrapComponent(body) {
-  return `
+  return `\
```
ljharb, line 27:
```suggestion
  return `
```
template literals don't need a backslash here.

Zamiell:
Thanks ljharb!

PR #2184, opened 2020-02-28
update: primitives type(add "bigint")

lagagain:
add "bigint" type to the primitives type section.

I believe there is another primitives type in JS(may like symbol is new?).

ref: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Data_structures

# use example:

```js
var bn = 123n
```
(with "n" as suffix)

### DIFF
--- a/README.md
+++ b/README.md
@@ -71,6 +71,7 @@ Other Style Guides
     - `null`
     - `undefined`
     - `symbol`
+    - `bigint`
 
     ```javascript
     const foo = 1;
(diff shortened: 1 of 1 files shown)


ljharb:
Indeed, `bigint` is a new primitive type, and we should add it here.

Note, though, that like `symbol`, `bigint` is not polyfillable and not supported by every browser/engine this guide supports - so neither can be used yet.

lagagain:
> Indeed, `bigint` is a new primitive type, and we should add it here.
> 
> Note, though, that like `symbol`, `bigint` is not polyfillable and not supported by every browser/engine this guide supports - so neither can be used yet.

I add the note.

PR #2186, opened 2020-03-02
update: import/extensions rule documentation

### LINKED ISSUE #2134: "import/extensions" rule is not documented (and I am also wondering about the justification)
Part 1:

The current Airbnb config has the "import/extensions" ESLint rule turned on. However, the reasoning for this is not given on the official style guide, and even the fact that the rule is turned on at all is not documented on the style guide. For example, there should be a blurb added to the [section on modules](https://github.com/airbnb/javascript#modules) that says something along the lines of:

```
// bad
import foo as './foo.js';

// good
import foo as './foo';
```

Question 1 - Can this be added / documented?

---

Part 2:

I searched around for some justification for this rule. In #1633, @ljharb says that 

> You're not supposed to use the extension there. The best practice for file extensions on imported code is to omit them, which is why our config requires that as well.

and:

>  the only time you should use extensions in require/import paths is if you're importing JSON files, or if you've made any nonstandard changes to allow requires of alternate extensions, like .css or .svg or .jpg. With .js, .jsx, or .mjs files, you should always omit the extension.

Question 2 - Why exactly is it best practice that they should be omitted? Is it just so that ES6-style module imports more-closely mimic the old, CommonJS-style imports? If so, that seems like a superficial reason.

And related: Even Mozilla has file extensions in their [official documentation](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Statements/import) for the import statement. So is there disagreement about what is best practice or is Mozilla in error here?

---

Part 3:

In #2121, some users bring up the fact that the "import/extensions" rule **breaks** code written for browsers without any transpilation. @ljharb counters with:

>browsers do not care about extensions whatsoever; only node does.

From a technical standpoint, browsers do not care about file extensions. But the fact of the matter is that on a lot of websites (or even most websites?), JavaScript files are served to the browser from a "public" subpath (or something similar) that maps to an internal file system. So, at least for a lot of websites,  **the file extension is required in the import statement** for ES6-style modules to work.

@ljharb also says:

> Regardless, you'll want to always use a bundler and transpile to CJS for the forseeable future.
> 
> If you make a different choice, you may want to override some of our linter settings, all of which assume you're following the best practice of using babel and a bundler.

I can understand that if most of the people in the JavaScript ecosystem are using a bundler, then the Airbnb default settings should be designed for people bundling. However... bundlers can handle file extensions just fine!

Question 3 - Would it make more sense to flip the rule and always require extensions? That way, any particular piece of JavaScript code will just work in **both** bundled-environments and non-bundled-environments.

---

Part 4:

As a final point here, I'd like to reflect on why Airbnb has decided to enforce ES6-style imports at all. The style guide gives the following justification in the modules section:

> Why? Modules are the future, let’s start using the future now.

But **why** exactly do we need to start using the future now? In my mind, it has to do with transpilation. The hypothetical, more-verbose version of the above sentence might read something like the following:

> Why? In the future, all browsers will support all of the latest JavaScript features. And when that day comes, you might feel the need to convert all of the modules in your application to ES6-style in order to get the code up-to-date. But there is a way to avoid this future workload - simply write your application with ES6-style modules **now** and use transpilation in the meantime until browsers catch up. One day, transpilation will be unnecessary, and your code will just work! Modules are the future, so let’s start using the future now.

If this is indeed the reason why we should start using the future now, then I can totally get on-board with the decision to require ES6-style imports as a ESLint rule - being future ready seems like a good thing. But if the end goal is to make transpilation unnecessary, then enforcing the "import/extensions" rule seems like a step in the **opposite direction**, because it forces code to be transpiled for perpetuity.

cain:
Add documentation for import extension rules https://github.com/airbnb/javascript/blob/master/packages/eslint-config-airbnb-base/rules/imports.js#L138

Let me know if example needs changing or added anywhere else :) 
```
    // bad
    import foo from './foo.js';
    import bar from './bar.jsx';

    // good
    import foo from './foo';
    import bar from './bar';
```

Fixes #2134

### DIFF
--- a/README.md
+++ b/README.md
@@ -1441,6 +1441,23 @@ Other Style Guides
     import barCss from 'bar.css';
     ```
 
+  <a name="modules--import-extensions"></a>
+  - [10.10](#modules--import-extensions) Do not include JavaScript filename extensions
+ eslint: [`import/extensions`](https://github.com/benmosher/eslint-plugin-import/blob/master/docs/rules/extensions.md)
+    > Why? Including extensions inhibits refactoring, and inappropriately hardcodes implementation details of the module you're importing in every consumer.
+
+    ```javascript
+    // bad
+    import foo from './foo.js';
+    import bar from './bar.jsx';
+    import baz from './baz/index.jsx';
+
+    // good
+    import foo from './foo';
+    import bar from './bar';
+    import baz from './baz';
+    ```
+
 **[⬆ back to top](#table-of-contents)**
 
 ## Iterators and Generators


cain:
Also curious why build is breaking?

TriMoon:
@cain, 
First off; I'm not using this software, was just following it.
That said, i have a note to you:
- `import` statements need to know from which file to import.
- `import x as y` is meant to import `identifier x` as `identifier y`, but you did not specify a sourcefile to import the `identifier x` from.
- The syntactically correct use is:
   - `import foo as bar from './foo';`
   - or `import foo as bar from './foo.js';`

README.md
```diff
@@ -1441,6 +1441,21 @@ Other Style Guides
     import barCss from 'bar.css';
     ```
 
+  <a name="modules--import-extensions"></a>
+  - [10.10](#modules--import-extensions) Do not include javascript filename extensions
```
TriMoon: See [the build logs](https://travis-ci.org/airbnb/javascript/jobs/657142404#L304), where it says:
```
README.md:1445 MD009/no-trailing-spaces Trailing spaces [Expected: 0; Actual: 1]
```

cain:
> * `import` statements need to know from which file to import.
> * `import x as y` is meant to import `identifier x` as `identifier y`, but you did not specify a sourcefile to import the `identifier x` from.
> * The syntactically correct use is:
>   
>   * `import foo as bar from './foo';`
>   * or `import foo as bar from './foo.js';`
> 
> > Also curious why build is breaking?
> 
> See [the build logs](https://travis-ci.org/airbnb/javascript/jobs/657142404#L304), where it says:
> 
> ```
> README.md:1445 MD009/no-trailing-spaces Trailing spaces [Expected: 0; Actual: 1]
> ```

🤦‍♂️facepalm.

Thanks so much for your help @TriMoon ill fix it up :)

cain:
@TriMoon I think I actually just copy pasted the wrong example into the issue comment, but the source code uses `from` not `as`

README.md
```diff
@@ -1441,6 +1441,21 @@ Other Style Guides
...
```
```suggestion
  - [10.10](#modules--import-extensions) Do not include JavaScript filename extensions
```

README.md
```diff
@@ -1441,6 +1441,21 @@ Other Style Guides
...
+ eslint: [`import/extensions`](https://github.com/benmosher/eslint-plugin-import/blob/master/docs/rules/extensions.md)
+    > Why? Ensure consistent use of file extension within the import path`.
```
```suggestion
    > Why? Including extensions inhibits refactoring, and inappropriately hardcodes implementation details of the module you're importing in every consumer.`.
```

README.md
```diff
@@ -1441,6 +1441,21 @@ Other Style Guides
...
+
+    ```javascript
+    // bad
+    import foo from './foo.js';
+    import bar from './bar.jsx';
+
+    // good
+    import foo from './foo';
+    import bar from './bar';
+    ```
```
```suggestion
    // bad
    import foo from './foo.js';
    import bar from './bar.jsx';
    import baz from './baz/index.jsx';

    // good
    import foo from './foo';
    import bar from './bar';
    import baz from './baz';
    ```
```

README.md
```diff
@@ -1441,6 +1441,23 @@ Other Style Guides
     import barCss from 'bar.css';
     ```
 
+  <a name="modules--import-extensions"></a>
+  - [10.10](#modules--import-extensions) Do not include Javascript filename extensions
```
```suggestion
  - [10.10](#modules--import-extensions) Do not include JavaScript filename extensions
```

README.md
```diff
@@ -1441,6 +1441,23 @@ Other Style Guides
...
+ eslint: [`import/extensions`](https://github.com/benmosher/eslint-plugin-import/blob/master/docs/rules/extensions.md)
+    > Why? Including extensions inhibits refactoring, and inappropriately hardcodes implementation details of the module you're importing in every consumer.`.
```
```suggestion
    > Why? Including extensions inhibits refactoring, and inappropriately hardcodes implementation details of the module you're importing in every consumer.
```

PR #2188, opened 2020-03-02
style: Update Arrays 4.7 to add best style rule

### LINKED ISSUE #2187: Add best conditional rule to Arrays 4.7
Add "best" style guide rule to Arrays [4.7](https://github.com/airbnb/javascript#arrays--callback-return) 
Currently, this is written on the documentation 
```
// good
inbox.filter((msg) => {
  const { subject, author } = msg;
  if (subject === 'Mockingbird') {
    return author === 'Harper Lee';
  }

  return false;
});
```

I think it would be better combining conditionals based on [Eslint Operator Line break](https://eslint.org/docs/rules/operator-linebreak)  and [no mixed operators](https://eslint.org/docs/2.13.1/rules/no-mixed-operators) 

So a simple && or || will be efficient

umairx97:
Closes #2187

### DIFF
--- a/README.md
+++ b/README.md
@@ -450,6 +450,12 @@ Other Style Guides
 
       return false;
     });
+
+    // best
+    inbox.filter((msg) => {
+      const { subject, author } = msg;
+      return subject === 'Mockingbird' && author === 'Harper Lee');
+    });
     ```
 
   <a name="arrays--bracket-newline"></a>


umairx97:
The example I added was on the assumption of the good case rule where there are just two conditions to be checked

ljharb:
and in the case of this *specific* code, I agree your "best" is better :-) however, adding that would be irrelevant to demonstrating the point of this section in the guide, which is to talk about array callback return patterns.

umairx97:
so would you consider adding this to this section ?  
```inbox.filter(({ subject, author }) => subject === 'Mockingbird' && author === 'Harper Lee';```

as it is implicit return pattern

ljharb:
I'm not convinced its presence would be helpful; i think it would distract from the purpose of the section.

umairx97:
Oh okay, understandable

PR #2201, opened 2020-03-23
fix example

roshal:


README.md
```diff
@@ -957,8 +957,7 @@ Other Style Guides
     ```javascript
     // bad
     [1, 2, 3].map((number) => {
```
ljharb: this change would get rid of the entire point of showing this bad example.

roshal: no

without return this function is invalid

grim--20-03-26--04-41-44

returning a single expression in the function body violates the rule

ljharb: That is the *whole point* - that it is an invalid example according to the guide (although it's perfectly valid javascript without the return). See the `// bad` above it?

PR #2203, opened 2020-03-28
Without inicjalized value for acc this code will crash

adrianprzybyla:
Error: Cannot read property 'concat' of undefined

### DIFF
--- a/README.md
+++ b/README.md
@@ -424,13 +424,13 @@ Other Style Guides
     // bad - no returned value means `acc` becomes undefined after the first iteration
     [[0, 1], [2, 3], [4, 5]].reduce((acc, item, index) => {
       const flatten = acc.concat(item);
-    });
+    }, []);
 
     // good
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -424,13 +424,13 @@ Other Style Guides
     // bad - no returned value means `acc` becomes undefined after the first iteration
     [[0, 1], [2, 3], [4, 5]].reduce((acc, item, index) => {
       const flatten = acc.concat(item);
-    });
+    }, []);
```
ljharb: this one is supposed to crash; it's a "bad" example.

README.md
```diff
@@ -424,13 +424,13 @@ Other Style Guides
...
 
     // good
     [[0, 1], [2, 3], [4, 5]].reduce((acc, item, index) => {
       const flatten = acc.concat(item);
       return flatten;
-    });
+    }, []);
```
ljharb: this isn't needed, because a new array is returned each time, and when you omit the initial value, the first value is used.

PR #2207, opened 2020-04-09
[eslint config] [patch] Loosen allowed version range for `eslint-plugin-react-hooks`

kripod:
As seen by [comparing changes](https://github.com/facebook/react/commits/master/packages/eslint-plugin-react-hooks) between `eslint-plugin-react-hooks@2.4.0` and `3.0.0`:

image

The only breaking change seems to be https://github.com/facebook/react/commit/d3368beeecb2b51f51814f95f977e8c11900f54e. Also, a notable change is that a recommended configuration [has been added](https://github.com/facebook/react/commit/79a25125b19c2d6deaf03d4ab0dfdd6082377d97):

```js
export const configs = {
  recommended: {
    plugins: ['react-hooks'],
    rules: {
      'react-hooks/rules-of-hooks': 'error',
      'react-hooks/exhaustive-deps': 'warn',
    },
  },
};
```

It's almost identical to [the current rules in place](https://github.com/airbnb/javascript/blob/06b3ab11d9a443ff46f052dd00375e271e5146e6/packages/eslint-config-airbnb/rules/react-hooks.js#L12-L20), as applicable by extending `airbnb/hooks`. The list of available rules hasn't changed since v2, hence no additional changes should be required.


kripod:
I think that v4 should also be allowed, as it contains no API changes: https://github.com/facebook/react/blob/master/packages/eslint-plugin-react-hooks/CHANGELOG.md

packages/eslint-config-airbnb/package.json
```diff
@@ -68,7 +68,7 @@
     "eslint-plugin-import": "^2.20.1",
     "eslint-plugin-jsx-a11y": "^6.2.3",
     "eslint-plugin-react": "^7.19.0",
-    "eslint-plugin-react-hooks": "^2.3.0 || ^1.7.0",
+    "eslint-plugin-react-hooks": "^4.0.0 || ^3.0.0 || ^2.3.0 || ^1.7.0",
```
ljharb, line 71:
```suggestion
    "eslint-plugin-react-hooks": "^4 || ^3 || ^2.3.0 || ^1.7.0",
```
the same line must be duplicated in peerDependencies - otherwise you haven't actually allowed anything :-)

kripod: Oops, that's right 😅 Thank you for correcting my mistake!

leonardofaria:
Hi folks
Thanks for this project ❤️ 

Out of curiosity, I noticed that this PR got merged but is this going to be release now or with the eslint 7 support (https://github.com/airbnb/javascript/issues/2223)? 

Thanks again

ljharb:
It will be released along with #2223.

Issue #2223, opened 2020-05-07
eslint v7 support


ljharb:
Just like #2036, #1834, #1447, I'm filing this in advance.

eslint v7 was released on 5/8.

The current plan is, if possible, to release a new semver-minor of both configs, that `5 || 6 || 7`.

The timeframe for eslint v7 support is "probably not as soon as you would like, but as soon as we can".
Please accept this, and please add no noise to this thread beyond emoji reactions - distractions won't help anyone get the work done any faster.

### **Please** do not file additional issues and do not file pull requests attempting to add support - please, instead, continue to use eslint v6 until such time as our configs support v7.

Peer Dep eslint combined v5, v6, and v7 support released:
 - [x] [`eslint-plugin-import`](https://github.com/benmosher/eslint-plugin-import/issues/1772)
    - [x] [commit](https://github.com/benmosher/eslint-plugin-import/commit/6a110dd16a7fd775f08601054bf14ffd503eea7b)
    - [x] Release ([PR](https://github.com/benmosher/eslint-plugin-import/pull/1804))
 - [x] [`eslint-find-rules`](https://github.com/sarbbottam/eslint-find-rules/issues/318)
   - [x] [PR](https://github.com/sarbbottam/eslint-find-rules/pull/319)
   - [x] [Release](https://github.com/sarbbottam/eslint-find-rules/pull/321)
 - [x] update the base config
 - [x] publish the base config, semver-minor
 - [x] `eslint-plugin-react`
   - [x] [PR](https://github.com/yannickcr/eslint-plugin-react/pull/2635)
   - [x] [Release](https://github.com/yannickcr/eslint-plugin-react/releases/tag/v7.20.0)
 - [x] [`eslint-plugin-jsx-a11y`](https://github.com/evcohen/eslint-plugin-jsx-a11y/issues/683)
   - [x] [PR](https://github.com/jsx-eslint/eslint-plugin-jsx-a11y/pull/684) (blocked on ~base config~ and [jsx-ast-utils](https://github.com/jsx-eslint/jsx-ast-utils/pull/97))
   - [x] [Release](https://github.com/jsx-eslint/eslint-plugin-jsx-a11y/pull/702)
 - [x] [`eslint-plugin-react-hooks`](https://github.com/facebook/react/issues/18876)
   - [x] [PR](https://github.com/facebook/react/pull/18878)
[...]

PR #2238, opened 2020-06-06
Fix no-multiple-empty-lines rule

vladshcherbin:
From readme:

> 19.20 Avoid multiple empty lines, only allow one newline at the end of files, and avoid a newline at the beginning of files. eslint: no-multiple-empty-lines

Which has a *bad* example:

```js
// bad - multiple empty lines
var x = 1;


var y = 2;
```

Currently, max empty lines [is set to 2](https://github.com/airbnb/javascript/blob/eslint-config-airbnb-v18.1.0/packages/eslint-config-airbnb-base/rules/style.js#L316) and above example passes as a correct one. Same goes for other useless empty lines - between function declarations, etc.

### DIFF
--- a/packages/eslint-config-airbnb-base/rules/style.js
+++ b/packages/eslint-config-airbnb-base/rules/style.js
@@ -313,7 +313,7 @@ module.exports = {
 
     // disallow multiple empty lines, only one newline at the end, and no new lines at the beginning
     // https://eslint.org/docs/rules/no-multiple-empty-lines
-    'no-multiple-empty-lines': ['error', { max: 2, maxBOF: 0, maxEOF: 0 }],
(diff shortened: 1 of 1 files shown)


vladshcherbin:
Since it's kinda breaking change it would be great to ship this with next major version, [with eslint v7](https://github.com/airbnb/javascript/issues/2223) support I believe.

ljharb:
This would be considered semver-minor, since the guide already specifies "no multiple empty lines".

I can't recall why this was set to 2 originally - git [suggests](https://github.com/airbnb/javascript/commit/daea8a638f1f42cd609f1b68b748ad323be31c10#diff-4876eba74c26fa01c8fe78da0a261f16R52) it was not intentional, and just matches the default.

PR #2244, opened 2020-06-19
Changed bad/good example for Array.from

chrishrtmn:


README.md
```diff
@@ -377,10 +377,10 @@ Other Style Guides
     ```javascript
     const foo = document.querySelectorAll('.foo');
 
-    // good
+    // bad
     const nodes = Array.from(foo);
```
ljharb, line 381: this is a good example - use of `Array.from` is very good, because of its mapper argument.

Issue #2245, opened 2020-06-19
latest version of eslint-config-airbnb is not compatible with eslint 7.3.0


nescalante:
when using latest version of eslint with latest version of eslint-config-airbnb is throwing the following error:

```
Oops! Something went wrong! :(

ESLint: 7.3.0

Error: .eslintrc.yml » eslint-config-airbnb » /blablabla/node_modules/eslint-config-airbnb-base/index.js » /blablabla/node_modules/eslint-config-airbnb-base/rules/imports.js:
	Configuration for rule "import/no-cycle" is invalid:
	Value null should be integer.
```

It seems that when parsing the rule configuration, eslint now receives:

```js
[ 'error', { maxDepth: null, ignoreExternal: false } ]
```

Removing `maxDepth` from [this line](https://github.com/airbnb/javascript/blob/54955410ee53fd82e6caa5879ccd0c413640706a/packages/eslint-config-airbnb-base/rules/imports.js#L236) fixes the issue.

brandly:
yes! here's [a build failing as you describe](https://github.com/brandly/react-tetris/runs/788965571)

Methuselah96:
I wonder if it has to do with the fact that JSON stringifying `Inifinity` becomes `null`?

ljharb:
Looks like this is due to an eslint bug: https://github.com/eslint/eslint/issues/13427

I'd suggest downgrading to eslint v7.2 for now.

brandly:
[v7.2 works as expected](https://github.com/brandly/react-tetris/runs/789429956). thanks!

j0k3r:
Fixed by ESLint 7.3.1 https://github.com/eslint/eslint/pull/13435

Knogobert:
They reverted this fix in the new [eslint@7.4.0](https://eslint.org/blog/2020/07/eslint-v7.4.0-released), until they find a nicer solution. So if you get a similar error, try running it in ESLint 7.3.1 for now.

`npm i eslint@7.3.1`

taion:
No, they reverted the original config cloning change, not the band-aid hack in v7.3.1. v7.4.0 restores the pre-v7.3.0 behavior, so things should work just fine here.

PR #2246, opened 2020-06-19
remove import/no-cycle maxDepth

nescalante:
removes import/no-cycle maxDepth since from [its description](https://github.com/benmosher/eslint-plugin-import/blob/v2.21.2/docs/rules/no-cycle.md) it seems to be Infinity by default.

Also fixes https://github.com/airbnb/javascript/issues/2245


nescalante:
got it, closing then :+1:

itajaja:
for the sake of practicality though, couldn't this be merged? It looks like the upstream eslint bug is quite contentious, and it's not going to be fixed anytime soon.

j0k3r:
It’s fixed in ESLint 7.3.1

PR #2249, opened 2020-06-20
bug: import/no-cycle specifies Infinity

philihp:
The configuration of the rule `import/no-cycle` specifies a maxDepth of `infinity`, which while reasonable, causes an error. This can be replicated by checking out the repo, and running `npm run lint`

```bash
/packages/eslint-config-airbnb-base ❯ npm run lint
...
...
...
> eslint-config-airbnb-base@14.2.0 lint /Users/philihp/work/javascript/packages/eslint-config-airbnb-base
> eslint --report-unused-disable-directives .

Oops! Something went wrong! :(

ESLint: 7.3.0

Error: .eslintrc » ./index.js » /Users/philihp/work/javascript/packages/eslint-config-airbnb-base/rules/imports.js:
	Configuration for rule "import/no-cycle" is invalid:
	Value null should be integer.

/packages/eslint-config-airbnb-base ❯ node -v && npm -v
v14.4.0
6.14.5
```

Since `Infinity` [is the default](https://github.com/benmosher/eslint-plugin-import/blob/master/src/rules/no-cycle.js#L35) for the rule, removing this would resolve the error.

### DIFF
--- a/packages/eslint-config-airbnb-base/package.json
+++ b/packages/eslint-config-airbnb-base/package.json
@@ -59,14 +59,14 @@
     "eclint": "^2.8.1",
     "eslint": "^5.16.0 || ^6.8.0 || ^7.2.0",
     "eslint-find-rules": "^3.5.0",
-    "eslint-plugin-import": "^2.21.2",
+    "eslint-plugin-import": "^2.22.0",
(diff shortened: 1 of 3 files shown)


philihp:
Closing, just saw #2246

PR #2284, opened 2020-09-08
Update Outdated URL for Translation

jiahao-c:
The username of the Chinese (Simplified) translator has changed. Github won't auto-redirect the URL. Therefore, the URL needs to be updated.

### DIFF
--- a/react/README.md
+++ b/react/README.md
@@ -740,7 +740,7 @@ We don’t recommend using indexes for keys if the order of items may change.
 
   This JSX/React style guide is also available in other languages:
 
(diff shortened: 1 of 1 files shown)


react/README.md
```diff
@@ -740,7 +740,7 @@ We don’t recommend using indexes for keys if the order of items may change.
 
   This JSX/React style guide is also available in other languages:
 
-  - ![cn](https://raw.githubusercontent.com/gosquared/flags/master/flags/flags/shiny/24/China.png) **Chinese (Simplified)**: [jhcccc/javascript](https://github.com/jhcccc/javascript/tree/master/react)
```
ljharb, line 743: this redirects just fine for me already. does it not for you?

jiahao-c: Oh! At first it didn't work for me. But now it does. Seems like it took Github a few minutes to figure out redirection. Thanks you and I'll close this PR

PR #2304, opened 2020-10-11
[guide] add eslint rule reference for `prefer-object-spread`

1pete:
https://eslint.org/docs/rules/prefer-object-spread


ljharb (approved):
Do you have any more of these? i'd prefer to get them all in one PR instead of bit by bit :-)

1pete:
No, I've not got any more, sorry for the inconvenience. :)

PR #2305, opened 2020-10-13
Update the rule about Avoid duplicate class members in README.md

danielpaz6:
It is unclear from rule [`9.7`](https://github.com/airbnb/javascript#classes--methods-use-this) if overloading is allowed.

Beginners might get the assumption that even overloading isn't allowed since it has the same method name.
Hence, an example showing *overloading* under the `good` comment will solve that possible wrong assumption.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1253,6 +1253,13 @@ Other Style Guides
     class Foo {
       bar() { return 2; }
     }
+    
+    // good
+    class Foo {
+      increase(x) { return x+1; }
+      // overloading is allowed
+      increase(x, y) { retun x+y; }
+    }
     ```
 
   <a name="classes--methods-use-this"></a>
--- a/README.md
+++ b/README.md
@@ -1253,7 +1253,7 @@ Other Style Guides
     class Foo {
       bar() { return 2; }
     }
-    
+
     // good
     class Foo {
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -1253,6 +1253,13 @@ Other Style Guides
     class Foo {
       bar() { return 2; }
     }
+
+    // good
+    class Foo {
+      increase(x) { return x+1; }
+      // overloading is allowed
+      increase(x, y) { return x+y; }
+    }
```
ljharb, line 1262: this isn't allowed, because it's not a feature in JS. The second one will overwrite the first one, and so the duplicate class member will obscure a bug.

ljharb (changes requested):

ljharb:
In other words, overloading *does not exist* in JavaScript, and as such, duplicate class members should *never* be allowed under any circumstances.

README.md
danielpaz6: @ljharb Yeah you are right! I was thinking about TypeScript in my head when I wrote that I guess.

ljharb: In that case, the overloads don't have a function body, and you' want to use the `@typescript-eslint/eslint-plugin`'s rules instead of the core rules.

PR #2306, opened 2020-10-14
[guide] add explanation about why to use radix when using parseInt

danielpaz6:
I added an explanation at [22.3](https://github.com/airbnb/javascript/blob/master/README.md#coercion--numbers) why to use radix once parseInt, this explanation has been taken from [Mozilla's docs](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/parseInt#Octal_interpretations_with_no_radix).

I think it's not clear without an explanation since people can think that it's obvious that string will be parsed to 10 base number, but it's not always the case.


ljharb (approved):
Thanks!

PR #2310, opened 2020-10-16
Better link reference to Drupal's linter

danielpaz6:
This link was taken from their Code standard page: https://www.drupal.org/docs/develop/standards/javascript/eslint-settings

The previous link wasn't very informative and required a download process in order to check their standards.


ljharb (approved):
thanks!

PR #2312, opened 2020-10-17
Amir Abdul Sattar

Amir-ASattar:
I Have added a file with name Amir_JS_Events.html .Using Javascript I have implement form using JS and we can change the background using drop-down menu. Please add me to the repository


ljharb:
Please don’t create spam pull requests.

PR #2315, opened 2020-10-22
[guide] Update reason for preferring object destructuring

### LINKED ISSUE #2293: Update the reason for preferring object destructuring?
The explanation for the preference for destructuring says:
> Destructuring saves you from creating temporary references for those properties.

I've found that devs used to _not_ destructuring find it such a big change that this explanation doesn't seem like a good reason, especially if there's an established habit of referencing object properties inline and not creating temporary references anyways. i.e. rather than:
```js
// bad
function getFullName(user) {
  const firstName = user.firstName;
  const lastName = user.lastName;

  return `${firstName} ${lastName}`;
}
```
They are writing
```js
// better?
function getFullName(user) {
  return `${user.firstName} ${user.lastName}`;
}
```

While I personally find this verbose and repetitive even in this small example, a developer coming from _never_ using destructuring finds it to be far more straightforward and are thus resistant to this rule.

I say all this to ask, are there any additional reasons Airbnb prefers destructuring, and should those be included in the explanation for the rule?

michaelaflores:
Fixes #2293

While this guide isn't intended to provide every possible reason for every preference, it perhaps should aim to provide a succinct and compelling reason for using a certain rule. The current reason for preferring object destructuring is quite narrow in scope. It could be improved to meet this standard, so I'm proposing adding some additional information to clarify the benefits of what is an often controversial rule (controversial only because its introduction can require many changes in a mature codebase and has no auto fix available).

### DIFF
--- a/README.md
+++ b/README.md
@@ -497,7 +497,7 @@ Other Style Guides
   <a name="destructuring--object"></a><a name="5.1"></a>
   - [5.1](#destructuring--object) Use object destructuring when accessing and using multiple properties of an object. eslint: [`prefer-destructuring`](https://eslint.org/docs/rules/prefer-destructuring)
 
-    > Why? Destructuring saves you from creating temporary references for those properties.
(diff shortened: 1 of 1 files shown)


ljharb (approved):
Thanks, this is great

PR #2316, opened 2020-10-22
README: Fix section 3.2 "good" example

yannoff:
In the 3.2 section of the **Airbnb JavaScript Style Guide**, the "good" and "bad" example don't appear to be equivalent.

### DIFF
--- a/README.md
+++ b/README.md
@@ -188,7 +188,7 @@ Other Style Guides
     const obj = {
       id: 5,
       name: 'San Francisco',
-      [getKey('enabled')]: true,
+      getKey('enabled'): true,
     };
     ```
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -188,7 +188,7 @@ Other Style Guides
     const obj = {
       id: 5,
       name: 'San Francisco',
-      [getKey('enabled')]: true,
+      getKey('enabled'): true,
```
ljharb, line 191: this, in fact, would cause a syntax error - the original version with the brackets is indeed correct and equivalent.

ljharb (changes requested):

PR #2318, opened 2020-10-23
Fix ignoreNonDOM typo in jsx-a11y/aria-role rule

amercier:
This should be `ignoreNonDOM` not `ignoreNonDom` according to [documentation](https://github.com/jsx-eslint/eslint-plugin-jsx-a11y/blob/master/docs/rules/aria-role.md).

### DIFF
--- a/packages/eslint-config-airbnb/rules/react-a11y.js
+++ b/packages/eslint-config-airbnb/rules/react-a11y.js
@@ -17,7 +17,7 @@ module.exports = {
 
     // Require ARIA roles to be valid and non-abstract
     // https://github.com/evcohen/eslint-plugin-jsx-a11y/blob/master/docs/rules/aria-role.md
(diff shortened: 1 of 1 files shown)


packages/eslint-config-airbnb/rules/react-a11y.js
```diff
@@ -17,7 +17,7 @@ module.exports = {
 
     // Require ARIA roles to be valid and non-abstract
     // https://github.com/evcohen/eslint-plugin-jsx-a11y/blob/master/docs/rules/aria-role.md
-    'jsx-a11y/aria-role': ['error', { ignoreNonDom: false }],
+    'jsx-a11y/aria-role': ['error', { ignoreNonDOM: false }],
```
ljharb, line 20: oof, thanks. altho this is the default, so i guess it's fine?

also, i think maybe setting this to `true` would be good, and it'd be semver-patch since it warns less.

ljharb (approved):

PR #2320, opened 2020-10-24
add better messages for extra restricted globals (isFinite, isNaN)

samsch:
Right now the lint error message for globals isFinite and isNaN is just "Unexpected use of '\<global\>'." which can be improved by suggesting the specific replacement.


ljharb (approved):
Thanks!

PR #2321, opened 2020-10-26
Allow missing React when using JSX (React 17)

naoisegolden:
Since React 17, `import React` is [no longer required](https://reactjs.org/blog/2020/09/22/introducing-the-new-jsx-transform.html) for JSX files to correctly transform. 

This is possible through an [update available since `@babel/plugin-transform-react-jsx` 7.9.0](https://github.com/babel/babel/pull/11154). `babel-preset-airbnb` is [already using this version](https://github.com/airbnb/babel-preset-airbnb/blob/master/package.json#L33).

### DIFF
--- a/packages/eslint-config-airbnb/rules/react.js
+++ b/packages/eslint-config-airbnb/rules/react.js
@@ -224,10 +224,6 @@ module.exports = {
       skipUndeclared: false
     }],
 
-    // Prevent missing React when using JSX
-    // https://github.com/yannickcr/eslint-plugin-react/blob/master/docs/rules/react-in-jsx-scope.md
-    'react/react-in-jsx-scope': 'error',
-
     // Require render() methods to return something
     // https://github.com/yannickcr/eslint-plugin-react/blob/master/docs/rules/require-render-return.md
     'react/require-render-return': 'error',


packages/eslint-config-airbnb/rules/react.js
```diff
@@ -224,10 +224,6 @@ module.exports = {
       skipUndeclared: false
     }],
 
-    // Prevent missing React when using JSX
-    // https://github.com/yannickcr/eslint-plugin-react/blob/master/docs/rules/react-in-jsx-scope.md
-    'react/react-in-jsx-scope': 'error',
-
```
ljharb, line 230: this lint rule *must* be enabled for anyone using react < 16.14, and anyone who's not using the new jsx transform.

Is there a way this can be detected, so we can do the right thing?

Note as well, the `react/jsx-uses-react` rule must also be disabled for this scenario.

ljharb (changes requested):

packages/eslint-config-airbnb/rules/react.js
naoisegolden: I see. It's more complex than anticipated in this case. Unfortunately I can't provide a better solution at the moment so I will close the PR. Thanks for the swift review.

PR #2329, opened 2020-11-07
Add `.eslintrc.js` to `import/no-extraneous-dependencies` devDeps

Kocal:
Similar to #1168 and #1522

Since you can import dev dependencies in a `.eslintrc.js` file, I may be interesting to disable devDeps check in this file. 

WDYT? Thanks!

### DIFF
--- a/packages/eslint-config-airbnb-base/rules/imports.js
+++ b/packages/eslint-config-airbnb-base/rules/imports.js
@@ -90,7 +90,8 @@ module.exports = {
         '**/Gruntfile{,.js}', // grunt config
         '**/protractor.conf.js', // protractor config
         '**/protractor.conf.*.js', // protractor config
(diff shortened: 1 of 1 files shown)


ljharb (approved):
Although I'd strongly encourage static configuration in an `.eslintrc` file, this seems reasonable.

Kocal:
It would be great to allow the following files too:
- `postcss.config.js`
- `tailwind.config.js`
- `vite.config.js` (https://github.com/vitejs/vite#config-file)

Would like me to open new PR or I can add those files in this PR?

Thanks

ljharb:
At some point, we can't keep adding config files for everyone's individual tools, especially when they're not tools airbnb uses. PostCSS is the only one of those three that airbnb uses.

Kocal:
I understand, I will configure the rule myself then :+1:

PR #2363, opened 2021-01-05
Update README.md

tamojit-123:
typo mistake


ljharb (changes requested):
This isn’t a mistake; a subtitle is not a sentence, and a period is grammatically incorrect here.

Issue #2301, opened 2020-10-04
JSX is not displayed correctly in webpage


aromalanil:
## Description
- The JSX example is not correctly displayed in the webpage, there is no problem with the Reame file though

## Screenshot
![image](https://user-images.githubusercontent.com/49222186/95022085-959cb880-0692-11eb-956c-e7df32b5cd6c.png)

## Reason
I think the markdown to HTML convertor is what causing the issue

aromalanil:
Here is the Readme Link: https://github.com/airbnb/javascript/blob/master/react/README.md#quotes

khareyash05:
Is this still open? I would like to work on the same...if its open

ljharb:
@khareyash05 yes, it's open, and there's no linked PR, and it has the "pull request wanted" label, so go for it.

JeremyRivera:
So just editing single and double quotes through out?

Blkwidows22:
Are these codes to like put on a chat site or on somebody's phone or I'm new at this I'm sorry I don't know how this works I'm still trying to figure out where to put these codes

JatinSainiOO7:
Hello!
Can I Work On This Issue.
Thankyou.

itsjeremyrivera:
Sure!Jeremy ***@***.*** / 813-352-1618On Nov 3, 2023, at 01:29, JATIN SAINI ***@***.***> wrote:﻿
Hello!
Can I Work On This Issue.
Thankyou.

Robbie-BOIII:
Kamino cloned this issue to [Robbie-BOIII/test](https://github.com/Robbie-BOIII/test/issues/3)

DetrojaRadhey:
I'm interested in working on this issue. Could you please confirm if it's still open?

ljharb:
@DetrojaRadhey you can see at the top of the issue that it's open, but you can also see that #2451 may fix it. Thus, if you have a contribution, please comment with a branch link on #2451, and I can pull it into the existing PR.

DetrojaRadhey:
@ljharb, I have used HTML character entities to display JSX code in a markdown format without being rendered as HTML.

Review this PR and let me know if any changes needed:
https://github.com/airbnb/javascript/pull/3060

ljharb:
@DetrojaRadhey i did that, yesterday.

PR #2365, opened 2021-01-08
ci: create issue labeled reply

xrkffgg:
ref: https://github.com/airbnb/javascript/issues/2301

When you add `pull request wanted`, then GitHub Actions will help you to comment this.


ljharb:
I'm not sure this is particularly helpful; it seems like it'd be more noisy than anything else.

xrkffgg:
OK, if you need, call me~😘

PR #2371, opened 2021-02-03
Update react.js

soupria:
test PR //const dangleRules = baseStyleRules['no-underscore-dangle'];


ljharb (changes requested):
this seems like a spam PR.

PR #2389, opened 2021-02-17
Update ARIA roles link

gabrielslach:
deprecated link


ljharb:
The current link works just fine - what about the current one is "deprecated"?

leon-dunamu (commented):
It works fine, but I think the original one is more correct.
Entries about "WAI-ARIA Roles" are pointed to by the original link.

As a result, I don't think there is a need to change.

ljharb (approved):
since w3.org has unfortunately violated the primary principle of the web, we should update this.

PR #2398, opened 2021-02-25
readme updated with good, bad, well practice in JS

iamraufu:


ljharb (changes requested):
Thanks, but I don't think this adds any clarity.

PR #2407, opened 2021-03-25
Alphabetize the rules for react-a11y.js

backwardok:
The ordering of the rules are currently not alphabetical, which can make it hard to tell whether there are rules missing or what's set for each rule from eslint-plugin-jsx-a11y. 

This change alphabetizes the list so it's easier to compare to the [list of supported rules](https://github.com/jsx-eslint/eslint-plugin-jsx-a11y#supported-rules)

### DIFF
--- a/packages/eslint-config-airbnb-base/package.json
+++ b/packages/eslint-config-airbnb-base/package.json
@@ -62,7 +62,7 @@
     "eslint-plugin-import": "^2.22.1",
     "in-publish": "^2.0.1",
     "safe-publish-latest": "^1.1.4",
-    "tape": "^5.1.0"
+    "tape": "^5.2.2"
   },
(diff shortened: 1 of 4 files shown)


backwardok:
@ljharb @lencioni looks like the builds are all failing due to some dependency vulnerabilities? Other PRs seem to be failing for similar reasons. Any idea what's going on there?

ljharb:
@backwardok yup, i have some things to fix, but since branch protections require review and i can no longer bypass them, i'd have to either make a PR or wait for someone else's PR to hijack :-) i'll add my commit to yours and get the builds passing.

ljharb (approved):

backwardok:
Awesome, thanks @ljharb !

PR #2412, opened 2021-04-22
Using template literals to break lines

henriquebarrosx:
## About
I get frustrated when I need to scroll to see the content, so when I saw this section, I remembered that we can use model literals to be more convenient, readable and we can see all the content at once.

### DIFF
--- a/README.md
+++ b/README.md
@@ -596,6 +596,11 @@ Other Style Guides
 
     // good
     const errorMessage = 'This is a super long error that was thrown because of Batman. When you stop to think about how Batman had anything to do with this, you would get nowhere fast.';
+
+    // best
+    const errorMessage = `This is a super long error that was thrown because 
+      of Batman. When you stop to think about how Batman had anything to do 
+      with this, you would get nowhere fast.`;
     ```
 
   <a name="es6-template-literals"></a><a name="6.4"></a>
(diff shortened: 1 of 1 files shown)


henriquebarrosx:
Also, we can use another approach to functions with multiple arguments

README.md
```diff
@@ -596,6 +596,11 @@ Other Style Guides
 
     // good
     const errorMessage = 'This is a super long error that was thrown because of Batman. When you stop to think about how Batman had anything to do with this, you would get nowhere fast.';
+
+    // best
```
ljharb, line 600: This isn’t actually best as-is, because it includes all the extra spaces you put in for indentation. You’d need to use https://npmjs.com/dedent , or manually do so in the example, neither of which is better than the one-line example.

Long lines are *just fine*, and our eslint config explicitly disables the max-len check on any line with a string.

README.md
```diff
@@ -911,6 +916,13 @@ Other Style Guides
       // ...
     }
 
+    // best
+    function foo(data) {
+      // This prevents the developer from confusing the order of the arguments.
```
ljharb, line 921: but it also defies convention, required arguments must always be positional. Options objects are only for optional arguments.

ljharb (changes requested):

PR #2413, opened 2021-04-23
[editorial] add `var` info to `let`/`const` section

Kelta-King:
People generally don't understand the difference between **var** and **let** variable declarations.
So, an explanation with a proper example might help. 
So, here is the explanation with an example.

### DIFF
--- a/README.md
+++ b/README.md
@@ -139,17 +139,21 @@ Other Style Guides
     ```
 
   <a name="references--block-scope"></a><a name="2.3"></a>
-  - [2.3](#references--block-scope) Note that both `let` and `const` are block-scoped.
+  - [2.3](#references--block-scope) Note that both `let` and `const` are block-scoped, whereas `var` is function-scoped.
 
     ```javascript
     // const and let only exist in the blocks they are defined in.
     {
       let a = 1;
       const b = 1;
+      var c = 1;
     }
     console.log(a); // ReferenceError
     console.log(b); // ReferenceError
+    console.log(c); // Prints 1
     ```
+
+    In the above code, you can see that referencing `a` and `b` will produce a ReferenceError, while `c` contains the number. This is because `a` and `b` are block scoped, while `c` is scoped to the containing function.
 
 **[⬆ back to top](#table-of-contents)**
 


README.md
```diff
@@ -150,6 +150,23 @@ Other Style Guides
...
+    }
+    console.log(i) // Prints 3
+    ```
+    For loop with **let** keyword
+    ```javascript
+    for (let j = 1; j <= 2; j++ ) {
+      console.log(j)
+    }
+    console.log(j) // Error
+    ```
```
ljharb, line 169: while explaining the difference is helpful, this guide forbids both `var` and loops, so i'm not sure this is the best example. Perhaps something that shows the hoisting differences - maybe just adding a `var` to the previous section?

README.md
```diff
@@ -150,6 +150,23 @@ Other Style Guides
     console.log(a); // ReferenceError
     console.log(b); // ReferenceError
     ```
+  <a name="references--difference-let-var"></a><a name="2.4"></a>
```
ljharb, line 153:
```suggestion
  <a name="references--difference-let-var"></a></a>
```
new sections don't get a numbered link; only legacy ones do.

ljharb (changes requested):

README.md
Kelta-King: I will open a new PR for the same. Thank you for your assistance.

ljharb: Please do not open a PR; please repurpose this one instead.

Kelta-King (commented):
Last updated PR has covered the difference of `let` and `var` in the section 2.3

README.md
```diff
@@ -150,6 +150,23 @@ Other Style Guides
...
+  - [2.4](#references--difference-let-var) Difference between `let` and `const`
+    
+    For loop with **var** keyword
```
Danello92: In the JS variable var is define in a scope local. we can recall it in a global scope.

Only in JS we can do it.

Variable let is working only in local scope.

ljharb: I'm not sure what this comment means. `var`, `let`, and `const` are all local scope - var is function scope, and the other two are block scope. None of them are globally scoped.

README.md
```diff
@@ -137,19 +137,22 @@ Other Style Guides
...
     {
       let a = 1;
       const b = 1;
+      var c = 1;
     }
     console.log(a); // ReferenceError
     console.log(b); // ReferenceError
+    console.log(c); // Prints 1
     ```
+    In above code, you can see that the `a` and `b` will generate reference error and `c` prints the number. It is because `a` and `b` are block scoped while `c` is not.
```
ljharb, line 154:
```suggestion
    In above code, you can see that the `a` and `b` will generate a ReferenceError, and `c` prints the number. This is because `a` and `b` are block scoped, while `c` is scoped to the containing function.
```

ljharb (changes requested):

ljharb (approved):

PR #2416, opened 2021-04-29
Fix Object (3.3) with `this` keyword

miko-github:
```javascript
// Objects (3.3) (Line 213)
...

return atom.value + value;
// actually you can access with `this` like `return this.value + value`
// but, this not change `atom.value`, i think this should be:
// `return this.value += value;`
// if we use `atom.value` may be crash, be cause it is globally, but when you use
// `this`, it is objective 🤗 ❤️

...
```

i think this is better, but i didn't any change :heart:

### DIFF
--- a/README.md
+++ b/README.md
@@ -210,7 +210,11 @@ Other Style Guides
       value: 1,
 
       addValue(value) {
-        return atom.value + value;
+        return atom.value + value; // actually you can access with `this` like `return this.value + value`
+                                   // but, this not change `atom.value`, i think this should be:
+                                   // `return this.value += value;`
+                                   // if we use `atom.value` may be crash, be cause it is globally, but when you use
+                                   // `this`, it is objective 🤗 ❤️
       },
     };
     ```


README.md
```diff
@@ -210,7 +210,11 @@ Other Style Guides
       value: 1,
 
       addValue(value) {
-        return atom.value + value;
+        return atom.value + value; // actually you can access with `this` like `return this.value + value`
+                                   // but, this not change `atom.value`, i think this should be:
+                                   // `return this.value += value;`
+                                   // if we use `atom.value` may be crash, be cause it is globally, but when you use
+                                   // `this`, it is objective 🤗 ❤️
```
ljharb, line 217: it's not actually objective, because `this` depends on how the function is invoked. For example:
```js
const { addValue } = atom;
addValue(2); // crashes, because `this` is `undefined` or the global.
```

ljharb (changes requested):

README.md
miko-github: wow, i don't tested this :smile: 
thanks man :+1:

miko-github: and if we use function instead object, may be successful :thinking:

PR #2423, opened 2021-05-11
Spread operator => Spread syntax

hsh2001:
https://stackoverflow.com/questions/44934828/is-it-spread-syntax-or-the-spread-operator


ljharb (approved):
Thanks, this is completely correct.

PR #2429, opened 2021-05-18
Airbnb styling rule recommendations

jsec:
I've notated most rules that we can keep, and I think we can utilize dropping a lot of these rules to 'warn', which will allow the VS Code ESLint plugin to automatically fix them on save instead of developers having to manually fix errors.

There are a couple rules here that I think we should disable as well, and I've left notations on those as well.

There are also 4-5 rules that can be disabled in favor of a Prettier configuration option, where if we set up our tooling to where we lint, then format with Prettier, we will get the same result.

### DIFF
--- a/packages/eslint-config-airbnb-base/rules/style.js
+++ b/packages/eslint-config-airbnb-base/rules/style.js
@@ -3,531 +3,823 @@ module.exports = {
     // enforce line breaks after opening and before closing array brackets
     // https://eslint.org/docs/rules/array-bracket-newline
     // TODO: enable? semver-major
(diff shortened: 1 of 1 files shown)


jsec:
I put this on the wrong base, sorry!!!!!

ljharb:
Note that vscode will autofix things that are set to warn or error; you don't have to set them to "warn" and risk letting them slip through in order to get that behavior.

If you use prettier, extend `eslint-config-prettier` and it will turn off all of the overlapping eslint rules.

jsec:
Good to know, thank you!

PR #2430, opened 2021-05-24
Update README.md

chety:
- Enhance method implementation by returning in a single line rather than using if and else.
- Add support for the default parameter value.

### DIFF
--- a/README.md
+++ b/README.md
@@ -452,9 +452,14 @@ Other Style Guides
       if (subject === 'Mockingbird') {
         return author === 'Harper Lee';
       }
-
       return false;
     });
+    
+    //best
+     inbox.filter(({ subject, author } ) => {
+       return subject === 'Mockingbird' && author === 'Harper Lee';
+    });
+    
     ```
 
   <a name="arrays--bracket-newline"></a>
@@ -522,6 +527,10 @@ Other Style Guides
     function getFullName({ firstName, lastName }) {
       return `${firstName} ${lastName}`;
     }
+    
+    // best - 2
+    const  getFullName = ({ firstName, lastName }) => `${firstName} ${lastName}`;
+    
     ```
 
   <a name="destructuring--array"></a><a name="5.2"></a>
@@ -766,6 +775,19 @@ Other Style Guides
     function handleThings(opts = {}) {
       // ...
     }
+    
+    // best
+    function isNullOrUndefined(obj){
+       // '==' handles both undefined and null
+       return obj == null;
+    }
+    function getDefaultValueIfNotPresent(obj,defaultValue){
+        return isNullOrUndefined(obj) ? defaultValue : obj;
+    }
+
+    function handleThings(opts = getDefaultValueIfNotPresent(opts,{})){
+       // ...
+    }
     ```
 
   <a name="functions--default-side-effects"></a><a name="7.8"></a>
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -452,9 +452,14 @@ Other Style Guides
...
       }
-
       return false;
     });
+    
+    //best
+     inbox.filter(({ subject, author } ) => {
+       return subject === 'Mockingbird' && author === 'Harper Lee';
+    });
+
```
ljharb, line 462: this isn't actually best - our linter config will warn on it. it'd be:
```suggestion
    
    //best
     inbox.filter(({ subject, author }) => subject === 'Mockingbird' && author === 'Harper Lee');
```

however, then we've defeated the whole point of this section, which is showing patterns around explicit `return` in callbacks.

so, i think this is best reverted.

README.md
```diff
@@ -522,6 +527,10 @@ Other Style Guides
     function getFullName({ firstName, lastName }) {
       return `${firstName} ${lastName}`;
     }
+    
+    // best - 2
+    const  getFullName = ({ firstName, lastName }) => `${firstName} ${lastName}`;
+
```
ljharb, line 533: using an arrow function isn't equally as good as using a normal named function; and in this case I'd say it's worse, because you're relying on implicit function name inference.

ljharb (changes requested):

README.md
```diff
@@ -766,6 +775,19 @@ Other Style Guides
     function handleThings(opts = {}) {
       // ...
     }
+    
+    // best
+    function isNullOrUndefined(obj){
+       // '==' handles both undefined and null
+       return obj == null;
+    }
```
ljharb, line 783: this shouldn't ever be a function, it should just be inline.

README.md
```diff
@@ -766,6 +775,19 @@ Other Style Guides
...
+    function getDefaultValueIfNotPresent(obj,defaultValue){
+        return isNullOrUndefined(obj) ? defaultValue : obj;
+    }
```
ljharb, line 786:
```suggestion
    function getDefaultValueIfNotPresent(obj,defaultValue) {
        return isNullOrUndefined(obj) ? defaultValue : obj;
    }
```
this is entirely unnecessary, though, since `??` exists - you can just do `obj ?? defaultValue`, without any need for a function.

README.md
```diff
@@ -951,6 +973,9 @@ Other Style Guides
       const y = x + 1;
       return x * y;
     });
+    
+    // best
+    [1, 2, 3].map((x) => x * (x + 1));
     ```
```
ljharb, line 979: this is fine, but it may not actually be better - the "y" might add clarity, depending on what the code is doing.

README.md
```diff
@@ -766,6 +775,19 @@ Other Style Guides
...
+       // '==' handles both undefined and null
+       return obj == null;
+    }
+    function getDefaultValueIfNotPresent(obj,defaultValue){
+        return isNullOrUndefined(obj) ? defaultValue : obj;
+    }
+
+    function handleThings(opts = getDefaultValueIfNotPresent(opts,{})){
+       // ...
+    }
```
ljharb, line 790: i don't think this adds anything over the existing example of `function handleThings(opts = {}) {`.
```suggestion
```

ljharb (changes requested):

README.md
chety: If you explicitly pass `null` to `handleThings` function then `opts` _will be null rather than the default_ `{}` value. Therefore I think we should also handle this case.

PR #2431, opened 2021-05-26
[readme] Fix inconsistency

gd-js:
Remove punctuation


ljharb:
How is this an inconsistency, exactly? There's only one item in this section.

gd-js:
Hello, in no other instance exists a fullstop in the document like that, specifically after a " - ", which makes it look slightly scrambled together.

PR #2434, opened 2021-06-02
Update imports.js

cojdev:
Added the glob `**/webpack.*.js` to `import/no-extraneous-dependencies` rule to prevent errors when importing dev dependencies to files like `webpack.dev.js`, `webpack.prod.js` and `webpack.common.js`. More info here: https://webpack.js.org/guides/production/


ljharb:
why not name them webpack.config.prod.js?

PR #2466, opened 2021-09-08
fix: add new line after list

akazyti:
This Just adds a new line after list in this markdown. 

newline

### DIFF
--- a/README.md
+++ b/README.md
@@ -73,6 +73,7 @@ Other Style Guides
     - `symbol`
     - `bigint`
 
+    &nbsp;
     ```javascript
     const foo = 1;
     let bar = foo;
@@ -91,6 +92,7 @@ Other Style Guides
     - `array`
     - `function`
 
+    &nbsp;
     ```javascript
     const foo = [1, 2];
     const bar = foo;


README.md
```diff
@@ -73,6 +73,8 @@ Other Style Guides
     - `symbol`
     - `bigint`
 
+    <br />
```
ljharb, line 76: i'd prefer a non-HTML way to add a line break here.

README.md
```diff
@@ -82,7 +84,7 @@ Other Style Guides
     console.log(foo, bar); // => 1, 9
     ```
 
-    - Symbols and BigInts cannot be faithfully polyfilled, so they should not be used when targeting browsers/environments that don’t support them natively.
+  - Symbols and BigInts cannot be faithfully polyfilled, so they should not be used when targeting browsers/environments that don’t support them natively.
```
ljharb, line 87: this change is incorrect, as it causes this note underneath 1.1 to be nested at the same level as 1.1 and 1.2. Please revert it.

ljharb (changes requested):

README.md
akazyti: I Just Changed the Branch and just replaced <br /> with &nbsp at ; #2467

README.md
akazyti: I Just Changed the Branch and just replaced <br /> with &nbsp at ; #2467

README.md
ljharb: Please don't post duplicate PRs.

README.md
ljharb: The change should have been made here; please never ever ever open redundant PRs.

README.md
```diff
@@ -73,6 +73,7 @@ Other Style Guides
     - `symbol`
     - `bigint`
 
+    &nbsp;
```
ljharb, line 76: this is still HTML; is there any way to use whitespace or markdown here?

PR #2474, opened 2021-09-24
Explain why rule `react/jsx-key` is turned off

datenreisender:
~~As you probably are aware, [one should define a `key` property when rendering a list in JSX](https://reactjs.org/docs/lists-and-keys.html#keys). And there is also [the ESLint rule `react/jsx-key`](https://github.com/yannickcr/eslint-plugin-react/blob/master/docs/rules/jsx-key.md) for that, so I am really surprised that this rule is not turned on by default in the AirBnB ESLint settings.~~

As [explained by @ljharb](https://github.com/airbnb/javascript/pull/2474#discussion_r715661547) the rule `react/jsx-key` is turned off on purpose. This adds the explanation for that, so that others also find this easier to understand.

### DIFF
--- a/packages/eslint-config-airbnb/rules/react.js
+++ b/packages/eslint-config-airbnb/rules/react.js
@@ -91,6 +91,7 @@ module.exports = {
 
     // Validate JSX has key prop when in array or iterator
     // https://github.com/yannickcr/eslint-plugin-react/blob/master/docs/rules/jsx-key.md
+    // Turned off because it has too many false positives
     'react/jsx-key': 'off',
 
     // Limit maximum of props on a single line in JSX
--- a/packages/eslint-config-airbnb/rules/react.js
+++ b/packages/eslint-config-airbnb/rules/react.js
@@ -545,6 +545,16 @@ module.exports = {
     // https://github.com/yannickcr/eslint-plugin-react/blob/c2a790a3472eea0f6de984bdc3ee2a62197417fb/docs/rules/no-unstable-nested-components.md
     // TODO: enable, semver-major
     'react/no-unstable-nested-components': 'off',
+
+    // Enforce that namespaces are not used in React elements
+    // https://github.com/yannickcr/eslint-plugin-react/blob/8785c169c25b09b33c95655bf508cf46263bc53f/docs/rules/no-namespace.md
+    // TODO: enable, semver-minor
+    'react/no-namespace': 'off',
+
+    // Prefer exact proptype definitions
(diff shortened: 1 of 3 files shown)


packages/eslint-config-airbnb/rules/react.js
```diff
@@ -91,7 +91,7 @@ module.exports = {
 
     // Validate JSX has key prop when in array or iterator
     // https://github.com/yannickcr/eslint-plugin-react/blob/master/docs/rules/jsx-key.md
-    'react/jsx-key': 'off',
+    'react/jsx-key': ['error', { checkFragmentShorthand: true }],
```
ljharb, line 94: The rule is off on purpose, because it has too many false positives.

ljharb (changes requested):

packages/eslint-config-airbnb/rules/react.js
datenreisender: I am surprised by this, but ok. Then I will just continue to set these in my projects manually.

Commit b294e7cbf, which originally set this, did not elaborate the reasoning. There all missing react rules where just disabled without any explanation why they are not enabled. And searching in the issues and PRs I also found no explanation. Maybe it is helpful, to add this reasoning?

I will change this PR, so that it keeps the rule turned off but adds a reasoning.

ljharb: We don't just enable every rule; some of them are actively harmful. That commit added all the rules that were already "not on", explicitly set to off.

We don't generally explain why a rule isn't turned on - there's too many rules for that to be sustainable.

datenreisender: I disagree whether explaining these rules is sustainable but you are making the calls here, so I will close this PR.

datenreisender:
As explained in the comments above this PR is unwanted.

packages/eslint-config-airbnb/rules/react.js
ljharb: I said "don't generally", i didn't say this PR wouldn't be accepted or not.

ljharb:
No, that's a blatant misunderstanding of the comments.

datenreisender:
Sorry, yes, then I really did not understand you. I really had the impression that you do not want to change anything in this regard. 

So, what kind of change would you like?

ljharb:
I don’t need any change; this is fine, i just hadn’t finished the discussion yet.

datenreisender:
Ok. Then I have the impression there is nothing more I can do about this PR and I will just wait.

ljharb (approved):

PR #2477, opened 2021-10-06
Grammatical Error

usmankh07:
Just corrected a grammar error

Screenshot_1


ljharb:
This isn’t a grammatical error, and your change would be incorrect - “the” implies it’s the only or primary reasonable approach, which it is not.

usmankh07:
I got it!

PR #2482, opened 2021-10-15
clarify hooks requirement in README

MustafaHaddara:
Contrary to what's currently written in the README, the default export does **not** include the lint rules for Hooks. This updates the description to make that clear. 

The other potential change is to update the default export so that it _does_ include the rules for Hooks but I think that would be a breaking change = new major version?


MustafaHaddara:
The section link doesn't work super well in the Rich Diff view, but you can see it in action [on my branch](https://github.com/MustafaHaddara/javascript/tree/update-hooks-readme/packages/eslint-config-airbnb)-- it seems to be working well for me there.

PR #2490, opened 2021-11-04
Add vite.config.js in import/no-extraneous-dependencies

spivet:
Support vite.config.js


ljharb:
See https://github.com/airbnb/javascript/pull/2329#issuecomment-723689369
> At some point, we can't keep adding config files for everyone's individual tools, especially when they're not tools airbnb uses.

spivet:
Ok, I got it. I'd add the configuration in my own `.eslintrc` file. THX!

PR #2493, opened 2021-11-05
docs: add an example using for..of

mrchief:
The document states 

> Don’t use iterators. Prefer JavaScript’s higher-order functions instead of loops like for-in or for-of. 

but doesn't provide code examples. This PR is an attempt to augment the existing code samples with an example.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1507,6 +1507,11 @@ Other Style Guides
     numbers.forEach((num) => {
       increasedByOne.push(num + 1);
     });
+   
+    // good (or better? )
+    for (const num of numbers) {
+      increasedByOne.push(num + 1);
+    }
 
     // best (keeping it functional)
     const increasedByOne = numbers.map((num) => num + 1);


README.md
```diff
@@ -1507,6 +1507,11 @@ Other Style Guides
     numbers.forEach((num) => {
       increasedByOne.push(num + 1);
     });
+   
+    // good (or better? )
+    for (const num of numbers) {
+      increasedByOne.push(num + 1);
+    }
```
ljharb, line 1514: this is worse than forEach, because it's a loop, and it *also* invokes the iterator protocol.

I'd rank this as worse than the standard for loop a few examples up.

ljharb (changes requested):

README.md
mrchief: Help me understand here. First of, loops are not evil - unbounded loops are. This can never be unbounded as it gets an iterable from the object itself which is what the functional equivalents also do. 

Second of all, if this is worse, then why does the first line in the header recommend `for..of`?

ljharb: On the contrary, loops are indeed evil. Here’s a snippet from an irc bot: Loops fail to clearly signal the intention of your code, and as such should be avoided whenever possible (prefer higher-order functions, such as .map, .reduce, .filter, ...). See also: http://qr.ae/RNMEGA and https://gist.github.com/robotlolita/7643014

This guide intends loops to be avoided as close to always as possible.

The wording says “instead of loops (like for-in or for-of)” - for-of is a loop just like for-in, and dispreferred by this guide.

Issue #2478, opened 2021-10-09
eslint v8 support


ljharb:
(see also #2223, #2036, #1834, #1447)

eslint v8 was released on 10/9/21.

~The current plan is, if possible, to release a new semver-minor of both configs, that supports `5 || 6 || 7 || 8`.~ The plan is now to release a semver-major of both configs, that supports eslint 7 and 8 latest (at the time of publishing).

The timeframe for eslint v8 support is "probably not as soon as you would like, but as soon as we can".
Please accept this, and please add no noise to this thread beyond emoji reactions - distractions won't help anyone get the work done any faster.

### **Please** do not file additional issues and do not file pull requests attempting to add support - please, instead, continue to use eslint v7 until such time as our configs support v8.

Peer Dep eslint combined v7 and v8 support released:
 - [x] `eslint-plugin-import`
    - [x] [PR](https://github.com/import-js/eslint-plugin-import/pull/2191)
    - [x] [Release](https://github.com/import-js/eslint-plugin-import/releases/tag/v2.25.0)
 - [x] [`eslint-find-rules`](https://github.com/sarbbottam/eslint-find-rules/issues/336)
   - [x] [PR](https://github.com/sarbbottam/eslint-find-rules/pull/337)
   - [x] [Release](https://github.com/sarbbottam/eslint-find-rules/pull/341)
 - [x] update the base config
   - [x] [PR](https://github.com/airbnb/javascript/pull/2495)
 - [x] publish the base config, semver-major
 - [x] `eslint-plugin-react`
   - [x] `@babel/eslint-parser` supports eslint 8 ([PR](https://github.com/babel/babel/pull/13782))
   - [x] [PR](https://github.com/yannickcr/eslint-plugin-react/pull/3059) (blocked on base config and [eslint-find-rules](https://github.com/sarbbottam/eslint-find-rules/pull/337))
   - [x] Release
 - [x] [`eslint-plugin-jsx-a11y`](https://github.com/evcohen/eslint-plugin-jsx-a11y/issues/683)
   - [x] [PR](https://github.com/jsx-eslint/eslint-plugin-jsx-a11y/pull/810) (blocked on base config)
   - [x] Release
 - [x] [`eslint-plugin-react-hooks`](https://github.com/facebook/react/issues/22533)
   - [x] [PR](https://github.com/facebook/react/pull/22248)
   - [x] [Release](https://github.com/facebook/react/pull/22248#issuecomment-964176674)
[...]

PR #2495, opened 2021-11-09
[eslint config] [base] [breaking] support eslint 8

ljharb:
Technically this wouldn't have to be a semver-major change - we could hold off on enabling the many rules that have been waiting for it, only run `eslint-find-rules` (which was forced to make a breaking change due to API changes in eslint 8) on eslint 7 and 8, and release as a semver-minor.

However, I'm not sure there'd be much value in that - lots of people will update to get eslint 8 support, so it's a nice opportunity to add "exports" and enable more useful rules and get maximum exposure for that. Separately, at this point I'm not sure who's still using eslint 5 or 6, considering how much new syntax requires eslint 7 or 8 to support (even via babel's or typescript's eslint parser) and how hungry people often are for new syntax.

Due to the way we share packages in this quasi-monorepo, I've been forced to disable a number of tests. These will be re-enabled once I open the PR to update the react config (after the base config is published) which hopefully won't be too long (and I won't be making any meaningful changes to the react config during that gap anyways).

This resolves part of #2478.

### DIFF
--- a/.github/workflows/node.yml
+++ b/.github/workflows/node.yml
@@ -13,7 +13,7 @@ jobs:
         with:
           versionsAsRoot: true
           type: 'majors'
-          preset: '^10 || ^12 || ^14 || ^16 || >= 17'
+          preset: '^12 || ^14 || ^16 || >= 17'
 
   base:
     needs: [matrix]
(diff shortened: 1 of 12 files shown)


ljharb:
Going to go ahead and merge and release; any problems can be addressed in patch releases.

PR #2501, opened 2021-11-10
[eslint config] set `namedComponents` option to match style guide

### LINKED ISSUE #2505: `react/function-component-definition` configuration violates style guide
In `eslint-config-airbnb@19.0.0` (specifically https://github.com/airbnb/javascript/commit/1bc8cabd447ef9ce69d27ae70e67eb47daeffc21), `react/function-component-definition` was configured to allow only function expressions for named components:

https://github.com/airbnb/javascript/blob/1bc8cabd447ef9ce69d27ae70e67eb47daeffc21/packages/eslint-config-airbnb/rules/react.js#L526-L531

This has the effect of disallowing function declarations:
```js
function Foo() {
  return <div />
}
```

While allowing only this style:
```js
const Foo = function () {
  return <div />
}
```

This seems like a configuration mistake, as the [React style guide](https://github.com/airbnb/javascript/tree/52e710c14ec0197cfefe131061e1fe41c966f93b/react) consistently uses function declarations for components (as does the ecosystem as a whole) and specifically points out that ["relying on function name inference is discouraged"](https://github.com/airbnb/javascript/tree/52e710c14ec0197cfefe131061e1fe41c966f93b/react#class-vs-reactcreateclass-vs-stateless). Let me know if you agree and I'll PR a fix.

petersendidit:
Change the `namedComponents` option to `function-declaration` to match what the style guide requires.

https://github.com/airbnb/javascript/commit/019e0f7e07477e85249e1d584f22773120c41b91#r60060792

Closes #2505

### DIFF
--- a/packages/eslint-config-airbnb/rules/react.js
+++ b/packages/eslint-config-airbnb/rules/react.js
@@ -524,9 +524,8 @@ module.exports = {
 
     // Enforce a specific function type for function components
     // https://github.com/yannickcr/eslint-plugin-react/blob/master/docs/rules/function-component-definition.md
-    // TODO: investigate if setting namedComponents to expression vs declaration is problematic
     'react/function-component-definition': ['error', {
-      namedComponents: 'function-expression',
+      namedComponents: ['function-declaration', 'function-expression'],
       unnamedComponents: 'function-expression',
     }],
 
--- a/packages/eslint-config-airbnb-base/rules/best-practices.js
+++ b/packages/eslint-config-airbnb-base/rules/best-practices.js
@@ -376,6 +376,11 @@ module.exports = {
     // https://eslint.org/docs/rules/prefer-named-capture-group
     'prefer-named-capture-group': 'off',
 
+    // Prefer Object.hasOwn() over Object.prototype.hasOwnProperty.call()
+    // https://eslint.org/docs/rules/prefer-object-has-own
+    // TODO: semver-major: enable thus rule, once eslint v8.5.0 is required
+    'prefer-object-has-own': 'off',
+
     // https://eslint.org/docs/rules/prefer-regex-literals
     'prefer-regex-literals': ['error', {
       disallowRedundantWrapping: true,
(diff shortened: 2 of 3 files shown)


packages/eslint-config-airbnb/rules/react.js
```diff
@@ -524,9 +524,8 @@ module.exports = {
 
     // Enforce a specific function type for function components
     // https://github.com/yannickcr/eslint-plugin-react/blob/master/docs/rules/function-component-definition.md
-    // TODO: investigate if setting namedComponents to expression vs declaration is problematic
     'react/function-component-definition': ['error', {
-      namedComponents: 'function-expression',
+      namedComponents: 'function-declaration',
```
ljharb, line 528: can you file an upstream issue on eslint-plugin-react to allow this to be an array, so we can do `['function-declaration', 'function-expression']`?

petersendidit: Opened a PR to make that change https://github.com/yannickcr/eslint-plugin-react/pull/3129

Harris-Miller:
The fact that this rule was changed from `'off'` to `'error'` in the `19.0.0` release and not mentioned in the change log is disconcerting

See: https://github.com/airbnb/javascript/blame/master/packages/eslint-config-airbnb/rules/react.js#L528

ljharb:
@Harris-Miller why? I changed a bunch of things from off to error in that commit.

Harris-Miller:
@ljharb yeah, and you generally mention those as breaking changes in the changelog, like for `18.0.0`: https://github.com/airbnb/javascript/blob/master/packages/eslint-config-airbnb/CHANGELOG.md#1800--2019-08-10
But you guys didn't for `19.0.0`: https://github.com/airbnb/javascript/blob/master/packages/eslint-config-airbnb/CHANGELOG.md#1900--2021-11-10

Companies like mine use those changelogs to pre-review those changes when we write our own `eslint` configs that extent yours (which is pretty much everyone nowadays) so we know what we may want to keep or revert. When you have 2.5+ million downloads a week, there is an expectation that breaking rule changes will be shown there

ljharb:
Fair enough. There were a lot, and the changelog is per commit, so i lumped them together. You can certainly review the commit itself as well.

I’ll hide all these comments, since they're off topic for this PR.

ljharb:
(note: this is blocked on v7.28.0 of eslint-plugin-react being released)

ljharb (approved):

PR #2503, opened 2021-11-11
[eslint config] [base] remove unneeded eslint version check

PaperStrike:
As eslint < 7 is dropped in #2495 :)

### DIFF
--- a/packages/eslint-config-airbnb-base/package.json
+++ b/packages/eslint-config-airbnb-base/package.json
@@ -89,7 +89,6 @@
   "dependencies": {
     "confusing-browser-globals": "^1.0.10",
     "object.assign": "^4.1.2",
-    "object.entries": "^1.1.5",
-    "semver": "^6.3.0"
+    "object.entries": "^1.1.5"
   }
 }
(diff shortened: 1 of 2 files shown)


ljharb (approved):
Thanks, the breaking change was unfortunate.

n-studio:
We should release this change as 15.0.1 to solve https://github.com/advisories/GHSA-c2qf-rxjj-qqgw

ljharb:
That isn’t actually a vulnerability here, and we’re on v19 - we won’t be back porting anything to v15.

n-studio:
@ljharb Sorry I was referring to the package https://www.npmjs.com/package/eslint-config-airbnb-base, not https://www.npmjs.com/package/eslint-config-airbnb, it hasn't a v19 release, right?

n-studio:
@ljharb Ok, no problem. I like fixing CVE alerts even if they are false positive so the CI doesn't block my releases. In the meanwhile I'll just load my package from master.

ljharb:
I strongly discourage doing that; there's no guarantee everything will work.

If your CI is blocking releases on false positive CVEs, i'd invite you to consider that it's not actually making your project more secure, but less.

PR #2506, opened 2021-11-15
Add 5.4 Destructuring alias for objects.

Yatin-kathuria:
Add 5.4 Destructuring alias for objects.
Hope this pull request might be helpful.

### DIFF
--- a/README.md
+++ b/README.md
@@ -563,6 +563,19 @@ Other Style Guides
     const { left, top } = processInput(input);
     ```
 
+  <a name="destructuring--alias"></a><a name="5.4"></a>
+    - [5.4](#destructuring--alias) Use object alias for getting store the value in the new variable instaed of assigning it to different variable explicitly.
+
+    ```javascript
+    const person = { name: "foo" };
+
+    // bad
+    const fullName = person.name;
+
+    // good
+    const { name: fullName } = person;
+    ```
+
 **[⬆ back to top](#table-of-contents)**
 
 ## Strings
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -563,6 +563,19 @@ Other Style Guides
     const { left, top } = processInput(input);
     ```
 
+  <a name="destructuring--alias"></a><a name="5.4"></a>
+    - [5.4](#destructuring--alias) Use object alias for getting store the value in the new variable instaed of assigning it to different variable explicitly.
```
ljharb, line 567:
```suggestion
    - [5.4](#destructuring--alias) Use object alias for getting store the value in the new variable instead of assigning it to different variable explicitly.
```

ljharb (changes requested):
This isn’t called an “alias”, it’s destructuring with a rename.

I don’t think this should be a guide section tho; it’s always clearest when there’s no renaming, due to the avoided repetition, but it is not always clearer to use a destructuring rename.

README.md
Yatin-kathuria: Yes, You are write . I explored the MDN docs where I get this [Assigning to new variable names](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Destructuring_assignment#assigning_to_new_variable_names) for this type of operation .

But I think , this is good add on in the docs because in real life problems we have to destructuring the properties from objects but with the different name

ljharb: Yes, but `const x = obj.y` is often cleaner than `const { x: y } = obj`, is my point.

Yatin-kathuria: I Think you are write.

PR #2507, opened 2021-11-15
Update Arrow function 8.1 : added best example in it

Yatin-kathuria:
### Hello Maintainers
added best case in arrow function in 8.1

### DIFF
--- a/README.md
+++ b/README.md
@@ -951,6 +951,9 @@ Other Style Guides
       const y = x + 1;
       return x * y;
     });
+    
+    // best
+    [1, 2, 3].map((x) => x * (x + 1));
     ```
 
   <a name="arrows--implicit-return"></a><a name="8.2"></a>


zahermasloub (approved):

README.md
```diff
@@ -951,6 +951,9 @@ Other Style Guides
       const y = x + 1;
       return x * y;
     });
+    
+    // best
+    [1, 2, 3].map((x) => x * (x + 1));
```
ljharb, line 956: I’m not so convinced this is best; intermediate variables are good and can add clarity. What you’ve lost here is “y”, which might be important in the general case, even if not here.

ljharb (changes requested):

README.md
Yatin-kathuria: My intention is to show that sometimes we can even do this in the single line without using any more variable and with no return statement.

ljharb: Is that particularly helpful in this section?

Yatin-kathuria: From my point of view , It is because that is power of arrow function to write most of the code in just single line without any return statement.

ljharb: Sure, but there's other examples in the guide that show that. It's also a power that's easy to abuse - many things *shouldn't* be done in a single line, and the `return` can often add clarity.

Yatin-kathuria: but this 8.1 is about arrow function and if we are still using the return statement in it , then I don't think their is any difference between the bad or good point except function keyword in above two example as well

ljharb: There is a difference in that it's more concise to convey the same information. You don't need to omit `return` to achieve a benefit.

Yatin-kathuria: My point is that, As this is the point on arrow function in `bad` we show the anonymous function in standard way with function keyword. In `good` one we are showing anonymous function with arrow function including return keyword and then my point  is where we can omit the return as well . My point might not be best but I think if we are discussing about arrow function it must deserve a place in this section.

PR #2508, opened 2021-11-15
Add 7.16 : prefer-rest-params

Yatin-kathuria:
### Hello Maintainer
Add 7.16 : `prefer-rest-params`

### DIFF
--- a/README.md
+++ b/README.md
@@ -929,7 +929,7 @@ Other Style Guides
     ```
 
   <a name="functions--prefer-rest-params"></a><a name="7.16"></a>
-  - [7.14](#functions--prefer-rest-params) Prefer the use `...rest` operator instead of `arguments` keyword for taking n number of arguments in function. eslint: [`prefer-rest-params`](https://eslint.org/docs/rules/prefer-rest-params)
+  - [7.16](#functions--prefer-rest-params) Prefer the use `...rest` operator instead of `arguments` keyword for taking n number of arguments in function. eslint: [`prefer-rest-params`](https://eslint.org/docs/rules/prefer-rest-params)
 
     > Why? It’s cleaner and descriptive, you get the array of arguments which are passed to the function.
 
--- a/README.md
+++ b/README.md
@@ -928,7 +928,7 @@ Other Style Guides
     );
     ```
 
-  <a name="functions--prefer-rest-params"></a><a name="7.16"></a>
+  <a name="functions--prefer-rest-params"></a>
   - [7.16](#functions--prefer-rest-params) Prefer the use `...rest` operator instead of `arguments` keyword for taking n number of arguments in function. eslint: [`prefer-rest-params`](https://eslint.org/docs/rules/prefer-rest-params)
 
     > Why? It’s cleaner and descriptive, you get the array of arguments which are passed to the function.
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -928,6 +928,23 @@ Other Style Guides
     );
     ```
 
+  <a name="functions--prefer-rest-params"></a><a name="7.16"></a>
```
ljharb, line 931: we don't add new sections with numbered links; those are legacy.

ljharb (changes requested):

README.md
Yatin-kathuria: I tried to be consistent with the above points as they had been written. If still their is something wrong then can you help me how can I fixed that.

ljharb: There’s no consistency here - just that anything added in the last couple years will not have a numbered link (`<a name`), and all the older sections will. New ones must not have it.

Yatin-kathuria: their is any refence from where I can take a look. Then I will update this.

ljharb: There are many other sections that are newly added; off the top of my head i can’t recall any specifically.

Yatin-kathuria: done.

README.md
```diff
@@ -928,6 +928,23 @@ Other Style Guides
     );
     ```
 
+  <a name="functions--prefer-rest-params"></a>
+  - [7.16](#functions--prefer-rest-params) Prefer the use `...rest` operator instead of `arguments` keyword for taking n number of arguments in function. eslint: [`prefer-rest-params`](https://eslint.org/docs/rules/prefer-rest-params)
```
ljharb, line 932: rest syntax isn't an operator
```suggestion
  - [7.16](#functions--prefer-rest-params) Prefer the use `...rest` syntax instead of the `arguments` keyword for taking a variable number of arguments in a function. eslint: [`prefer-rest-params`](https://eslint.org/docs/rules/prefer-rest-params)
```

README.md
```diff
@@ -928,6 +928,23 @@ Other Style Guides
...
+
+    > Why? It’s cleaner and descriptive, you get the array of arguments which are passed to the function.
```
ljharb, line 934:
```suggestion
    > Why? It’s cleaner and descriptive, you get the array of arguments which are passed to the function. It's also the only option in an arrow function.
```

README.md
```diff
@@ -928,6 +928,23 @@ Other Style Guides
...
+    ```javascript
+    // bad
+    function foo() {
+      console.log(arguments);
+    }
+
+    // good
+    function foo(...args) {
+      console.log(args);
+    }
```
ljharb, line 945:
```suggestion
    // good
    const foo = (...args) => {
      console.log(args);
    }
    
    // good
    function foo(...args) {
      console.log(args);
    }z
```

ljharb (changes requested):

README.md
Yatin-kathuria: done

README.md
Yatin-kathuria: done

README.md
Yatin-kathuria: done

Yatin-kathuria: done

ljharb (changes requested):
Doesn't this section already exist? https://github.com/airbnb/javascript#es6-rest

PR #2509, opened 2021-11-15
Add 7:16 : rest-operator-last

Yatin-kathuria:
### Hello Maintainers
Add 7:16 : rest-operator-last

### DIFF
--- a/README.md
+++ b/README.md
@@ -310,7 +310,7 @@ Other Style Guides
     ```
 
   <a name="objects--rest-spread"></a>
-  - [3.8](#objects--rest-spread) Prefer the object spread syntax over [`Object.assign`](https://developer.mozilla.org/en/docs/Web/JavaScript/Reference/Global_Objects/Object/assign) to shallow-copy objects. Use the object rest operator to get a new object with certain properties omitted. eslint: [`prefer-object-spread`](https://eslint.org/docs/rules/prefer-object-spread)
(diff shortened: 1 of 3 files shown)


README.md
```diff
@@ -928,6 +928,23 @@ Other Style Guides
     );
     ```
 
+  <a name="functions--rest-operator-last"></a><a name="7.16"></a>
```
ljharb, line 931: we don't add new sections with numbered links; those are legacy.

ljharb (changes requested):

README.md
Yatin-kathuria: I tried to be consistent with the above points as they had been written. If still their is something wrong then can you help me how can I fixed that.

Yatin-kathuria: done

ljharb (changes requested):
`...` is rest syntax; it's not an operator.

However, this is invalid syntax - it's *impossible* to put the rest operator anywhere but last. Why does this need a guide section?

ljharb:
@Yatin-kathuria
> However, this is invalid syntax - it's impossible to put the rest operator anywhere but last. Why does this need a guide section?

Yatin-kathuria:
> @Yatin-kathuria
> 
> > However, this is invalid syntax - it's impossible to put the rest operator anywhere but last. Why does this need a guide section?

ok, it means we are not showing the wrong syntax. I think we can close this PR then.

PR #2512, opened 2021-11-17
Support collocated test and storybook files

sharils:


ljharb:
Why? Colocation isn't a great pattern.

sharils:
Do we both refer to the same definition of collation, https://kentcdodds.com/blog/colocation, ?

Back to the topic, I'd say since https://airbnb.io/javascript/ doesn't discourage it. Then it should be supported.

ljharb:
Yes, that’s the thing I’m saying is a bad idea.

Either way, it’s not scalable to keep expanding this list forever; you can override it yourself in your own config, using overrides, even.

sharils:
> Yes, that’s the thing I’m saying is a bad idea.
> 
> Either way, it’s not scalable to keep expanding this list forever; you can override it yourself in your own config, using overrides, even.

Would you agree it’s not scalable to keep expanding this list forever because we support a lot of framework specific files here and it's scalable if we just include a different paradigm considering how many paradigms are there, collocated and separated, just two.

How about this, I send another PR minimise framework specific codes. E.g. I can see webpack patterns can be combine maybe? That would reduce the lines.

Thank you for advising me to override it myself in my own config. There are two problems.

First, I'm not aware of a way to just insert an item to the array instead of providing the same list again. I'm aiming for not having to maintain this list in our repo but having the community maintain this list.

Secondly, because of the tyranny of default, it eliminates some paradigms for unknown reason that is never addressed in the official document. Perhaps we could document why that's a bad idea so have a common grounds?

My pledge is about being inclusive to different paradigms not just a new framework.

As for the framework problem, I don't see a way to solve this problem, it'll just get worse, and not to matching up to the trend would be even worse. That being said, that isn't a concern of this PR though.

ljharb:
You can use eslint overrides to override this rule for all test and storybook files, and since `overrides[].files` already has the glob, you won't need any globs in the rule's config - so you don't NEED the array.

sharils:
Thanks, I'll investigate more and close for now.

sharils:
For the record, here is how I did it:

```json
{
  "overrides": [
    {
      "files": ["next.config.js"],
      "rules": {
        "import/no-extraneous-dependencies": [
          "error",
          {
            "devDependencies": [
              "next.config.js"
            ]
          }
        ]
      }
    }
  ]
}
```

Wiilliam:
这是来自QQ邮箱的假期自动回复邮件。
 
您好，我现在没空，无法亲自回复您的邮件。我将尽快给您回复。

PR #2513, opened 2021-11-18
[readme] update blocks 16.1

Yatin-kathuria:
I gone through the eslint: `nonblock-statement-body-position` documentation where they didn't mention about the `function`  as block for this rule so, I removed the function example and clearly mention the block statements on which this rule implies.

### DIFF
--- a/README.md
+++ b/README.md
@@ -2144,8 +2144,8 @@ Other Style Guides
 
 ## Blocks
 
-  <a name="blocks--braces"></a><a name="16.1"></a>
-  - [16.1](#blocks--braces) Use braces with all multiline blocks. eslint: [`nonblock-statement-body-position`](https://eslint.org/docs/rules/nonblock-statement-body-position)
+  <a name="blocks--braces"></a>
+  - [16.1](#blocks--braces) Use braces with `if`, `else`, `while`, `do-while`, and `for` statements when writing multiline blocks. eslint: [`nonblock-statement-body-position`](https://eslint.org/docs/rules/nonblock-statement-body-position)
 
     ```javascript
     // bad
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -2144,8 +2144,8 @@ Other Style Guides
 
 ## Blocks
 
-  <a name="blocks--braces"></a><a name="16.1"></a>
```
ljharb, line 2147: existing numbered links must never be removed; URLs must work forever.

README.md
```diff
@@ -2144,8 +2144,8 @@ Other Style Guides
...
-  - [16.1](#blocks--braces) Use braces with all multiline blocks. eslint: [`nonblock-statement-body-position`](https://eslint.org/docs/rules/nonblock-statement-body-position)
+  <a name="blocks--braces"></a>
+  - [16.1](#blocks--braces) Use braces with `if`, `else`, `while`, `do-while`, and `for` statements when writing multiline blocks. eslint: [`nonblock-statement-body-position`](https://eslint.org/docs/rules/nonblock-statement-body-position)
```
ljharb, line 2148: That's because function bodies aren't a block, they're a function body.

The eslint rule doesn't ever have to cover the entirety of the guide section - the guide is the source of truth, and the guide applies to all blocks.

ljharb (changes requested):
It'd be nice if we discussed these kinds of changes in issues before making a PR, since most of them have turned out to be unnecessary.

README.md
Yatin-kathuria: So their is no need to update according to eslint rule.

Yatin-kathuria: ok, got the point.

PR #2535, opened 2022-01-04
Add Class Private Identifier

DawChihLiou:
## Include Class Private Identifier 

ECMA2022 standardized the private identifier `#`. It marks class members as private.

#### Additional Section:

- 9.8 Class members should use `#` to mark them as private members.

#### Modified Section:

- 23.4 Do not use trailing or leading underscores.

#### References:

- [ECMAScript 2022 Language Specification](https://tc39.es/ecma262/#sec-static-semantics-privateboundidentifiers)
- [TC39 Proposal - Private methods and getter/setters for JavaScript classes](https://github.com/tc39/proposal-private-methods)
- [TC39 Proposal  - Class field declarations for JavaScript](https://github.com/tc39/proposal-class-fields)

### DIFF
--- a/README.md
+++ b/README.md
@@ -1291,6 +1291,39 @@ Other Style Guides
       }
     }
     ```
+    
+    <a name="classes--private-identifier"></a>
+  - [9.8](#classes--private-identifier) Class members should use `#` to mark them as private members. 
+
+    ```javascript
+    // bad
+    class Foo {
+      _bar() {
+        console.log('bar');
+      }
+    }
+    
+    // bad
+    class Foo {
+      bar_() {
+        console.log('bar');
+      }
+    }
+    
+    // bad
+    class Foo {
+      __bar__() {
+        console.log('bar');
+      }
+    }
+    
+    // good
+    class Foo {
+      #bar() {
+        console.log(this.bar);
+      }
+    }
+    ```
 
 **[⬆ back to top](#table-of-contents)**
 
@@ -3297,7 +3330,7 @@ Other Style Guides
   <a name="naming--leading-underscore"></a><a name="22.4"></a>
   - [23.4](#naming--leading-underscore) Do not use trailing or leading underscores. eslint: [`no-underscore-dangle`](https://eslint.org/docs/rules/no-underscore-dangle.html)
 
-    > Why? JavaScript does not have the concept of privacy in terms of properties or methods. Although a leading underscore is a common convention to mean “private”, in fact, these properties are fully public, and as such, are part of your public API contract. This convention might lead developers to wrongly think that a change won’t count as breaking, or that tests aren’t needed. tl;dr: if you want something to be “private”, it must not be observably present.
+    > Why? ECMA2022 introduced [`#` as the private identifier](https://tc39.es/ecma262/#sec-static-semantics-privateboundidentifiers). Although a leading underscore is a common convention to mean “private”, in fact, these properties are fully public, and as such, are part of your public API contract. This convention might lead developers to wrongly think that a change won’t count as breaking, or that tests aren’t needed. tl;dr: if you want something to be “private”, it must not be observably present.
 
     ```javascript
     // bad
@@ -3306,7 +3339,7 @@ Other Style Guides
     this._firstName = 'Panda';
 
     // good
-    this.firstName = 'Panda';
+    this.firstName = 'Panda'; 
 
     // good, in environments where WeakMaps are available
     // see https://kangax.github.io/compat-table/es6/#test-WeakMap


README.md
```diff
@@ -3306,7 +3339,7 @@ Other Style Guides
     this._firstName = 'Panda';
 
     // good
-    this.firstName = 'Panda';
+    this.firstName = 'Panda';
```
DawChihLiou, line 3342:
```suggestion
    this.firstName = 'Panda';
```

ljharb:
This isn't something the guide can support until babel-preset-airbnb transforms them, and until eslint fully supports linting them with ecmaVersion 2022, which I don't think it quite does yet.

README.md
```diff
@@ -3297,7 +3330,7 @@ Other Style Guides
...
   - [23.4](#naming--leading-underscore) Do not use trailing or leading underscores. eslint: [`no-underscore-dangle`](https://eslint.org/docs/rules/no-underscore-dangle.html)
 
-    > Why? JavaScript does not have the concept of privacy in terms of properties or methods. Although a leading underscore is a common convention to mean “private”, in fact, these properties are fully public, and as such, are part of your public API contract. This convention might lead developers to wrongly think that a change won’t count as breaking, or that tests aren’t needed. tl;dr: if you want something to be “private”, it must not be observably present.
+    > Why? ECMA2022 introduced [`#` as the private identifier](https://tc39.es/ecma262/#sec-static-semantics-privateboundidentifiers). Although a leading underscore is a common convention to mean “private”, in fact, these properties are fully public, and as such, are part of your public API contract. This convention might lead developers to wrongly think that a change won’t count as breaking, or that tests aren’t needed. tl;dr: if you want something to be “private”, it must not be observably present.
```
ljharb, line 3333: technically, `#` is part of the name - it's not a "private identifier". `#foo` isn't "private foo", it's a distinct name from `foo`.

ljharb (changes requested):

README.md
DawChihLiou: Hey @ljharb, thank you so much for your comment! [`#`](https://tc39.es/ecma262/#prod-PrivateIdentifier) is defined as an official reserved keyword in JavaScript syntax. With `#` prefix, [we can create private members](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Classes/Private_class_fields).

```javascript
class ClassWithPrivateField {
  #privateField;

  constructor() {
    this.#privateField = 42;
    delete this.#privateField;   // Syntax error
    this.#undeclaredField = 444; // Syntax error
  }
}

const instance = new ClassWithPrivateField()
instance.#privateField === 42;   // Syntax error
```

ljharb: No, it very much is not a "reserved keyword". What you're linking to is a grammar production in the specification, which has nothing to do with reserved words or keywords.

The MDN article you linked says what I did:
> The # is a part of the name itself.

I was an editor of the specification when the class fields proposal was merged; I'm quite certain of what I'm talking about.

DawChihLiou:
> This isn't something the guide can support until babel-preset-airbnb transforms them, and until eslint fully supports linting them with ecmaVersion 2022, which I don't think it quite does yet.

I did some research and found out that [ESLint seems to lint private class member](https://eslint.org/docs/rules/no-unused-private-class-members#rule-details) since `v8.1.0`.

README.md
DawChihLiou: Understood:) Thanks a lot @ljharb! I learned a lot from your comments.

DawChihLiou:
```suggestion
    > Why? ECMA2022 introduced [**# names**](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Classes/Private_class_fields) to create private class members with a `#` prefix. Although a leading underscore is a common convention to mean “private”, in fact, these properties are fully public, and as such, are part of your public API contract. This convention might lead developers to wrongly think that a change won’t count as breaking, or that tests aren’t needed. tl;dr: if you want something to be “private”, it must not be observably present.
```

ljharb:
That’s great - but our eslint config still supports eslint 7, so we wouldn’t be able to recommend using private fields (which are great) until we drop eslint < 8.1.

DawChihLiou:
Understood. Thanks for taking the time:)

Jassi10000 (changes requested):
@DawChihLiou 
Buddy your PR shows " this branch is out-of-date with the base branch "

I think you need to first do git checkout your-branch. get the latest changes from the upstream to your-branch by pulling from master and then push the changes you've got from upstream

ljharb:
@Jassi10000 while that's technically true, it's not that important until the PR is ready to be merged - and it's something maintainers can often take care of.

PR #2557, opened 2022-03-01
Update ecmaVersion

yanhaijing:


ljharb (changes requested):
This is quite intentional, and it matches babel-preset-airbnb.

Please don't submit unsolicited PRs to projects when it's not clear it's wanted.

PR #2558, opened 2022-03-01
[readme] use Portuguese flag

doriandevtech:
Brazilian flag replaced by the Portuguese flag for the Portuguese language


ljharb:
Is the linked translation in Brazilian Portuguese (pt-BR) or Portuguese (pt)? If the former, then the Brazilian flag would be correct.

doriandevtech:
> Is the linked translation in Brazilian Portuguese (pt-BR) or Portuguese (pt)? If the former, then the Brazilian flag would be correct.

I just looked at the file linked to the Brazilian flag and I didn't find any mention of "pt-BR" or "pt" but it seems that the author is from Brazil so... Maybe the Brazilian flag is more accurate 🤷🏻‍♂️

ljharb:
Google suggests:
>  To say “you”, Brazilians use the word você. In Portugal, this is done with the word tu.

It seems like the linked repo is indeed the pt-BR dialect, given its many usages of the term. Can you confirm?

doriandevtech:
> Google suggests:
> 
> > To say “you”, Brazilians use the word você. In Portugal, this is done with the word tu.
> 
> It seems like the linked repo is indeed the pt-BR dialect, given its many usages of the term. Can you confirm?

Exactly you're right ! 😊

Issue #2457, opened 2021-08-20
Why is eslint-plugin-import configured to ignore node_modules?


MustafaHaddara:
In [eslint-config-airbnb-base/rules/imports.js](https://github.com/airbnb/javascript/blob/master/packages/eslint-config-airbnb-base/rules/imports.js#L26), `import/ignore` is set to ignore `node_modules` and a number of non-js files (coffee, css, etc.) 

Ignoring `node_modules` means that none of the lint analysis can happen against external libraries. In particular, I care about `import/named` flagging things like

```
import { badBadBad } from 'redux';
```

What's the rationale for this change? We can already use eslint's native ignore functionality to avoid linting all of node_modules, right?

ljharb:
I added that in https://github.com/airbnb/javascript/commit/b45fdcd60f6e08f13718bc2348e8d6ea1314e91c, because it's what airbnb was using internally at the time.

https://github.com/import-js/eslint-plugin-import#importignore suggests that in fact we should remove node_modules from this setting - I'd want to treat that as semver-major, however, since it could result in a lot more warnings.

MustafaHaddara:
I'm happy to write that PR. 

How do you manage releases on this library? Is there a schedule/readmap for upcoming semver-major releases? Is this change big enough to merit its own major release?

ljharb:
There’s no schedule, and major releases are avoided for as long as possible. It’s definitely not big enough to warrant one solely for that, since it’s a change you can make in your own config.

We can queue up the PR, though, and then whenever a major release does happen, it will be pulled in.

MustafaHaddara:
Fair enough, that makes sense. 

I'll put up a PR and link it to this issue. 

Thanks!

PR #2567, opened 2022-03-07
[eslint config] [imports] fix imports.js remove node_modules from imp…

hossein13m:
remove `node_module` from `packages/eslint-config-airbnb-base/rules/imports.js`

this pull request resolves the following issue: #2457

### DIFF
--- a/packages/eslint-config-airbnb-base/rules/imports.js
+++ b/packages/eslint-config-airbnb-base/rules/imports.js
@@ -24,7 +24,6 @@ module.exports = {
     'import/core-modules': [
     ],
     'import/ignore': [
-      'node_modules',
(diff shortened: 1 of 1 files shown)


ljharb:
This is a duplicate of #2464. Please do not close it; both PRs must now stay open until they can both be landed at once.

hossein13m:
any update on this? @ljharb

hossein13m:
Any update on this? @ljharb 
How about I merge #2464 into my branch and then add a co-authored commit so we can have both of these PRs?

ljharb:
@hossein13m since it's a semver-major change, and it's unlikely we'll have a semver-major change for a year or two, both PRs will need to stay open until that time.

PR #2589, opened 2022-05-03
Fix broken ESLint eol-last link; remove .html ending in ESLint urls

terrablue:
* Fixed broken link to the ESLint `eol-last` rule
* Aligned ESLint URLs ending with `.html` with the linking style in rest of the document


ljharb (approved):
Thanks!

PR #2591, opened 2022-05-14
Allow devDependencies import for vite/vitest

gavvvr:


ljharb (changes requested):
I don’t think it’s scalable to keep adding specific exceptions here.

Unless Airbnb uses vite (or any tool), i don’t think it should be added here. You can add this in your own config.

gavvvr:
@ljharb makes sense. But at the same time I do not see why karma should be included(#2121) to base airbnb config and vitest not.


Actually, I would like to extend, but for now I do not know how.
How can I __add__ files to the list for `devDependencies` without completely overwriting the rule's configuration?
I've posted a detailed question here: https://stackoverflow.com/q/72244396/3581539

ljharb:
Airbnb uses karma.

You’d need to import the value in a JS config, and then add to the array.

gavvvr:
> You’d need to import the value in a JS config, and then add to the array.

A code snippet for this is required

ljharb:
That sounds pretty imperious; you can write your own code snippet.

I'm going to close this as a wontfix. Happy to reopen if anyone from Airbnb comments that Vite is used internally.

gavvvr:
Hi @ljharb
Sorry for being so concise in my previous comment. Posted it from my mobile device on the go.

I meant, that I get the generic idea. I need to:
1) somehow import the rule computed by inheriting from airbnb config
2) and then only adjust the `devDependencies` array

The problem is that I have no idea how to do the 1st step. That's what my stackoverflow question was about. Could you please point me to any example which shows how to get the current rule's value inherited from configs you `extends`? or provide an approximate code snippet to illustrate the idea?

Regarding closing the PR, I have no objections. I just thought that if Karma is there, then other test framework can be easily added, because it makes sense for everyone, who will start using vitest. But if this base configuration supports only tools which Airbnb uses in it's project (like Karma), then I got the idea and there is no problem. Now I only need to know how to "import the value in a JS config, and then add to the array"

Issue #2428, opened 2021-05-18
no-restricted-syntax for...of warning potentially confusing


gpoole:
The warning about using for...of states that regenerator-runtime is required:

> iterators/generators require regenerator-runtime, which is too heavyweight for this guide to allow them. Separately, loops should be avoided in favor of array iterations.

The explanation creates the impression regenerator-runtime is essential for any code using for...of, which is no longer true, as [for...of seems to be fairly widely supported](https://caniuse.com/?search=for-of).

I think it could make sense to reword the explanation to state that regenerator _may_ be required and that imperative loops should be avoided.

ljharb:
It’s still true if you support older browsers, which most companies do indeed have to do.

It’s fine to improve the message tho.

ANaphade:
I would like to rephrase the message

PR #2595, opened 2022-05-22
Updated style.js for rephrasing message for #2428

ayushi-naphade:


packages/eslint-config-airbnb-base/rules/style.js
```diff
@@ -338,7 +338,7 @@ module.exports = {
       },
       {
         selector: 'ForOfStatement',
-        message: 'iterators/generators require regenerator-runtime, which is too heavyweight for this guide to allow them. Separately, loops should be avoided in favor of array iterations.',
+        message: 'iterators/generators may require regenerator-runtime, which is too heavyweight for this guide to allow them. Separately, imperative loops should be avoided in favor of array iterations.',
```
ljharb, line 341: what loops are non-imperative?

ANaphade: Array.map, Array.reduce, Array.filter etc are non-imperative approaches

ljharb: Those aren't loops, those are iterations.

ANaphade: Yes

ljharb: how about this:
```suggestion
        message: 'iterators/generators require regenerator-runtime in older browsers/engines, which is too heavyweight for this guide to allow them. Separately, loops should be avoided in favor of array iterations.',
```

PR #2600, opened 2022-06-03
[eslint config] Prefer default arguments to default props

runmoore:
defaultProps on function components is getting deprecated by React (see: [here](https://github.com/reactjs/rfcs/blob/createlement-rfc/text/0000-create-element-changes.md#deprecate-defaultprops-on-function-components))

The official recommendation in future versions of React is to use ES6 default values for function components.

### Before

```js
Header.defaultProps = {
  currency: 'GBP',
};

export default function Header({ currency }) {
```

### After

```js
export default function Header({ currency = 'GBP' }) {
```

### DIFF
--- a/packages/eslint-config-airbnb/rules/react.js
+++ b/packages/eslint-config-airbnb/rules/react.js
@@ -388,9 +388,11 @@ module.exports = {
     'react/no-array-index-key': 'error',
 
     // Enforce a defaultProps definition for every prop that is not a required prop
-    // https://github.com/jsx-eslint/eslint-plugin-react/blob/843d71a432baf0f01f598d7cf1eea75ad6896e4b/docs/rules/require-default-props.md
+    // Enforce defaultArguments rather than defaultProps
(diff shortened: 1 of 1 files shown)


runmoore:
Thanks for looking @ljharb.  Just so I understand when this might be accepted. It's been deprecated in React for functional components in so much that React will throw a warning: 

https://github.com/facebook/react/blob/aec575914a0fd76c9db8998ea62c3ee975de70d7/packages/react-server/src/ReactFizzServer.js#L880-L887

Could you point me to the recommendation that defaultProps should be used in functional components because this warning to me suggests default arguments is the recommended way.

ljharb:
Thanks, that indeed is an actual deprecation warning; looks like it was added in react 18.

That said, defaultProps remain superior, and I hope React won't follow through with this ill-advised deprecation.

runmoore:
So is that still a no for this PR?

Could you share any resources that explain why `defaultProps` is superior to default arguments for functional components?

ljharb:
It’s visible externally, and for non-primitive values, it only evaluates once (instead of every invocation), which avoids a rerendering footgun.

PR #2613, opened 2022-07-08
Better support for `test`, `tests`, and `spec` directories for `import/no-extraneous-dependencies` rule

moberegger:
Updates the globs for `test`, `tests`, and `spec` directories so that they will be picked up from anywhere in the project. This gets it to work identically to `__tests__`.

Right now the current rule will only pick up files in these directories if the directory is in the root of the project. If a project co-locates `test` directories with the source (ex: under `/src/foo/test/foo.test.js`), the rule does not take affect and you'll get linting errors when `require`ing or `import`ing dev dependencies.


ljharb:
I'm not comfortable making it easier to colocate tests.

PR #2615, opened 2022-07-14
js-fix

j0pgrm:
fixes issue #2301 with jsx error.


ljharb (changes requested):
This isn't a fix, it would make the syntax invalid.

It's currently passing an object literal.

PR #2623, opened 2022-08-08
[eslint config] [base] disable the deprecated `no-spaced-func` rule

sodatea:
This rule was deprecated since ESLint v3.3.0 and replaced by the
`func-call-spacing` rule.
https://eslint.org/docs/latest/rules/no-spaced-func

`func-call-spacing` was enabled in this package long ago:
https://github.com/airbnb/javascript/commit/27dcb99c6d80718946d10017144b2aa68bed8b9e

So there's no need for it to stay in the ruleset.

Besides, it is causing some tricky issues when overriding the rules for
TypeScript support (https://github.com/iamturns/eslint-config-airbnb-typescript/issues/246)

### DIFF
--- a/packages/eslint-config-airbnb-base/rules/style.js
+++ b/packages/eslint-config-airbnb-base/rules/style.js
@@ -351,7 +351,8 @@ module.exports = {
     ],
 
     // disallow space between function identifier and application
-    'no-spaced-func': 'error',
+    // deprecated in favor of func-call-spacing
+    'no-spaced-func': 'off',
 
     // disallow tab characters entirely
     'no-tabs': 'error',


packages/eslint-config-airbnb-base/whitespaceRules.js
```diff
@@ -23,7 +23,6 @@ module.exports = [
   'no-mixed-spaces-and-tabs',
   'no-multi-spaces',
   'no-regex-spaces',
-  'no-spaced-func',
```
ljharb, line 26: this is still a whitespace rule, so it still belongs here

packages/eslint-config-airbnb/whitespaceRules.js
```diff
@@ -23,7 +23,6 @@ module.exports = [
   'no-mixed-spaces-and-tabs',
   'no-multi-spaces',
   'no-regex-spaces',
-  'no-spaced-func',
```
ljharb, line 26: also here

ljharb (changes requested):

packages/eslint-config-airbnb-base/rules/style.js
```diff
@@ -350,9 +350,6 @@ module.exports = {
       },
     ],
 
-    // disallow space between function identifier and application
-    'no-spaced-func': 'error',
-
```
ljharb, line 355: actually on second thought, let's leave this here, set to "off", with a comment explaining that it's deprecated in favor of the newer rule.

sodatea: Yeah, that makes sense

ljharb (approved):
Thanks!

PR #2641, opened 2022-09-06
Create Jenkinsfile

ahmadhassanSystemsltd:
may be it will use forintegration of eslint with jenkins through jenkinsfile


ljharb (changes requested):
airbnb doesn't use jenkins

PR #2643, opened 2022-09-14
[chore] Remove country flags for translation list

lizheming:
It doesn't suitable that add a country flags near by language name. Because a language may be used by many of countries.

### DIFF
--- a/README.md
+++ b/README.md
@@ -3939,22 +3939,22 @@ Other Style Guides
 
   This style guide is also available in other languages:
 
-  - ![br](https://raw.githubusercontent.com/gosquared/flags/master/flags/flags/shiny/24/Brazil.png) **Brazilian Portuguese**: [armoucar/javascript-style-guide](https://github.com/armoucar/javascript-style-guide)
-  - ![bg](https://raw.githubusercontent.com/gosquared/flags/master/flags/flags/shiny/24/Bulgaria.png) **Bulgarian**: [borislavvv/javascript](https://github.com/borislavvv/javascript)
-  - ![ca](https://raw.githubusercontent.com/fpmweb/javascript-style-guide/master/img/catala.png) **Catalan**: [fpmweb/javascript-style-guide](https://github.com/fpmweb/javascript-style-guide)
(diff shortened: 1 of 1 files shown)


ljharb:
That's a reasonable take, but it's visually nice to have icons to identify them, and flags are the closest thing to a language icon we have.

Do you have a better suggestion?

lizheming:
@ljharb I think there has no icon can represent a language name. But let we think about it another way, why we need a nice icon for a translation list? If we have no applicable to express them, why not remove them? At last, I think the right infomation is more important than nice style for a standard repo.

ps. Let we look at wikipedia, it also hasn't language icon at language list.

image

ljharb:
I'm not sure I find that convincing.

What's the downside to having the flags here?

lizheming:
@ljharb For example, both [South Korea](https://en.wikipedia.org/wiki/South_Korea#Language) and [North Korea](https://en.wikipedia.org/wiki/North_Korea#Language) are use Korea language, but their country flags are difference. Same situation like [Germany](https://en.wikipedia.org/wiki/Germany#Languages) and [Austria](https://en.wikipedia.org/wiki/Austria#Language) both use german language, [Argentina](https://en.wikipedia.org/wiki/Argentina#Languages) and [Mexico](https://en.wikipedia.org/wiki/Mexico#Languages) and [Spain](https://en.wikipedia.org/wiki/Spain#Languages) both use Spanish language. It's not suitable to use one country flag represent language name.

ljharb:
Sounds like an argument to include multiple country flags rather than just removing them?

lizheming:
@ljharb every national language may have different count countries, I think both country flags add to list may misaligned. So i think remove them is better than add all of country flags.

ljharb:
True. The list could instead be turned into a table, though, which would address that problem without having to remove any flags.

lizheming:
@ljharb I can't agree with table solution. The main purpose for translation list just a link that we can go to translation repo, It's not emphasis for our repo. Is it too heavy for us that a table which has both countries flags?

ljharb:
That purpose is served by providing a visual cue so readers can quickly find the language they're looking for in the list.

lizheming:
@ljharb A language list is enough to visualize to user find target. I think more and more flag icons would make user dazzling. If you still decide to use table, I still respect it~

ljharb:
I think the only change that makes sense here is adding any missing flags, and converting to a table so alignment is preserved.

PR #2653, opened 2022-10-03
[guide] add missing full stop in rule 10.10

BoringBoredom:


ljharb:
I don't think this is particularly necessary.

PR #2662, opened 2022-10-25
Add Optional Chaining

mrna0:
I think this code could do better with Optional Chaining.
But in this example it could not exceed 100 characters.
Can you check eligibility?

More;
https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Optional_chaining

### DIFF
--- a/README.md
+++ b/README.md
@@ -2852,6 +2852,14 @@ Other Style Guides
       && jsonData.foo.bar.baz.quux
       && jsonData.foo.bar.baz.quux.xyzzy;
 
+    // better
+    const foo = jsonData
+      ?.foo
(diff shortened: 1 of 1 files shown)


ljharb (changes requested):
Until babel-preset-airbnb supports optional chaining (which it already might), this guide shouldn't. Can you confirm?

Otherwise, looks good.

mrna0:
> Until babel-preset-airbnb supports optional chaining (which it already might), this guide shouldn't. Can you confirm?
> 
> Otherwise, looks good.

Yes I confirm.
Thank you.

PR #2676, opened 2022-11-06
Update .editorconfig

Coderayush13:
remove blanks lines!


ljharb (changes requested):
These are there intentionally for readability.

PR #2686, opened 2022-12-03
[Tests] refactor test paths

mertcanaltin:
- Edited fs.readdirSync call
- Operations done within the loop of the files object are set as function
- This function is called in a loop by returning the keys of the files object using the Object.keys function 🚀

### DIFF
--- a/packages/eslint-config-airbnb/test/test-base.js
+++ b/packages/eslint-config-airbnb/test/test-base.js
@@ -6,13 +6,14 @@ const base = require('../base');
 
 const files = { base };
 
-fs.readdirSync(path.join(__dirname, '../rules')).forEach((name) => {
+const rulesDir = path.join(__dirname, '../rules');
+fs.readdirSync(rulesDir).forEach((name) => {
   if (name === 'react.js' || name === 'react-a11y.js') {
     return;
   }
 
   // eslint-disable-next-line import/no-dynamic-require
-  files[name] = require(`../rules/${name}`); // eslint-disable-line global-require
+  files[name] = require(path.join(rulesDir, name)); // eslint-disable-line global-require
 });
 
 Object.keys(files).forEach((name) => {
--- a/packages/eslint-config-airbnb-base/test/test-base.js
+++ b/packages/eslint-config-airbnb-base/test/test-base.js
@@ -6,9 +6,10 @@ import index from '..';
 
 const files = { ...{ index } }; // object spread is to test parsing
 
-fs.readdirSync(path.join(__dirname, '../rules')).forEach((name) => {
+const rulesDir = path.join(__dirname, '../rules');
+fs.readdirSync(rulesDir).forEach((name) => {
   // eslint-disable-next-line import/no-dynamic-require
-  files[name] = require(`../rules/${name}`); // eslint-disable-line global-require
+  files[name] = require(path.join(rulesDir, name)); // eslint-disable-line global-require
 });
 
 Object.keys(files).forEach((


ljharb:
why is this an improvement?

mertcanaltin:
> bu neden bir gelişme?

clean use :/

ljharb:
What specifically about your approach is cleaner?

mertcanaltin:
> Yaklaşımınız hakkında özellikle ne daha temiz?

Defining base and files objects can be done in one line as follows
`const files = {
  base: require('../base'),
};
 `

When calling fs.readdirSync, it can be changed as follows to merge the folder path with path.join and verify that the directory is rules: ı think

`const rulesDir = path.join(__dirname, '../rules');
fs.readdirSync(rulesDir).forEach((name) => {
  if (name === 'react.js' || name === 'react-a11y.js') {
    return;
  }

  // eslint-disable-next-line import/no-dynamic-require
  files[name] = require(path.join(rulesDir, name));
});`

Operations done inside the loop of the files object can be passed to a function like this

Operations done within the loop of the files object can be transferred to the function

I thought it would be healthier for it to be more organized and understandable, my friend.

packages/eslint-config-airbnb/test/test-base.js
```diff
@@ -2,22 +2,21 @@ import fs from 'fs';
...
-fs.readdirSync(path.join(__dirname, '../rules')).forEach((name) => {
+const rulesDir = path.join(__dirname, '../rules');
+fs.readdirSync(rulesDir).forEach((name) => {
   if (name === 'react.js' || name === 'react-a11y.js') {
     return;
   }
 
   // eslint-disable-next-line import/no-dynamic-require
-  files[name] = require(`../rules/${name}`); // eslint-disable-line global-require
+  files[name] = require(path.join(rulesDir, name));
```
ljharb, line 16: these changes seem nice (altho the eslint override comment is still needed)

ljharb (changes requested):
There's nothing inherently cleaner about defining something in one line versus two.

I commented on the changes that seem good; the rest don't add anything imo.

packages/eslint-config-airbnb/test/test-base.js
mertcanaltin: I edited @ljharb  thank you 🚀

packages/eslint-config-airbnb/test/test-base.js
```diff
@@ -6,18 +6,17 @@ const base = require('../base');
...
```
ljharb, line 16:
```suggestion
  files[name] = require(path.join(rulesDir, name)); // eslint-disable-line global-require
```

packages/eslint-config-airbnb/test/test-base.js
```diff
@@ -31,4 +30,9 @@ Object.keys(files).forEach((name) => {
       .filter((ruleId) => ruleId.indexOf('react/') === 0);
     t.deepEquals(reactRuleIds, [], 'there are no react/ rules');
   });
-});
+};
+
+Object.keys(files).forEach((name) => {
+  const config = files[name];
+  checkRules(name, config);
+});
```
ljharb, line 38: let's revert this

packages/eslint-config-airbnb/test/test-base.js
```diff
@@ -6,18 +6,17 @@ const base = require('../base');
...
 
   // eslint-disable-next-line import/no-dynamic-require
-  files[name] = require(`../rules/${name}`); // eslint-disable-line global-require
+  files[name] = require(path.join(rulesDir, name));
 });
 
-Object.keys(files).forEach((name) => {
-  const config = files[name];
-
+const checkRules = (name, config) => {
```
ljharb, line 19: and this.

ljharb (changes requested):

packages/eslint-config-airbnb/test/test-base.js
mertcanaltin: i got it back dude @ljharb

ljharb (approved):
Reverted the rest of the changes, re-added the eslint override, and made the same changes in the other test-base file.

PR #2688, opened 2022-12-05
[guide] Add `Object.hasOwn`

mrna0:
Hello,

18.2 Use // for single line comments. Place single line comments on a newline above the subject of the comment. Put an empty line before the comment unless it’s on the first line of a block.
18.1 Use /** ... */ for multiline comments.

For hasOwn;
https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Object/hasOwn

### DIFF
--- a/README.md
+++ b/README.md
@@ -311,6 +311,8 @@ Other Style Guides
     /* or */
     import has from 'has'; // https://www.npmjs.com/package/has
     console.log(has(object, key));
+    /* or */
+    console.log(Object.hasOwn(object, key)); // https://www.npmjs.com/package/object.hasown
     ```
 
   <a name="objects--rest-spread"></a>


README.md
```diff
@@ -308,9 +308,11 @@ Other Style Guides
     // best
     const has = Object.prototype.hasOwnProperty; // cache the lookup once, in module scope.
     console.log(has.call(object, key));
-    /* or */
+    // or
```
ljharb, line 311: this style is used intentionally to attract attention; let's revert please.

README.md
```diff
@@ -2763,7 +2765,7 @@ Other Style Guides
 
 
       getAge(today, birthday) {
-        // ..
+        // ...
```
ljharb, line 2768: i don't think it's worth the diff churn to add a third period :-) if you wanted to change them to a proper ellipsis, i guess that'd be fine, otherwise revert.

ljharb (changes requested):

README.md
mrna0: this document is so beautiful that I want it to look perfect :)

README.md
mrna0: [10.4](https://github.com/airbnb/javascript#modules--no-duplicate-imports)

Do you mind if I change it as shown here then?
Because then we will not have made the mistake mentioned in the document. (18.2)

README.md
mrna0: then I decided to fix this inside the same commit while fixing something else.

mrna0: I changed this under commit named Add console.log.

ljharb: Sorry if my original comment wasn't clear - these should be reverted to two dots.

mrna0: Ok, I've recommitted.

ljharb (approved):

adamwillisMastery (commented):
Markdown changes been reviewed.

joneardley1:
Hi my name is Jon Eardley and I am new to GitHub and I don't really
understand a whole lot about it I would love to be able to talk to somebody
if that's possible

PR #2702, opened 2022-12-29
Rename README.md to .readme.md

Juliusproduction:


ljharb (changes requested):
no? this would be worse and would break the convention that every repo on github follows.

PR #2748, opened 2023-03-23
added rule import/no-extraneous-dependencies to docs

### LINKED ISSUE #2727: Documentation for 'import/no-extraneous-dependencies' is missing
The rule is set to error in the config:
https://github.com/airbnb/javascript/blob/master/packages/eslint-config-airbnb-base/rules/imports.js#L71

But there's no mention of it in the guide:
https://github.com/airbnb/javascript/blob/master/README.md

This rule should be mentioned under the Modules section.

meddy672:
Added the eslint rule **import/no-extraneous-dependencies** to the modules section. Fixes #2727

### DIFF
--- a/README.md
+++ b/README.md
@@ -1469,6 +1469,23 @@ Other Style Guides
     import bar from './bar';
     import baz from './baz';
     ```
+  <a name="modules--import/no-extraneous-dependencies"></a><a name="10.3"></a>
+  - [10.11](#modules--import/no-extraneous-dependencies) Forbid the use of extraneous packages.
+  eslint: [`import/no-extraneous-dependencies`](https://github.com/benmosher/eslint-plugin-import/blob/master/docs/rules/no-extraneous-dependencies.md)
+
+    > Why? You should only allow modules that are declared in the package.json's dependencies, devDependencies, optionalDependencies, peerDependencies, or bundledDependencies to be imported and used within the project.
+
+    ```javascript
+    // bad
+    // Not declared in package,json - es6.js
+    import es6 from 'es6';
+    const es6 = require('es6');
+
+    // good
+    // Declared in package.json es6.js
+    import es6 from 'es6';
+    const es6 = require('es6');
+    ```
 
 **[⬆ back to top](#table-of-contents)**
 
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -1468,6 +1468,27 @@ Other Style Guides
     import foo from './foo';
     import bar from './bar';
     import baz from './baz';
+
+    ```
+
+  <a name="modules--import/no-extraneous-dependencies"></a><a name="10.11"></a>
```
ljharb, line 1474: new sections don't get numbered links.
```suggestion
  <a name="modules--import/no-extraneous-dependencies"></a>
```

README.md
```diff
@@ -1468,6 +1468,27 @@ Other Style Guides
...
+
+  - [10.11](#modules--import/no-extraneous-dependencies) Forbid the use of extraneous packages.
+  eslint: [`import/no-extraneous-dependencies`](https://github.com/benmosher/eslint-plugin-import/blob/master/docs/rules/no-extraneous-dependencies.md)
```
ljharb, line 1477:
```suggestion
  eslint: [`import/no-extraneous-dependencies`](https://github.com/import-js/eslint-plugin-import/blob/master/docs/rules/no-extraneous-dependencies.md)
```

README.md
```diff
@@ -1468,6 +1468,27 @@ Other Style Guides
...
+    ```javascript
+    // bad
+    // Not declared in package.json - es6.js
+    import es6 from 'es6';
+    const es6 = require('es6');
+
+    // good
+    // Declared in package.json es6.js
+    import es6 from 'es6';
+    const es6 = require('es6');
```
ljharb, line 1490: let's use a package name someone's actually likely to use here.

but also, i'm not sure how helpful this example is since it's not showing package.json. but maybe that'd be overkill?

ljharb (changes requested):

README.md
cseas: I added this lint rule in our project when someone tried to use "axios" without mentioning it in package.json. That can be used as an example.

README.md
mo74m3ed: > new sections don't get numbered links.
> 
> ```suggestion
> 
>   <a name="modules--import/no-extraneous-dependencies"></a>
> 
> ```

mo74m3ed (approved):

PR #2764, opened 2023-04-22
fix(README): replace wrong word (Enforce) with the right one (Avoid)

vaishnavyogesh:
Replace Enforce with Avoid to make the correct statement.

### DIFF
--- a/README.md
+++ b/README.md
@@ -2903,7 +2903,7 @@ Other Style Guides
     ```
 
   <a name="whitespace--computed-property-spacing"></a>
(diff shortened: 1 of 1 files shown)


ljharb:
Enforce isn’t the wrong word here.

vaishnavyogesh:
image

The code above says **Enforce spacing inside of computed property brackets** but there is no spacing inside the brackets in the code section commented as **good**

vaishnavyogesh:
@ljharb ^

ljharb:
“Enforce” can mean either “require” or “forbid”.

vaishnavyogesh:
Okay! thank you for clarification

PR #2774, opened 2023-05-20
fix: Updated grammatical errors on LICENSE.md

Zadid13:
The changes made primarily involve the placement of commas to enhance readability and clarify the intended meaning of the sentence.

### DIFF
--- a/LICENSE.md
+++ b/LICENSE.md
@@ -4,7 +4,7 @@ Copyright (c) 2012 Airbnb
 
 Permission is hereby granted, free of charge, to any person obtaining a copy
 of this software and associated documentation files (the "Software"), to deal
-in the Software without restriction, including without limitation the rights
+in the Software without restriction, including, without limitation, the rights
 to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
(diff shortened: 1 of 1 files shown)


LICENSE.md
```diff
@@ -4,7 +4,7 @@ Copyright (c) 2012 Airbnb
 
 Permission is hereby granted, free of charge, to any person obtaining a copy
 of this software and associated documentation files (the "Software"), to deal
-in the Software without restriction, including without limitation the rights
+in the Software without restriction, including, without limitation, the rights
```
ljharb, line 7: these commas are absent in the canonically recognized version of the MIT license, so i think it's best not to add them.

ljharb (changes requested):

LICENSE.md
amantank: I think a comma alone would not likely invalidate the MIT License or change its fundamental nature.

amantank (approved):

LICENSE.md
ljharb: I agree, but it might make static license detection difficult.

PR #2775, opened 2023-05-21
fix: Updated README.md and alligned removing extra spaces

Zadid13:
Aligned Removing extra spaces in the comment-out


ljharb (changes requested):
This would be incorrect, as the comment is supposed to be aligned with displayName (the object properties), not the closing curly braces.

PR #2810, opened 2023-08-12
[readme] correct no-use-before-define justification

hamzahuda:
The recommendation is to declare the variables, classes and functions before. So the contrasting negative example should say after.


ljharb (approved):
thanks, this isn't correcting grammar, but correcting the meaning of the sentence

PR #2815, opened 2023-08-21
[eslint config] [base] Add a message for `confusing-browser-globals`

vmohir:
I suggest adding a message for confusing-browser-globals. Currently, it only says you shouldn't use this global variable which is confusing.

### DIFF
--- a/packages/eslint-config-airbnb-base/rules/variables.js
+++ b/packages/eslint-config-airbnb-base/rules/variables.js
@@ -28,7 +28,10 @@ module.exports = {
         message:
           'Use Number.isNaN instead https://github.com/airbnb/javascript#standard-library--isnan',
       },
-    ].concat(confusingBrowserGlobals),
+    ].concat(confusingBrowserGlobals.map((g) => ({
+      name: g,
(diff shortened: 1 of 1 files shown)


packages/eslint-config-airbnb-base/rules/variables.js
```diff
@@ -28,7 +28,15 @@ module.exports = {
...
           'Use Number.isNaN instead https://github.com/airbnb/javascript#standard-library--isnan',
       },
-    ].concat(confusingBrowserGlobals),
+    ].concat(confusingBrowserGlobals.map(function (g) {
+      return {
+        name: g,
+        message:
+          "Use window." +
+          g +
+          " instead. https://github.com/facebook/create-react-app/blob/main/packages/confusing-browser-globals/README.md",
```
ljharb, line 37: this smells of prettier :-)
```suggestion
        message: "Use window." + g + " instead. https://github.com/facebook/create-react-app/blob/HEAD/packages/confusing-browser-globals/README.md",
```

Good change tho!

ljharb (changes requested):

ljharb (approved):

PR #2818, opened 2023-08-24
Imports rules: add eslint flat config support

OlivierZal:


ljharb:
This change is fine, but it doesn't actually add flat config support, it just adds a default for a flat config *config file*.

OlivierZal:
@ljharb, basically flat config *can* already be supported, the only blockers are:
- deprecated key `extends`
- deprecated key `env`
- former key `parserOptions` replaced by new key `languageOptions`
- key `plugins` not a list anymore but an object

so that we need some architectural decision in order to make something like this:

```javascript
const airbnbConfig = require('eslint-config-airbnb-base/config') // for example
```

import this:

```javascript
const airbnbBestPractices = require('eslint-config-airbnb-base/rules/best-practices')
const airbnbErrors = require('eslint-config-airbnb-base/rules/errors')
const airbnbES6 = require('eslint-config-airbnb-base/rules/es6')
const airbnbImports = require('eslint-config-airbnb-base/rules/imports')
const airbnbNode = require('eslint-config-airbnb-base/rules/node')
const airbnbStrict = require('eslint-config-airbnb-base/rules/strict')
const airbnbStyle = require('eslint-config-airbnb-base/rules/style')
const airbnbVariables = require('eslint-config-airbnb-base/rules/variables')
const globals = require('globals')

airbnbES6.languageOptions = {
  globals: {
    ...globals.es2015,
  },
  parserOptions: airbnbES6.parserOptions,
}
delete airbnbES6.env
delete airbnbES6.parserOptions

airbnbImports.languageOptions = {
  globals: {
    ...globals.es2015,
  },
  parserOptions: airbnbImports.parserOptions,
}
airbnbImports.plugins = airbnbImports.plugins.reduce((acc, plugin) => {
  acc[plugin] = plugin
  return acc
}, {})
airbnbImports.plugins = newObj;
delete airbnbImports.env
delete airbnbImports.parserOptions
delete airbnbImports.plugins

delete airbnbNode.env

module.exports = [
  languageOptions: {
    globals: {
      ...globals.node,
    },
  ],
  airbnbBestPractices,
  airbnbErrors,
  airbnbNode,
  airbnbStyle,
  airbnbVariables,
  airbnbES6,
  airbnbImports,
  airbnbStrict,
]
```

ljharb:
Is the only way to support that a new entrypoint? if so, then a `/flat` entrypoint in both configs seems fine with me.

OlivierZal:
@ljharb, I modified the PR and successfully tested it.

OlivierZal:
Hold on the review, tomorrow I will propose an improvement 😊

OlivierZal:
@ljharb, I reshaped.

OlivierZal:
@ljharb, I modified the PR and successfully tested it.

I also updated the npm documentation: https://github.com/airbnb/javascript/blob/72fd2bcd8ba863e1aa7f4d6bd294ed1a52ff6fcc/packages/eslint-config-airbnb-base/README.md

OlivierZal:
@ljharb, I can get rid of `.reduce()` if you think it's not readable.

packages/eslint-config-airbnb-base/flat.js
```diff
@@ -0,0 +1,64 @@
...
+  es6: 'es2015',
+};
+const handledPlugins = {
+  import: importPlugin,
+};
+
+function convertIntoEslintFlatConfig(config) {
+  const newConfig = { ...config, languageOptions: {} };
+
+  // Handle the `extends` key
```
ljharb, line 26:
```suggestion
  // Handle the `env` key
```

packages/eslint-config-airbnb-base/flat.js
```diff
@@ -0,0 +1,64 @@
...
+
+const {
+  extends: airbnbRules,
+  ...airbnbConfig
+} = require('eslint-config-airbnb-base');
+const importPlugin = require('eslint-plugin-import');
+const globals = require('globals');
+
+const mappedEnvs = {
+  es6: 'es2015',
```
ljharb, line 17: we should probably have an exhaustive mapping here just in case. does eslint expose it at all?

packages/eslint-config-airbnb-base/flat.js
```diff
@@ -0,0 +1,64 @@
...
+};
+const handledPlugins = {
```
ljharb, line 19: shouldn’t we land a flat entrypoint in the import plugin instead of handling it here?

packages/eslint-config-airbnb-base/flat.js
```diff
@@ -0,0 +1,64 @@
...
+};
+
+function convertIntoEslintFlatConfig(config) {
+  const newConfig = { ...config, languageOptions: {} };
+
+  // Handle the `extends` key
+  if ('env' in newConfig) {
+    newConfig.languageOptions.globals = Object.keys(newConfig.env)
+      .filter((key) => newConfig.env[key] === true)
+      .reduce((acc, key) => {
```
ljharb, line 30: reduce is fine, but this might be cleaner with flatMap and Object.fromEntries

packages/eslint-config-airbnb-base/flat.js
OlivierZal: The mapping is exhaustive: all the accepted eslint `env` values are the same except `es6`.
`es2016`, `es2018`, `es2019`, `es2022`, `es2023`, `es2024` are not in `globals`, but are not used by `eslint-config-airbnb` as for now (and `globals` has `es5` which is not accepted by `eslint`).

I will add a comment to make it clear.

OlivierZal: Indeed I can improve based on https://github.com/eslint/eslint/blob/main/conf/globals.js

packages/eslint-config-airbnb-base/flat.js
OlivierZal: Another argument against implementing it in the plugin side is that in fact the plugin key can be customized, provided that it prefixes the plugin rules. 

So the logic is more for an implementation on the config side more than on the plugin one.

OlivierZal: Another argument against implementing it on the plugin side is that in fact the plugin key can be customized, provided that it prefixes the plugin rules. 

ESlint flat config allows to do the following:
```
plugin: { importPlugin },
rules: { ˋ${importPlugin}/no-unresolved`: 'off' }
```

So the logic is more for an implementation on the config side more than on the plugin one.

ljharb: what i mean is, eslint-config-airbnb-base should be able to import a single object that can be spread into a flat config, that Just Works.

OlivierZal:
@ljharb, updated accordingly (with a fake `import` plugin entry point, for the moment).

packages/eslint-config-airbnb-base/flat.js
```diff
@@ -0,0 +1,96 @@
...
+    delete newConfig.parserOptions;
+  }
+
+  // Remove the `plugins` key as it will be spread directly during export
+  if ('plugins' in newConfig) {
+    delete newConfig.plugins;
+  }
+
+  return newConfig;
+}
```
ljharb, line 90: this should do it without any mutation:
```suggestion
function convertIntoEslintFlatConfig({
  env, // convert to explicit globals list
  parserOptions, // move into languageOptions
  plugins, // Remove the `plugins` key as it will be spread directly during export
  ...oldConfig
}) {
  return {
    ...oldConfig,
    languageOptions: {
      ...('env' in config && {
        globals: Object.fromEntries(
          Object.keys(env)
            .filter((key) => env[key] === true && key in envMapping && envMapping[key] in globals)
            .flatMap((key) => Object.entries(globals[envMapping[key]]))
        ),
        ...('parserOptions' in config && {
          languageOptions: {
            parserOptions,
          },
        }),
      })
    }
  };
}
```

ljharb:
This looks good - obv it's blocked on the import plugin supporting flat config.

OlivierZal:
Working on it :-)

OlivierZal:
Fixes & successfully tested.

OlivierZal:
Given the fact `eslint-doc-generator` is probably not ready (see the related `eslint-import-plugin` PR) and the flat config logic according to which the rule prefix can be redefined / customized on the config side, can we reconsider spreading the plugin directly into the config?

ljharb:
Why did you delete your fork, making this PR unrecoverable??

packages/eslint-config-airbnb-base/flat.js
OlivierZal: New PR: https://github.com/airbnb/javascript/pull/2826

PR #2819, opened 2023-08-24
[readme] Added Proper meaningful word in section 2.3

chetannada:
Added Proper meaningful word in section [2.3](https://github.com/airbnb/javascript#references--block-scope)

Here a, b and c contains the number 1 so I corrected the sentence `while c prints the number 1.` from `while c contains the number.`

### DIFF
--- a/README.md
+++ b/README.md
@@ -157,7 +157,7 @@ Other Style Guides
     console.log(c); // Prints 1
     ```
 
-    In the above code, you can see that referencing `a` and `b` will produce a ReferenceError, while `c` contains the number. This is because `a` and `b` are block scoped, while `c` is scoped to the containing function.
+    In the above code, you can see that referencing `a` and `b` will produce a ReferenceError, while `c` prints the number 1. This is because `a` and `b` are block-scoped, while `c` is scoped to the containing function.
 
 **[⬆ back to top](#table-of-contents)**
 


README.md
```diff
@@ -157,7 +157,7 @@ Other Style Guides
     console.log(c); // Prints 1
     ```
 
-    In the above code, you can see that referencing `a` and `b` will produce a ReferenceError, while `c` contains the number. This is because `a` and `b` are block scoped, while `c` is scoped to the containing function.
+    In the above code, you can see that referencing `a` and `b` will produce a ReferenceError, while `c` prints the number 1. This is because `a` and `b` are block-scoped, while `c` is scoped to the containing function.
```
ljharb, line 160: "contains" is correct; "prints" is simply incorrect, and the "1" is fine, but is not needed.

While "block-scoped" is also correct, the hyphen is not required - MDN uses a hyphen consistently, but almost everything else on the web omits it.

chetannada: Hi @ljharb 
Well, If we write `c` contains the number then what about `a` and `b` ? It also contain number right! but here both produce a ReferenceError so I added that `c` `prints the number 1.` means `it doesn't produce a ReferenceError` and `it prints the number` not due to `c` contains the number` because `a` and `b` also `contains number.`.

Correct me If i'm wrong. Thanks.

ljharb: a and b contain nothing, because they're in TDZ, which is why they produce a reference error.

chetannada: okay, so can I add TDZ in above sentence for solving confusion?
For ex: 
>In the above code, you can see that referencing `a` and `b` will produce a ReferenceError because they're in Temporal Dead Zone (TDZ), while `c` contains the number.  This is because `a` and `b` are block scoped, while `c` is scoped to the containing function.

What's your view? Thanks.

ljharb: I don't think it's valuable to talk about the TDZ here.

PR #2823, opened 2023-08-31
remove: unused a tag

0o001:


ljharb (changes requested):
This is actually used and very important, since URLs should last forever, and at one time the misspelling was the main link.

0o001:
> This is actually used and very important, since URLs should last forever, and at one time the misspelling was the main link.

Ah, alright then, I understand. It was good that I learned, thank you :)

PR #2824, opened 2023-09-02
Support flat config

OlivierZal:


ljharb:
I'm not likely to drop support for the non-flat config for the better part of a decade, to be clear.

packages/eslint-config-airbnb-base/flat/rules/errors.js
```diff
@@ -0,0 +1,189 @@
+module.exports = {
```
ljharb, line 1: it’s definitely untenable to duplicate the config and make more than one source of truth.

OlivierZal: It was still an incomplete draft but I give up on the direction I tried.

ljharb: That's fine, but please reuse PRs instead of closing them and opening new ones.

PR #2826, opened 2023-09-03
Support eslint flat config

OlivierZal:


OlivierZal:
Blocked by https://github.com/import-js/eslint-plugin-import/pull/2873

Pilaton:
@ljharb, It would be worth adopting this PR, if only as a temporary solution to support a flat configuration. 🤷‍♂️

ljharb:
@Pilaton a temporary solution would be worse than no solution at all, so it remains blocked.

rarenatoe:
@OlivierZal @Pilaton @ljharb 2 days ago flat config was allowed for eslint-plugin-import

https://github.com/import-js/eslint-plugin-import/releases/tag/v2.30.0

ljharb:
There's no need to close and delete the PR - it's just not quite time to land it yet.

PR #2840, opened 2023-09-28
`no-return-await` Rule Deprecated

kawikabader:
The `no-return-await` rule has been deprecated in ESLint v8.46.0.

https://eslint.org/docs/latest/rules/no-return-await


ljharb:
Well, that was a terrible decision since v8 isn’t the only engine. This config will be keeping this rule forever, and if it’s removed in an future eslint version, we’ll just stay on old eslint forever.

PR #2842, opened 2023-09-29
Docsify page

diomed:
I thought it would be nice to have HTML version of repository so I docsified it. 

docsify uses README file to create web page content which can be displayed in github pages.
so you can enter link in description of repository and content will then be displayed as standalone web page like it is in my repo:
[preview here](https://diomed.github.io/javascript)

Cheers!

### DIFF
--- a/index.html
+++ b/index.html
@@ -59,7 +59,7 @@
     <div id="app"></div>
     <script>
       window.$docsify = {
-        name: 'AirBNB Javascript guide ',
+        name: 'AirBNB Javascript Style Guide ',
         //coverpage: true,//
         repo: '',
         basePath:
--- a/index.html
+++ b/index.html
@@ -1,0 +1,82 @@
+<!DOCTYPE html>
+<html lang="en">
+  <head>
+    <meta charset="UTF-8" />
+    <title>JavaScript Style Guide</title>
+    <meta http-equiv="X-UA-Compatible" content="IE=edge,chrome=1" />
+    <meta name="description" content="Description" />
+    <meta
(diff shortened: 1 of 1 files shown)


ljharb (changes requested):
Thees already one; it’s linked from the repo.

Either way we’d never want the html committed, because then there’s 2 sources of truth.

diomed:
what do you mean by 'there's 2 sources of truth?'

ljharb:
When the guide is updated, it has to be changed in two places.

Generated output doesn’t ever belong committed to source control..

diomed:
> When the guide is updated, it has to be changed in two places.
> 
> Generated output doesn’t ever belong committed to source control..

no. you don't seem to understand. it doesnt have to be changed in two places. 
this is HTML version of content of your README file.  when you update readme, docsify just generates webpage with content of your readme. that's the beauty of it.

diomed:
I also don't see this link anywhere

> Thees already one; it’s linked from the repo.

ljharb:
If it’s generated, it doesn’t belong in the repo.

https://airbnb.io/javascript/ (you’re right tho, we don’t link to it because the markdown is a better experience)

diomed:
allright, thanks for pointing out existence of it to me. I was unaware. 
that link should be in repo description, tho. so it's anyone's choice how they're gonna see it. 

✌️ 

thank you for creating this very useful repository. cheers!

ljharb:
It’s not there because we prefer people to use the markdown.

PR #2848, opened 2023-10-07
fixing README title

RandomNose:
Leaving a more beautiful head title to the README of this repo.


ljharb (changes requested):
That’s not more beautiful, it just erases the joke. There’s a matching } at the end.

PR #2849, opened 2023-10-12
Allow importing dev dependencies in ESLint flat config file

filiptammergard:
Makes it match the behavior for the legacy `.eslintrc.js` config file.

### DIFF
--- a/packages/eslint-config-airbnb-base/rules/imports.js
+++ b/packages/eslint-config-airbnb-base/rules/imports.js
@@ -92,7 +92,7 @@ module.exports = {
         '**/protractor.conf.*.js', // protractor config
         '**/karma.conf.js', // karma config
         '**/.eslintrc.js', // eslint config
-        '**/eslint.config.{js,mjs,cjs}' // eslint flat config
+        '**/eslint.config.js' // eslint flat config
       ],
       optionalDependencies: false,
     }],
--- a/packages/eslint-config-airbnb-base/rules/imports.js
+++ b/packages/eslint-config-airbnb-base/rules/imports.js
@@ -91,7 +91,7 @@ module.exports = {
         '**/protractor.conf.js', // protractor config
         '**/protractor.conf.*.js', // protractor config
         '**/karma.conf.js', // karma config
-        '**/.eslintrc.js' // eslint config
+        '**/.eslintrc.js', // eslint config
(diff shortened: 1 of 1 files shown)


packages/eslint-config-airbnb-base/rules/imports.js
```diff
@@ -91,7 +91,8 @@ module.exports = {
         '**/protractor.conf.js', // protractor config
         '**/protractor.conf.*.js', // protractor config
         '**/karma.conf.js', // karma config
-        '**/.eslintrc.js' // eslint config
+        '**/.eslintrc.js', // eslint config
+        '**/eslint.config.{js,mjs,cjs}' // eslint flat config
```
ljharb, line 95: I won’t believe eslint allows any extension but .js here.

ljharb (changes requested):

packages/eslint-config-airbnb-base/rules/imports.js
filiptammergard: @ljharb True. They mention the other extensions might be needed in different circumstances, but only the .js extension is automatically picked up. 

Updated now.

packages/eslint-config-airbnb-base/rules/imports.js
```diff
@@ -91,7 +91,8 @@ module.exports = {
         '**/protractor.conf.js', // protractor config
         '**/protractor.conf.*.js', // protractor config
         '**/karma.conf.js', // karma config
-        '**/.eslintrc.js' // eslint config
+        '**/.eslintrc.js', // eslint config
```
filiptammergard, line 94: Do we want to specify that this format is called eslintrc?

```suggestion
        '**/.eslintrc.js', // eslint eslintrc config
```

filiptammergard:
Do you have any further input on this one @ljharb?

packages/eslint-config-airbnb-base/rules/imports.js
ljharb: it's called "flat config", they're all eslintrc.

Tseafidelis: 2210030129

ljharb:
@filiptammergard i know it'd been awhile, but by deleting the fork, this PR is unrecoverable.

PR #2864, opened 2023-11-09
Provide example of mid-variable acronym naming

danbeam:
Technically 23.9 says "Acronyms and initialisms should always be all uppercased, or all lowercased", however some counterexamples came up during a recent code review.

Counterexample: a variable that captures a GHE (GitHub Enterprise) API (Application Programming Interface) URL (Uniform Resource Locator) in the current rules can be (if all acronyms were to be used):

* `GHEAPIURL` or
* `gheapiurl`

but it seems reasonable to allow:

* `gheApiUrl`

based on the spirit of "readability" (making the word boundaries clearer).

So I'm just adding an example in hopes of not really needing to change the rule but still making sense.

### DIFF
--- a/README.md
+++ b/README.md
@@ -3521,7 +3521,7 @@ Other Style Guides
     ];
 
     // also good
-    const httpRequests = [
+    const httpUrlRequests = [
       // ...
     ];
 


README.md
```diff
@@ -3521,7 +3521,7 @@ Other Style Guides
     ];
 
     // also good
-    const httpRequests = [
+    const httpUrlRequests = [
```
ljharb, line 3524: first, let's keep `httpRequests`; but second, `Url` is literally never correct, ever, and should never appear in your code, so this would be a very bad example to provide.

ljharb (changes requested):

ljharb:
iow, `httpURLRequests` is infinitely better than `httpUrlRequests`. We're not java, let's aim higher :-)

README.md
danbeam: `Url` is correct in this case in Google's TypeScript, JavaScript, and C++ style guides

https://google.github.io/styleguide/tsguide.html#camel-case
https://google.github.io/styleguide/jsguide.html#naming-camel-case-defined
https://google.github.io/styleguide/cppguide.html#General_Naming_Rules

danbeam:
> iow, `httpURLRequests` is infinitely better than `httpUrlRequests`. We're not java, let's aim higher :-)

I'm not sure what this means; is something being divided by zero / why infinitely?  what does it mean to aim higher?

danbeam:
@ljharb you [requested changes](https://github.com/airbnb/javascript/pull/2864#pullrequestreview-1725264256); what would you like me to change?

README.md
ljharb: Unfortunately Google's style guides haven't ever been considered idiomatic by the wider ecosystem.

ljharb:
"aim higher" means, let's do better than Java. `Url` is flatly incorrect, and in JS, we should avoid it.

I don't have a concrete change to suggest; if you can think of an example of mid-variable acronym naming that doesn't improperly case an initialism then I'd be happy to consider it. If not, then it may be best to close that.

PR #2876, opened 2023-12-28
Update README.md

Omchaudhary1220:
Added the need of comments in the JavaScript code.

### DIFF
--- a/README.md
+++ b/README.md
@@ -2427,7 +2427,7 @@ Other Style Guides
     // bad
     // make() returns a new element
     // based on the passed in tag name
-    //
+    // Used to provide some information about the code in short term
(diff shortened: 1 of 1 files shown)


ljharb (changes requested):
I’m not really sure what value this adds, can you explain?

Omchaudhary1220:
Actually I wanted to start open source contribution so I added very basic
line which is line specifying why we need comments.

ljharb:
ok, but not only isn’t that the reason we make comments, it’s not what this section is for.

While the contribution is appreciated, generally speaking you should contribute to fix a problem you have, not just for the sake of contributing.

PR #2880, opened 2024-01-10
fixing README title

LcasFerrei:
Apenas para teste estudantil.


ljharb:
This is spam, not for learning, and the readme title is not broken. Please do not do this.

If it comes from a course or book or something, can you please link me to what's recommending this bad advice?

Issue #2804, opened 2023-08-06
Support of new eslint flat config


OlivierZal:
Hi,

`eslint.config.js` should be added in https://github.com/airbnb/javascript/blob/cda44dad403979d313a4b02ce55f0bb6cc3a24b2/packages/eslint-config-airbnb-base/rules/imports.js#L94 in order to support new eslint flat config. 

Thanks!

ljharb:
sadly yes, that's true.

OlivierZal:
I can do the PR if you want, is there a release branch?

ljharb:
Actually we may need to wait until all of our peer deps support flat config. If that’s already working, then a PR to the default branch is fine.

Krushna-Nanda:
how should i contribute , where to start

OlivierZal:
@ljharb, I don't know the exact process on the project, but here's a PR from a fork: https://github.com/airbnb/javascript/pull/2818/files

Aishanya12:
its already resolved, you can close this

filiptammergard:
A PR for adding `eslint.config.js` is up: https://github.com/airbnb/javascript/pull/2849

You can already use flat config, even when using plugins and configs that don't yet support it, by using `...compat`: https://eslint.org/docs/latest/use/configure/migration-guide#using-eslintrc-configs-in-flat-config Many project already to this.

I don't think we need to wait for full flat config support to allow importing dev dependencies from flat config files.

sergeushenecz:
After upgrade to eslint 9.0.0. now i see error 
I used ` ...compat.extends('airbnb'),` and all was ok. 
`Error: Failed to load config "airbnb" to extend from.` 
Maybe someone know how does it fix?

OlivierZal:
I gave up with airbnb config, now I configure the rules I want directly in my flat config (without the airbnb package).

sergeushenecz:
is this library not suported ?

OlivierZal:
Every config needs to be updated to support new v9 eslint API, which has breaking changes. 

Since v9 has been released yesterday, we probably need to wait for configs’ compatibility (see eslint/typescript, which has a dedicated issue with thread about it and tracking).

ljharb:
Correct; just like with every eslint major, all the plugins this config depends on have to support it before we can.

adrian-gierakowski:
For reference, tracking issues for plugins which need to be resolved first:

https://github.com/import-js/eslint-plugin-import/issues/2556

[...]

PR #2904, opened 2024-02-18
Issue #2804 Added support for .eslintrc.js in imports.js

SahilThakur02:
Issue #2804 
Implemented support for `.eslintrc.js` configuration files in the `imports.js` resolver. This change enables ESLint to recognize and use `.eslintrc.js` files within the project, enhancing configurability and compatibility with the new flat ESLint configuration format.


ljharb (changes requested):
That’s not how this setting works; it won’t have any effect.

PR #2928, opened 2024-03-05
[spam]

ganeshwhere:
Import useState from the react package. Use the useState hook to manage the modal's visibility state. Replace the handleClose and handleShow methods with arrow functions that call setShow with the appropriate value.


ljharb (changes requested):
this makes no sense

PR #2929, opened 2024-03-05
[spam]

ganeshwhere:
Enforces the use of single quotes for strings. Requires semicolons at the end of statements. No-unused-vars: Warns when a variable is declared but not used.


ljharb (changes requested):
the syntax isn't even valid

PR #2938, opened 2024-03-07
Update README file - Extend explanations, fix formatting mistakes and expand examples

bstashchuk:
I am author of few JavaScript courses on Udemy.
While recording new JavaScript course I used AirBnb Style Guide in one of the sections and noticed few ways for improvements.

- Expanded examples
- Fixed some formatting mistakes
- Extended explanations


README.md
```diff
@@ -46,7 +46,7 @@ Other Style Guides
   1. [Accessors](#accessors)
   1. [Events](#events)
   1. [jQuery](#jquery)
-  1. [ECMAScript 5 Compatibility](#ecmascript-5-compatibility)
```
bstashchuk, line 49: Previous link to ES5 compatibility table is broken now.
I decided to add lint to ES6 Can I use feature search.

README.md
```diff
@@ -412,7 +411,7 @@ Other Style Guides
     ```
 
   <a name="arrays--mapping"></a>
-  - [4.6](#arrays--mapping) Use [`Array.from`](https://developer.mozilla.org/en/docs/Web/JavaScript/Reference/Global_Objects/Array/from) instead of spread `...` for mapping over iterables, because it avoids creating an intermediate array.
+  - [4.6](#arrays--mapping) Use [`Array.from`](https://developer.mozilla.org/en/docs/Web/JavaScript/Reference/Global_Objects/Array/from) instead of spread `...` for mapping over iterables like Sets, Maps or NodeLists, because it avoids creating an intermediate array.
```
bstashchuk, line 414: It wasn't clear that this approach should not be used for regular Arrays

README.md
```diff
@@ -740,13 +738,13 @@ Other Style Guides
 
     ```javascript
     // bad
-    function concatenateAll() {
+    function concatenateAllStrings() {
```
bstashchuk, line 741: To make it clear that we concatenate strings in this example

README.md
```diff
@@ -854,6 +852,11 @@ Other Style Guides
     function f2(obj) {
       const key = Object.prototype.hasOwnProperty.call(obj, 'key') ? obj.key : 1;
     }
+
+    // best
+    function f2(obj) {
+      const key = Object.hasOwn(obj, 'key') ? obj.key : 1;
```
bstashchuk, line 858: As per one of the previous rules

README.md
```diff
@@ -1640,12 +1643,22 @@ Other Style Guides
 ## Variables
 
   <a name="variables--const"></a><a name="13.1"></a>
-  - [13.1](#variables--const) Always use `const` or `let` to declare variables. Not doing so will result in global variables. We want to avoid polluting the global namespace. Captain Planet warned us of that. eslint: [`no-undef`](https://eslint.org/docs/rules/no-undef) [`prefer-const`](https://eslint.org/docs/rules/prefer-const)
+  - [13.1](#variables--const) Always use `const` or `let` to declare variables. Not doing so will result in global variables in case strict mode is disabled. We want to avoid polluting the global namespace. Captain Planet warned us of that. eslint: [`no-undef`](https://eslint.org/docs/rules/no-undef) [`prefer-const`](https://eslint.org/docs/rules/prefer-const)
+
+    > Note: The entire contents of JavaScript modules are automatically in [strict mode](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Strict_mode) and therefore usage of the undeclared variables will not lead to creation of the global variables
```
bstashchuk, line 1648: It's important note. Please keep it

README.md
```diff
@@ -2091,7 +2100,7 @@ Other Style Guides
         let x = 1;
         break;
       case 2:
-        const y = 2;
+        const x = 2; // SyntaxError: Identifier 'x' has already been declared
```
bstashchuk, line 2103: Otherwise example doesn't make sense. Variables should have the same names in both cases

README.md
```diff
@@ -2156,7 +2165,7 @@ Other Style Guides
     const foo = a ? a : b;
     const bar = c ? true : false;
     const baz = c ? false : true;
-    const quux = a != null ? a : b;
+    const quux = (a !== undefined && a !== null) ? a : b;
```
bstashchuk, line 2168: This is full equivalent to a ?? b

README.md
```diff
@@ -374,9 +374,8 @@ Other Style Guides
     // bad
     const len = items.length;
     const itemsCopy = [];
-    let i;
 
-    for (i = 0; i < len; i += 1) {
+    for (let i = 0; i < len; i += 1) {
```
ljharb, line 378: this would make it a less "bad" example, so i'm not sure why it's an improvement

README.md
```diff
@@ -2208,25 +2217,32 @@ Other Style Guides
     > Why? It provides precision by distinguishing null/undefined from other falsy values, enhancing code clarity and predictability.
 
     ```javascript
-    // bad
-    const value = 0 ?? 'default';
-    // returns 0, not 'default'
-
-    // bad
-    const value = '' ?? 'default';
-    // returns '', not 'default'
```
bstashchuk, line 2217: ?? operator is not bad! It solves issues with || operator. 
That's why we should show it in brighter light

README.md
```diff
@@ -2208,25 +2217,32 @@ Other Style Guides
...
+    };
+    const age = user.age ?? 18; // 18
+
+    // good
+    const user = {
+      name: 'John',
+      age: 0
+    };
+    const age = user.age ?? 18; // 0
+    const defaultAge = user.age || 18 // 18
```
bstashchuk, line 2245: This better demonstrates difference between || and ??

README.md
```diff
@@ -710,9 +709,8 @@ Other Style Guides
     }
 
     // good
-    let test;
     if (currentUser) {
-      test = () => {
+      let test = () => {
```
ljharb, line 713: this change isn't wrong, but is immaterial to the example - why is it an improvement?

README.md
```diff
@@ -2688,10 +2704,10 @@ Other Style Guides
     // good
     $('#items')
       .find('.selected')
-        .highlight()
-        .end()
+      .highlight()
+      .end()
       .find('.open')
-        .updateCount();
+      .updateCount();
```
bstashchuk, line 2710: Better to place dots under each other

README.md
```diff
@@ -854,6 +852,11 @@ Other Style Guides
     function f2(obj) {
       const key = Object.prototype.hasOwnProperty.call(obj, 'key') ? obj.key : 1;
     }
+
+    // best
+    function f2(obj) {
+      const key = Object.hasOwn(obj, 'key') ? obj.key : 1;
+    }
```
ljharb, line 859: unfortunately until https://npmjs.com/airbnb-es-shims includes this polyfill, it can't be recommended here.

README.md
```diff
@@ -2948,11 +2966,12 @@ Other Style Guides
...
-      method: 'POST',
-      url: 'https://airbnb.com/',
-      data: { name: 'John' },
-    })
+    $
+      .ajax({
+        method: 'POST',
+        url: 'https://airbnb.com/',
+        data: { name: 'John' },
+      })
```
bstashchuk, line 2974: I am in doubts regarding this. Maybe leave like was before.

README.md
```diff
@@ -3743,10 +3776,10 @@ Other Style Guides
 
 **[⬆ back to top](#table-of-contents)**
 
-## ECMAScript 5 Compatibility
+## ECMAScript 6 Compatibility
 
-  <a name="es5-compat--kangax"></a><a name="26.1"></a>
-  - [27.1](#es5-compat--kangax) Refer to [Kangax](https://twitter.com/kangax/)’s ES5 [compatibility table](https://kangax.github.io/es5-compat-table/).
```
bstashchuk, line 3749: This link is not available anymore. Decided to add link to Can I use

README.md
```diff
@@ -1031,9 +1034,9 @@ Other Style Guides
     ```
 
   <a name="arrows--one-arg-parens"></a><a name="8.4"></a>
-  - [8.4](#arrows--one-arg-parens) Always include parentheses around arguments for clarity and consistency. eslint: [`arrow-parens`](https://eslint.org/docs/rules/arrow-parens)
+  - [8.4](#arrows--one-arg-parens) Always include parentheses around parameters for clarity and consistency. eslint: [`arrow-parens`](https://eslint.org/docs/rules/arrow-parens)
 
-    > Why? Minimizes diff churn when adding or removing arguments.
+    > Why? Minimizes diff churn when adding or removing parameters.
```
ljharb, line 1039: while this is technically correct, colloquially the two terms are synonymous, and the eslint docs describe the rule as "Require parentheses around arrow function arguments", so i don't think this change helps much.

README.md
```diff
@@ -1194,8 +1198,7 @@ Other Style Guides
 
     const luke = new Jedi();
 
-    luke.jump()
-      .setHeight(20);
+    luke.jump().setHeight(20);
```
ljharb, line 1201: this is intended to show an example of chaining.

README.md
```diff
@@ -1326,7 +1329,7 @@ Other Style Guides
     ```
 
   <a name="modules--no-wildcard"></a><a name="10.2"></a>
-  - [10.2](#modules--no-wildcard) Do not use wildcard imports.
+  - [10.2](#modules--no-wildcard) Do not use wildcard imports unless you have multiple named exports and want to import all of them as single object.
```
ljharb, line 1332: no, there are no circumstances, EVER, where you should use wildcard imports.

ljharb (changes requested):

README.md
bstashchuk: But this point [4.6](https://github.com/airbnb/javascript/blob/master/README.md#arrays--mapping) is about conversion of other iterables (except arrays) to arrays. And both examples show that
```
// bad
const baz = [...foo].map(bar);

// good
const baz = Array.from(foo, bar);
```
It doesn't make sense to make new array from other array using Array.from.
We should simply use map for arrays

`const baz = foo.map(bar)`

And [11.1](https://github.com/airbnb/javascript/blob/master/README.md#iterators--nope) clearly tells that.

That's why I wanted to emphasise that such approach should not be used for arrays but for other iterables .

ljharb: It's certainly not the clearest way to map one array to another, but there's nothing wrong with doing it. Imagine an argument that could be an array OR a Set - it's fine to put it in Array.from.

README.md
```diff
@@ -710,9 +709,8 @@ Other Style Guides
     }
 
     // good
-    let test;
     if (currentUser) {
-      test = () => {
+      const test = () => {
```
bstashchuk, line 713: Actually here I wanted to use `const` inside of the `if` block.

I thought that this point [7.4](https://github.com/airbnb/javascript/blob/master/README.md#functions--note-on-blocks) should answer the question -
_What if I need to create a function in the block?_

Answer: In such case don't use function declaration, create a variable using const (because in this good example we don't assign new value and therefore should use const) in the block and assign a function to it instead.

If I misunderstood idea, let's keep it as it is now.

README.md
```diff
@@ -1479,7 +1482,7 @@ Other Style Guides
 ## Iterators and Generators
 
   <a name="iterators--nope"></a><a name="11.1"></a>
-  - [11.1](#iterators--nope) Don’t use iterators. Prefer JavaScript’s higher-order functions instead of loops like `for-in` or `for-of`. eslint: [`no-iterator`](https://eslint.org/docs/rules/no-iterator) [`no-restricted-syntax`](https://eslint.org/docs/rules/no-restricted-syntax)
+  - [11.1](#iterators--nope) Don’t use iterators. Prefer JavaScript’s higher-order functions instead of loops like `for`, `for-in` or `for-of`. eslint: [`no-iterator`](https://eslint.org/docs/rules/no-iterator) [`no-restricted-syntax`](https://eslint.org/docs/rules/no-restricted-syntax)
```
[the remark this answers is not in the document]
bstashchuk: Yes, I agree. But maybe it does make sense to mention `for` loop somewhere here in this section as well (probably in the Note or separate bullet point)?
Because it is really very often used to iterate over Arrays and other iterables instead of array methods.

README.md
```diff
@@ -1640,12 +1643,22 @@ Other Style Guides
 ## Variables
 
   <a name="variables--const"></a><a name="13.1"></a>
-  - [13.1](#variables--const) Always use `const` or `let` to declare variables. Not doing so will result in global variables. We want to avoid polluting the global namespace. Captain Planet warned us of that. eslint: [`no-undef`](https://eslint.org/docs/rules/no-undef) [`prefer-const`](https://eslint.org/docs/rules/prefer-const)
+  - [13.1](#variables--const) Always use `const` or `let` to declare variables. Not doing so will result in global variables in case strict mode is disabled. We want to avoid polluting the global namespace. Captain Planet warned us of that. eslint: [`no-undef`](https://eslint.org/docs/rules/no-undef) [`prefer-const`](https://eslint.org/docs/rules/prefer-const)
```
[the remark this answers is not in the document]
bstashchuk: All right.
Instead of _**in case strict mode is disabled**_ we can write _**in case strict mode is not enabled**_

Just tell me if we should rephrase it or simply remove this strict mode mention and keep it as it is?

README.md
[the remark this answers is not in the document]
bstashchuk: You are right. 
Should we write the note like this? 

Note: The entire contents of modules `(import/export)` are automatically in [strict mode](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Strict_mode) and therefore usage of the undeclared variables will not lead to creation of the global variables

Or just keep without note as it is now?

README.md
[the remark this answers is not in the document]
bstashchuk: Because it demonstrates which error exactly appears if we don't put case bodies in the blocks {}.
And existing note tells about that
`This causes problems when multiple case clauses attempt to define the same thing`

So, in both examples we have to declare same variables in different `case` blocks to demonstrate the difference.

README.md
```diff
@@ -2455,11 +2471,11 @@ Other Style Guides
 
     ```javascript
     // bad
-    const active = true;  // is current tab
+    const isActive = true;  // is current tab
 
     // good
     // is current tab
-    const active = true;
+    const isActive = true;
```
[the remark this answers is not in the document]
bstashchuk: First example is "bad", but second example is "good". 
And this guide recommends to add "is" or "has" to boolean variables.

Maybe change `active` to `isActive` only in the "good" example?
Or keep as is?

README.md
```diff
@@ -2700,14 +2716,16 @@ Other Style Guides
...
+      .data(data)
+      .enter()
+      .append('svg:svg')
+      .classed('led', true)
+      .attr('width', (radius + margin) * 2)
       .append('svg:g')
-        .attr('transform', `translate(${radius + margin}, ${radius + margin})`)
-        .call(tron.led);
+      .attr('transform', `translate(${radius + margin}, ${radius + margin})`)
+      .call(tron.led);
```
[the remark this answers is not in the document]
bstashchuk: All right. Maybe add a Note with recommendation when to add additional indentation before chained methods calls? 
Because I am sure question about such formatting is asked by many who read those "good" examples.

README.md
```diff
@@ -2208,25 +2217,32 @@ Other Style Guides
...
-    // bad
-    const value = '' ?? 'default';
-    // returns '', not 'default'
+    // good
+    const value = '' ?? 'default'; // '', not 'default'
 
     // good
-    const value = null ?? 'default';
-    // returns 'default'
+    const value = null ?? 'default'; // 'default'
```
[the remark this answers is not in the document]
bstashchuk: Should I revert all changes which I made in the [15.9](https://github.com/airbnb/javascript/blob/master/README.md#nullish-coalescing-operator)?

Or keep previous "bad" examples and add few additional "good" examples which I added below (with comparison with || as well)?

README.md
bstashchuk: Yes, but in the next sentences on the [same page](https://eslint.org/docs/latest/rules/arrow-parens) they write

_Arrow functions can omit parentheses when they have exactly one parameter. In all other cases the parameter(s) must be wrapped in parentheses_

From my experience understanding of the difference between arguments and parameters is very important especially for beginners and it's not a good idea to use those terms interchangeably.

So, let's keep arguments?

bstashchuk (commented):
Thanks for the quick feedback @ljharb 
I hope we will be able to incorporate few of my changes.

README.md
ljharb: Fair enough.

README.md
ljharb: Sure, either of these are fine. how about adding your example, so both are present?

README.md
ljharb: I don't think it's super important - but it's fine to keep this change.

README.md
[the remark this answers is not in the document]
ljharb: not in a section about iterators tho.

if we mentioned it elsewhere, it'd be to avoid using loops at all, and that's a bit more controversial.

README.md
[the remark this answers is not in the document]
ljharb: i think it'd be best to leave it as is; there's no value in pointing out that the engine will fix this for you in some cases, since the lint rule and guide just tell you to never do it. simple is better :-)

README.md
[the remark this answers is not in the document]
ljharb: I'd say "The entire contents of ES Modules, and `class` bodies, are automatically in strict mode" if at all, but i think we just don't need to mention strict mode here.

README.md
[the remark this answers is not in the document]
ljharb: it's not just about the error tho - the concern applies even if the variable names are different, when you won't necessarily get the error you expect.

README.md
ljharb: it's bad when you care about falsiness.

README.md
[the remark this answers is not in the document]
ljharb: let's do that (keep previous, add additional good's)

README.md
ljharb: yes, i like this.

README.md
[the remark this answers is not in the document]
ljharb: sure, let's only change it in the good example.

README.md
ljharb: not in this case, because in jquery, `.end()` is like "stepping out"

README.md
```diff
@@ -3743,10 +3775,10 @@ Other Style Guides
 
 **[⬆ back to top](#table-of-contents)**
 
-## ECMAScript 5 Compatibility
+## ECMAScript 6 Compatibility
 
-  <a name="es5-compat--kangax"></a><a name="26.1"></a>
-  - [27.1](#es5-compat--kangax) Refer to [Kangax](https://twitter.com/kangax/)’s ES5 [compatibility table](https://kangax.github.io/es5-compat-table/).
```
ljharb, line 3749: existing links can't ever change, so please revert that part.

README.md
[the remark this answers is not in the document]
bstashchuk: I added some additional comments which highlight that as well.
What do you think?

README.md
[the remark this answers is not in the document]
bstashchuk: I restore previous version. If you have thought about possible note here please let me know

README.md
bstashchuk: Which link do you mean? Link to the https://twitter.com/kangax/ ?

We have already replaced 
_https://kangax.github.io/es5-compat-table_
with 
_https://compat-table.github.io/compat-table/es5/_

Do we still need to mention here Kangax?

ljharb: The anchor link - `es5-compat--kangax` has to work for eternity (as should all URLs)

bstashchuk:
@ljharb I am still thinking about [19.6](https://github.com/airbnb/javascript/blob/094eba3ae99c0bb6c4181e718d6b2bea9d2995d4/README.md#whitespace--chains)

There are actually general ESLint rules regarding method chains formatting but for some reason all examples are about jQuery and DOM manipulation.

No pure JS examples of methods chaining for arrays or some objects manipulation.

May I reduce quantity of the jQuery examples (or make them less confusing due to the long chains of method calls with different indents) and add some additional examples with bad/good formatting?

README.md
```diff
@@ -3743,10 +3788,10 @@ Other Style Guides
 
 **[⬆ back to top](#table-of-contents)**
 
-## ECMAScript 5 Compatibility
+## ECMAScript 6 Compatibility
```
ljharb, line 3791:
```suggestion
## ECMAScript 5 Compatibility
```
otherwise this is breaking a link

README.md
```diff
@@ -46,7 +46,7 @@ Other Style Guides
   1. [Accessors](#accessors)
   1. [Events](#events)
   1. [jQuery](#jquery)
-  1. [ECMAScript 5 Compatibility](#ecmascript-5-compatibility)
+  1. [ECMAScript 5 Compatibility](#es5-compat--kangax)
```
ljharb, line 49:
```suggestion
  1. [ECMAScript 5 Compatibility](#ecmascript-5-compatibility)
```

bstashchuk: I already changed anchor link to the previous `#es5-compat--kangax` as you suggested before.
So, no need to change to the `#ecmascript-5-compatibility` here

README.md
```diff
@@ -3406,7 +3451,7 @@ Other Style Guides
     this.firstName = 'Panda';
 
     // good, in environments where WeakMaps are available
-    // see https://kangax.github.io/compat-table/es6/#test-WeakMap
+    // see https://compat-table.github.io/compat-table/es6/ -> Built-ins -> WeakMap
```
bstashchuk, line 3454: Found another broken link. Can we change it like this?

ljharb: sure!

README.md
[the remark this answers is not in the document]
bstashchuk: @ljharb this is actually last thing which requires your approval.
If you are happy with those comments which I added in this example, you could finalize and merge this PR.
Bogdan

ExarAlan:
ni

bstashchuk:
Hi @ljharb ,
let me know if something else is needed to change in this PR.
It seems all questions were resolved.
Bogdan

anandfresh (approved):

README.md
```diff
@@ -3190,6 +3221,20 @@ Other Style Guides
...
+      secondVariable,
+      thirdVariable
+    } from './someModule';
+
+    // good
+    import {
+      firstVariable,
+      secondVariable,
+      thirdVariable,
+    } from './someModule';
```
ljharb, line 3237: these are identical. why is one bad and the other good?

ljharb (changes requested):

README.md
arturkhm: Trailing comma in good example

README.md
```diff
@@ -1646,6 +1653,14 @@ Other Style Guides
     // bad
     superPower = new SuperPower();
 
+    // bad
+    function foo() {
+      bar = 10 // bar will appear in the global scope
+    }
+
+    foo()
+    console.log(bar) // 10
```
```suggestion
    console.log(bar); // 10
```

README.md
```diff
@@ -1646,6 +1653,14 @@ Other Style Guides
     // bad
     superPower = new SuperPower();
 
+    // bad
+    function foo() {
+      bar = 10 // bar will appear in the global scope
+    }
+
+    foo()
```
```suggestion
    foo();
```

README.md
```diff
@@ -1646,6 +1653,14 @@ Other Style Guides
     // bad
     superPower = new SuperPower();
 
+    // bad
+    function foo() {
+      bar = 10 // bar will appear in the global scope
```
```suggestion
      bar = 10; // bar will appear in the global scope
```

README.md
```diff
@@ -1493,18 +1500,18 @@ Other Style Guides
...
     numbers.forEach((num) => {
       sum += num;
     });
-    sum === 15;
+    sum === 15; // true
 
     // best (use the functional force)
     const sum = numbers.reduce((total, num) => total + num, 0);
-    sum === 15;
+    sum === 15; // true
```
ljharb, line 1514: these are implicit.

README.md
```diff
@@ -1646,6 +1653,14 @@ Other Style Guides
     // bad
     superPower = new SuperPower();
 
+    // bad
+    function foo() {
+      bar = 10; // bar will appear in the global scope
```
ljharb, line 1658: only in sloppy mode; worth calling out.

README.md
```diff
@@ -3861,7 +3906,7 @@ Other Style Guides
   - [Latest ECMA spec](https://tc39.github.io/ecma262/)
   - [ExploringJS](https://exploringjs.com/)
   - [ES6 Compatibility Table](https://compat-table.github.io/compat-table/es6/)
-  - [Comprehensive Overview of ES6 Features](https://web.archive.org/web/20240404212626/http://es6-features.org/)
+  - [Comprehensive Overview of ES6 Features](http://es6-features.org/)
```
ljharb, line 3909: please revert this; http://es6-features.org is now some kind of russian ad site.

ljharb (changes requested):

PR #2946, opened 2024-03-16
chore: fix typo

avoidaway:


ljharb (changes requested):
We can’t remove the old link ever, and the typo is already fixed in the actual document.

PR #2965, opened 2024-04-10
feat!: use ecmaVersion latest

aladdin-add:
The following code is often used when using @eslint/eslintrc with `eslint-config-airbnb-base`:

```js
// mimic CommonJS variables -- not needed if using CommonJS
const __filename = fileURLToPath(import.meta.url);
const __dirname = path.dirname(__filename);
const compat = new FlatCompat({baseDirectory: __dirname, recommendedConfig: pluginJs.configs.recommended});
```

However, `import.meta` is not supported in ecmaVersion: '2018'. This change allows users to be able to use it that way.

### DIFF
--- a/packages/eslint-config-airbnb-base/index.js
+++ b/packages/eslint-config-airbnb-base/index.js
@@ -10,7 +10,7 @@ module.exports = {
     './rules/strict',
   ].map(require.resolve),
   parserOptions: {
-    ecmaVersion: 2018,
+    ecmaVersion: 'latest',
     sourceType: 'module',
   },
   rules: {},


aladdin-add:
related: https://github.com/eslint/create-config/issues/51

packages/eslint-config-airbnb-base/index.js
```diff
@@ -10,7 +10,7 @@ module.exports = {
     './rules/strict',
   ].map(require.resolve),
   parserOptions: {
-    ecmaVersion: 2018,
+    ecmaVersion: 'latest',
```
ljharb, line 13: this is a change we can't make unless babel-preset-airbnb is handling it by default. unfortunately i no longer maintain it, so I'm not familiar with what the latest version supports.

ljharb (changes requested):

packages/eslint-config-airbnb-base/index.js
aladdin-add: seems it's no longer actively maintained - the last version was released ~4 years ago.
image

ljharb: accurate. but this is airbnb's eslint config, so my hands are kind of tied :-/

PR #2993, opened 2024-05-31
Update README.md && Attempt to make code concise.

aadil42:
```javascript
 function dogs(x) {
      if (x) {
        if (z) {
          return y;
        }
      } else {
        return z;
      }
    }
``` 

**We can shorten it like this so it's easier to read.**

```javascript
  function dogs(x) {
     if(x && z) return y;
     if(!x) return z;
  }
```

### DIFF
--- a/README.md
+++ b/README.md
@@ -2335,13 +2335,8 @@ Other Style Guides
 
     // good
     function dogs(x) {
-      if (x) {
-        if (z) {
-          return y;
-        }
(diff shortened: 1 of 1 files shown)


ljharb (changes requested):
This is both much harder to read, and violates our own style guide in multiple ways.

aadil42:
Okay cool.

PR #3054, opened 2024-10-14
fix: mistake in allowArrowFunction

### LINKED ISSUE #2255: Configuration of react/jsx-no-bind allows to use arrow functions, while the style guide strictly prohibits it
The current configuration of `react/jsx-no-bind` allows to use arrow functions as event handlers (and ignores the use of bind in DOM components too): https://github.com/airbnb/javascript/blob/master/packages/eslint-config-airbnb/rules/react.js#L104

image

However, here's what the style guide has to say bout this:

---

image

---

1. I agree with the guide's argument and belive that it would be best to update the rule configuration to match it. Using arrow function to initialize an event handler creates a brand new function on each render same as `bind()` does, which can lead to unnecessary re-renders.

2. To be honest, I'd even turn `ignoreDOMComponents` to `false` too — it is entirely unclear why DOM components deserve a different treatment. I'd say the reasoning from the style guide fully applies to them too.

Hamza-jamshaid:
## What does this PR do?
This PR fixes the incorrection in allowArrowFuntion "true" to "false" in react.js.

Fixes #2255

image

##Type of change
Bug fix (non-breaking change which fixes an issues).

##How should this be tested?
-[]Go to react.js file
-[]Check the line no 112 that bug in allowArrowFunction "false"

##Mandatory Tasks

-[X] Make sure you have self-reviewed the code.A decent size PR without self-review might be rejected.

### DIFF
--- a/packages/eslint-config-airbnb/rules/react.js
+++ b/packages/eslint-config-airbnb/rules/react.js
@@ -109,7 +109,7 @@ module.exports = {
     // https://github.com/jsx-eslint/eslint-plugin-react/blob/master/docs/rules/jsx-no-bind.md
     'react/jsx-no-bind': ['error', {
       ignoreRefs: true,
-      allowArrowFunctions: true,
+      allowArrowFunctions: false,
       allowFunctions: false,
       allowBind: false,
       ignoreDOMComponents: true,


ljharb:
See my comment here: https://github.com/airbnb/javascript/issues/2255#issuecomment-1379572865

In other words, arrow functions are fine in an HTML element, and in a custom component, are only a problem if the custom component isn't just stamping them onto an HTML element.

PR #3060, opened 2024-11-09
Bug Fix: Use HTML character entities to display JSX code

DetrojaRadhey:
I have used HTML character entities such as "&lt;" and "&gt;" to display JSX code in a markdown format without being rendered as HTML.

And It works perfect in both cases.

In readme file:
{F5E8D47C-64BC-43AB-B636-C59C641553B9}

In webpage:
{E60F6308-8CAB-4D57-8BF4-8F5AA6716862}

### DIFF
--- a/react/README.md
+++ b/react/README.md
@@ -245,16 +245,16 @@ This style guide is mostly based on the standards that are currently prevalent i
 
     ```jsx
     // bad
-    <Foo bar='bar' />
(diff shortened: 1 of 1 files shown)


ljharb (changes requested):
This causes them to render in the markdown incorrectly, as well as being much less ergonomic to edit.

Issue #2961, opened 2024-04-06
eslint v9 support


ljharb:
(see also #2478, https://github.com/airbnb/javascript/issues/2223, https://github.com/airbnb/javascript/issues/2036, https://github.com/airbnb/javascript/issues/1834, https://github.com/airbnb/javascript/issues/1447)

eslint v9 was released on 2024.04.05.

The current plan is, if possible, to release a new semver-minor of both configs, that supports `7 || 8 || 9`.

Note that this is _different_ from whether flat config is supported or not by these plugins, and from whether this project will convert to flat config.

The timeframe for eslint v9 support is "not as soon as you would like, but as soon as we can".
Please accept this, and please add no noise to this thread beyond emoji reactions - distractions won't help anyone get the work done any faster.

### **Please** do not file additional issues and do not file pull requests attempting to add support - please, instead, continue to use eslint v8 until such time as our configs support v9.

Peer Dep eslint combined v7, v8, and v9 support released:
 - [ ] [`eslint-plugin-import`](https://github.com/import-js/eslint-plugin-import/issues/2948)
 - [ ] `eslint-find-rules`
 - [ ] update the base config
 - [ ] publish the base config, semver-minor
 - [ ] [`eslint-plugin-react`](https://github.com/jsx-eslint/eslint-plugin-react/issues/3699)
 - [ ] `eslint-plugin-jsx-a11y`
 - [ ] `eslint-plugin-react-hooks`
 - [ ] update the main config
 - [ ] publish the main config, semver-minor

curdin:
Heads-up that the PR for `eslint-plugin-jsx-a11y` was closed in favour of https://github.com/jsx-eslint/eslint-plugin-jsx-a11y/pull/993. This has now been merged.

controversial:
Updates to the checklist above:
- eslint-plugin-react now has support for eslint 9 and flat config [[1]](https://github.com/jsx-eslint/eslint-plugin-react/releases/tag/v7.35.0)
- eslint-plugin-react-hooks has support for eslint 9 in its canary [[2]](https://github.com/facebook/react/pull/28773). They’re waiting for React 19 to ship (which is held up indefinitely by their issues with suspense orchestration) to release stable. Flat config works with the canary release.

eslint-plugin-jsx-a11y still has flat config support but not eslint 9 support; and there’s nothing completed for eslint-plugin-import or eslint-find-rules

oxyno-zeta:
Hello,

[...]

PR #3068, opened 2024-12-07
[eslint config] [deps] update `eslint-plugin-react-hooks`

ptmkenny:
[eslint-plugin-react-hooks had a new major version a couple months ago.](https://github.com/facebook/react/blob/main/packages/eslint-plugin-react-hooks/CHANGELOG.md)

This new version has several new useful checks, so it would be great if the config could be updated to use this version.

### DIFF
--- a/packages/eslint-config-airbnb/package.json
+++ b/packages/eslint-config-airbnb/package.json
@@ -78,7 +78,7 @@
     "eslint-plugin-import": "^2.30.0",
     "eslint-plugin-jsx-a11y": "^6.10.0",
     "eslint-plugin-react": "^7.36.1",
-    "eslint-plugin-react-hooks": "^4.6.2",
+    "eslint-plugin-react-hooks": "^5.1.0",
     "in-publish": "^2.0.1",
     "react": ">= 0.13.0",
     "safe-publish-latest": "^2.0.0",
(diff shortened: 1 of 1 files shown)


ljharb (changes requested):
you'll also need to change the peer deps entry.

ptmkenny:
Thanks for the quick feedback!  I updated the peer dependency.

anandfresh (approved):

ljharb (approved):

renchris:
How do we use this? The current release version is 19.0.4 from Dec 25, 2021.

ljharb:
@renchris you don't use it, until it's released, which it's not yet (it was only merged 2 days ago)

renchris:
With the current and last package release being a few years ago, what is a rough timeframe expectation for the next release containing this update to be?

ljharb:
There's no expected timeframe and no release schedule, and the time since the last release has no bearing on anything - no release has been needed since the last one, and the next one will include full eslint 9 support. feel free to follow #2961 for updates.

ptmkenny:
@renchris If you want to use this now, if you're using npm, you can install `eslint-plugin-react-hooks` 5 with `--force`.  That will override the v4 specified by the current release.

renchris:
Thank you, you two. I will override the peer dependency `eslint-plugin-react-hooks` version 5 with `--force` now and follow the https://github.com/airbnb/javascript/issues/2961 thread for the next eslint v9 version with this update as well

PR #3072, opened 2024-12-15
ISSUE#3032

### LINKED ISSUE #3032: Incorrect constant export example
> _What about all const variables? - This is unnecessary, so uppercasing should not be used for constants within a file. **It should be used for exported constants however.**_

There are two points:

1. Based on the above description the provided example appears incorrect
2. Also, Since const variables can't be changed through reassignment, this example (i.e., variable name) feels odd. 

Current example:
```
// bad
export const THING_TO_BE_CHANGED = 'should obviously not be uppercased';
```

Perhaps it could be:
```
// bad
export const some_exported_constant = 'should obviously be uppercased';
```

JASH9898567:
Made some changes such as-
1) export const MY_CONSTANT = 'should obviously not be uppercased' . In place of THING_TO_BE_CHANGED, it created unnecessary confusion as export const is not going to change across different files. 

2) Another is an example of //good coding in README - export const MY_EXPORTED_FUNCTION = () => {
  return 'hello function';
}; . As no example of this was already available in file.

Fixes #3032.

### DIFF
--- a/README.md
+++ b/README.md
@@ -3184,9 +3184,9 @@ Other Style Guides
     );
 
     //good
-    export const MY_EXPORTED_FUNCTION = () => {
-  return 'hello function';
-};
+    export function MY_EXPORTED_FUNCTION() {
+      return 'hello function';
+    };
 
     // good (note that a comma must not appear after a "rest" element)
     createHero(
--- a/README.md
+++ b/README.md
@@ -3183,6 +3183,11 @@ Other Style Guides
       inventorOf,
     );
 
+    //good
+    export const MY_EXPORTED_FUNCTION = () => {
+  return 'hello function';
+};
+
     // good (note that a comma must not appear after a "rest" element)
     createHero(
       firstName,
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -3183,6 +3183,11 @@ Other Style Guides
       inventorOf,
     );
 
+    //good
+    export const MY_EXPORTED_FUNCTION = () => {
+  return 'hello function';
+};
```
ljharb: the indentation on this is incorrect, and there's no reason for this to be an arrow function, so it shouldn't be:
```suggestion
    export function MY_EXPORTED_FUNCTION() {
      return 'hello function';
    };
```

but also this example has nothing to do with dangling commas, which is the section you've placed it in

ljharb (changes requested):

README.md
ljharb: this still doesn't make any sense in this section.

Scutua (approved):

PR #3080, opened 2025-01-25
Support flat config

ghost:
This is a pull request to support flat config for eslint-config-airbnb-base and eslint-config-airbnb and do some rewriting.

Example:
  ```js
  // eslint.config.cjs
  const airbnb = require('eslint-config-airbnb/flat');

  module.exports = [
    ...airbnb,
    // ......
  ];
  ```

  ```js
  // eslint.config.mjs
  import airbnb from 'eslint-config-airbnb/flat';

  export default [
    ...airbnb,
    // ......
  ];
  ```


ghost:
By the way, I haven't fully tested it yet. So I'm not sure if it will really work.

ljharb (changes requested):
I’m confused, what is the “base” folder? The existing files directly inside `rules` will remain where they are, and shouldn’t be duplicated.

ygysn:
Oh, At first I thought of trying to separate the rules into a special file. I didn't think it would cause redundancies. Sorry

ljharb:
Why was this closed and replaced with a duplicate #3081?

ghost:
I tried updating, I'm new to github so I don't really understand how this works. Sorry.

PR #2706, opened 2023-01-06
fixed execSync issue in windows #2700

kapilsriram:

PR #3100, opened 2025-03-04
fix whitespace module execSync under Windows with no shebang

### LINKED ISSUE #2700: Unexpected end of JSON input
I am facing similar issue as mentioned in  https://github.com/airbnb/javascript/issues/2540

whenever I run "eslint ." i am thrown "Unexpected end of JSON input"
image

zaycker:
fixes #2700

In case the eslint version is 8, `whitespace.js` file tries to execute `whitespace-async.js` file.
`whitespace-async.js` file starts with shebang header `#!/usr/bin/env node` which doesn't work under Windows.
It cannot be executed right away without definition of interpreter.
In case we try to execute it with `execSync` or any other fork tool it opens up a dialog for choosing a program to open the .js file.
It requires explicit definition of `node` interpreter at the place of execution.
`windowsHide` option prevents opening of `cmd` window during execution and it is ignored under non-Windows systems.

there is another PR which is outdated. made a fresh one

### DIFF
--- a/packages/eslint-config-airbnb-base/whitespace.js
+++ b/packages/eslint-config-airbnb-base/whitespace.js
@@ -53,11 +53,10 @@ if (CLIEngine) {
   const isWindows = process.platform === 'win32';
   const whitespaceAsyncPath = path.join(__dirname, 'whitespace-async.js');
   const execSyncCommand = isWindows ? `node ${whitespaceAsyncPath}` : whitespaceAsyncPath;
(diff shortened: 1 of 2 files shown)


ljharb:
This is a duplicate of #2706. Please don't file duplicate PRs - if one is outdated, you can comment on that PR with your branch and maintainers can pull it in.

anandfresh (approved):

PR #3102, opened 2025-03-05
Update base.js

anuragchvn-blip:
Modify packages/eslint-config-airbnb/base.js to include the new rule.


ljharb (changes requested):
The rule (that you tried to add in #3101) doesn't exist yet, and this PR is nonsensical on its own. Please don't submit unsolicited PRs in the future, especially two of them.

PR #3104, opened 2025-03-11
Update package.json

arj24092003:


ljharb (changes requested):
are you trying to make it work in windows? rmdir doesn’t exist on any other platform, so this would break everything.

arj24092003:
Ok thank you so much

ljharb:
If that's the goal, then you could use the `rimraf` package to do `rm -rf` cross-platform.

anandfresh (approved):

PR #3120, opened 2025-04-18
Update README.md, update an outdated link

NicholasWilsonDEV:
devchat.tv changes their domain to topenddevs.com
I updated the redirected links to their current destination.

### DIFF
--- a/README.md
+++ b/README.md
@@ -3932,7 +3932,7 @@ Other Style Guides
 **Podcasts**
 
   - [JavaScript Air](https://javascriptair.com/)
-  - [JavaScript Jabber](https://devchat.tv/js-jabber/)
+  - [JavaScript Jabber](https://topenddevs.com/podcasts/javascript-jabber)
 
 **[⬆ back to top](#table-of-contents)**
 


README.md
```diff
@@ -3932,7 +3932,7 @@ Other Style Guides
 **Podcasts**
 
   - [JavaScript Air](https://javascriptair.com/)
-  - [JavaScript Jabber](https://devchat.tv/js-jabber/)
+  - [JavaScript Jabber](https://topenddevs.com/podcasts/javascript-jabber)
```
anandfresh, line 3935: https://devchat.tv/js-jabber/ is redirecting to https://topenddevs.com/podcasts/javascript-jabber link only , do we still need this change ?

anandfresh (changes requested):

README.md
ljharb: it's fine to update to include the redirect, but it's also find to just let the redirect continue, since in theory it'll never break.

anandfresh: It will not break , but what if 'https://devchat.tv/js-jabber/' redirects to a different link in future ?

PR #3123, opened 2025-05-02
Support flat config - rebased

chrstrock:
Updated Tamiroh's PR to rebase from the master branch.

This is to support the flat config of ESLint 9.


ljharb:
There is absolutely no benefit for opening up an extra PR to do a rebase that a maintainer can take care of directly on the original PR.

PR #3139, opened 2025-07-10
Update LICENSE.md

mono7272:


ljharb:
The end year is optional, and should always be omitted to avoid yearly churn to increment it.

PR #3140, opened 2025-07-24
feat: add ESLint config for markdown linting with eslint-plugin-markdown

AbhishekG-Codes:
Points to be noted:

- Some examples in the README are intentionally invalid (for illustration purposes).
- With this setup, we can start linting the "good" examples to ensure they stay clean and consistent.

Looking ahead:

- We can explore ways to skip linting the intentionally "bad" examples either by using eslint-disable comments or setting up smarter filters.

### DIFF
--- a/eslint.config.js
+++ b/eslint.config.js
@@ -1,0 +1,26 @@
+// eslint.config.js
+const markdown = require('eslint-plugin-markdown');
+
+module.exports = [
+  {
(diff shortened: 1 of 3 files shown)


ljharb (changes requested):
Adding a new plugin is both a breaking change and a big lift - we won't be doing that.

PR #3146, opened 2025-08-15
Fix array index out of bounds by changing someStack.length to someStack.length  - 1 in the Arrays section

nafasebra:
Changed array.length to `someStack.length - 1` to prevent an array index out of bounds error. The original line was causing an error, likely because the intended behavior was to access the last element or iterate safely within the array's bounds. This adjustment ensures proper array access without exceeding its limits.

### DIFF
--- a/README.md
+++ b/README.md
@@ -361,7 +361,7 @@ Other Style Guides
     const someStack = [];
 
     // bad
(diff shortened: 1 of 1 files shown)


ljharb:
There's no such thing as an  "array out of bounds error", and this is attempting to push, not mutate.

PR #3153, opened 2025-08-22
docs: add note about ESLint --ext for .jsx files (see eslint/eslint#1402)

### LINKED ISSUE #1402: Documentation: react/jsx-filename-extension can break checks without --ext arguments
Adding the `react/jsx-filename-extension` is just fine as a rule, but adding some direction that you will need to add the `--ext=js --ext=jsx` arguments to the eslint cli for it to work. It might cause issues for users who are not aware of this when they start using the `eslint-preset-airbnb` package. Adding a line to the README might be helpful, and I would be happy to do so. 😄 

Ideally, you could configure this as an option in `.eslintrc` but that is not currently available: 
https://github.com/eslint/eslint/issues/801

prateekbisht23:
By default, ESLint only lints `.js` files. This PR updates the README(s) to note
that users need to pass `--ext .js,.jsx` (or `.tsx` for TypeScript) when running
ESLint so that rules like `react/jsx-filename-extension` are applied correctly.

Also updated package.json scripts to demonstrate this usage.

Fixes #1402

### DIFF
--- a/packages/eslint-config-airbnb-base/README.md
+++ b/packages/eslint-config-airbnb-base/README.md
@@ -62,6 +62,16 @@ Our default export contains all of our ESLint rules, including ECMAScript 6+. It
 
 2. Add `"extends": "airbnb-base"` to your .eslintrc.
 
+> **Note**: ESLint only lints `.js` files by default.
+
+  If your project uses `.jsx` (or `.tsx` with TypeScript), you need to pass extensions to the CLI:
+
+  ```sh
+  eslint . --ext .js, .jsx, .mjs
+  ```
+
+  Without this, JSX-related rules will not apply to `.jsx` files.
+
 ### eslint-config-airbnb-base/legacy
 
 Lints ES5 and below. Requires `eslint` and `eslint-plugin-import`.
--- a/packages/eslint-config-airbnb/README.md
+++ b/packages/eslint-config-airbnb/README.md
@@ -58,6 +58,15 @@ If you don't need React, see [eslint-config-airbnb-base](https://npmjs.com/eslin
 
 2. Add `"extends": "airbnb"` to your `.eslintrc`
 
+> **Note**: ESLint only lints `.js` files by default.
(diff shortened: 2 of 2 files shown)


package.json
```diff
@@ -8,6 +8,7 @@
     "install:config": "cd packages/eslint-config-airbnb && npm prune && npm install",
     "install:config:base": "cd packages/eslint-config-airbnb-base && npm prune && npm install",
     "lint": "markdownlint --config linters/.markdownlint.json README.md */README.md",
+    "lint:js": "eslint -c linters/.eslintrc . --ext .js --ext .jsx",
```
ljharb: why are we adding a script that's not being run?

packages/eslint-config-airbnb-base/README.md
```diff
@@ -62,6 +62,16 @@ Our default export contains all of our ESLint rules, including ECMAScript 6+. It
...
 
+⚠️ **Note**: ESLint only lints `.js` files by default.  
+  
+  If your project uses `.jsx` (or `.tsx` with TypeScript), you need to pass extensions to the CLI:  
+
+  ```sh
+  eslint . --ext .js,.jsx
+  ```
+
+  Without this, JSX-related rules will not apply to `.jsx` files.
```
ljharb: let's also include `.mjs` in this list.

package.json
prateekbisht23: You're right, the lint:js script isn’t currently being run!!
Would you prefer I remove it for now, or update the existing lint script to include the same logic (with the --ext flags for .js, .jsx, and .mjs) directly?

Happy to adjust based on your preference!!

packages/eslint-config-airbnb-base/README.md
prateekbisht23: Sure!!

prateekbisht23:
Hey!! Can you approve the pull request or any changes you want!!

PR #3155, opened 2025-08-25
Switched markdownlint config to stable MD rule IDs

himaenshuu:
Made the install scripts Windows-friendly by replacing Unix-only rm -rf with cross-platform rimraf and adding rimraf as a dev dependency. Cleaned and modernized markdownlint configuration by switching to canonical MD rule IDs and removing extraneous comment keys, preserving original linting intent. Lint now runs reliably across environments without behavioral changes.

### DIFF
--- a/linters/.markdownlint.json
+++ b/linters/.markdownlint.json
@@ -1,155 +1,42 @@
 {
-  "comment": "Be explicit by listing every available rule. https://github.com/DavidAnson/markdownlint/blob/master/doc/Rules.md",
-  "comment": "Note that there will be numeric gaps, not every MD number is implemented in markdownlint.",
-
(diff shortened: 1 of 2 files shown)


ljharb (changes requested):
I'm fine with the rimraf usage, but removing the comments seems like a massive decrease in DX, and changing to rule IDs seems like it is a minor decrease in DX. Why is that needed at all? Those rule names can't change without a semver-major just like the IDs - they seem equally stable.

ljharb:
Deleting the fork renders the PR unrecoverable :-/ please don't do that in the future.

himaenshuu:
> Deleting the fork renders the PR unrecoverable :-/ please don't do that in the future.

actually it was a mistake , what now ? i don't know

PR #3174, opened 2025-09-15
docs: add Comments section to CSS-in-JavaScript Style Guide

### LINKED ISSUE #1506: Clarification for Inline Comments
Thanks for the excellent style guide!

One thing that is not explicitly defined by the guide is inline comments. Can a section on this topic be added to the "Comments" section block, please?

For example,
```
let foo = 145; // This is modified below in the "baz()" function
```

Is this a legal line in the airbnb style?

And the follow-up question: if it is not legal, then I assume that the correct fix to have it be moved above, like this?

```
// This is modified below in the "baz()" function
let foo = 145;
```

Yet another follow-up question: if inline comments are legal, then what is the correct fix for in-line comments that extend past 100 character? Like the following:

```
// This is a comment describing the following block of variable initializations
let foo = 123; // used in various contexts, but mostly for soccer games, basketball, and football games
let bar = 456; // used only for horse races; it is explicitly reset at the beginning of the "endRace()" function
let baz = 789; // used only for car races; it will be overwritten once we reach the "oilChange()" function
```

If it were moved above, it would overlap with the block comment header. So then does it make the most sense to move them down to the next line, like this?

```
// This is a comment describing the following block of variable initializations
let foo = 123;
// used in various contexts, but mostly for soccer games, basketball, and football games
let bar = 456;
// used only for horse races; it is explicitly reset at the beginning of the "endRace()" function
let baz = 789;
// used only for car races; it will be overwritten once we reach the "oilChange()" function
```

prateekbisht23:
## What’s Changed

- Added a new **Comments** section to the CSS-in-JavaScript Style Guide
- Documented guidance on:
  - Allowing inline comments
  - Moving comments above lines when clearer
  - Handling long inline comments ( >100 chars )
  - Examples with // bad and // good patterns
  - Recommendation to refactor if excessive comments are needed

## Why

Inline comments weren’t explicitly covered in the style guide.  
This addition clarifies acceptable patterns and improves consistency with existing formatting.

---

fixes: #1506

### DIFF
--- a/css-in-javascript/README.md
+++ b/css-in-javascript/README.md
@@ -427,6 +427,48 @@
     });
     ```
 
+## Comments
+
+  - Inline comments (`// comment`) are allowed.
+
+    ```js
+    // good
+    let foo = 145; // This is modified below in the "baz()" function
+    ```
+
+  - Comments may also be placed above the line if it improves clarity.
+
+    ```js
+    // good
+    // This is modified below in the "baz()" function
+    let foo = 145;
+    ```
+
+  - When inline comments extend the line length beyond the maximum (100 characters), move the comment to the line above to preserve readability.
+
+    ```js
+    // bad
+    // This is a comment describing the following block of variable initializations
+    let foo = 123; // used in various contexts, but mostly for soccer games, basketball, and football games
+    let bar = 456; // used only for horse races; it is explicitly reset at the beginning of the "endRace()" function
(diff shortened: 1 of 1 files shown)


prateekbisht23:
@ljharb can you review the PR!!

anandfresh (approved):

ljharb:
Why is this unique to CSS in JS?

PR #3185, opened 2025-10-07
README.md

Riteshmaurya1:
Here i added TDZ inside the hoisting part eith 14.1.1

### DIFF
--- a/README.md
+++ b/README.md
@@ -1908,6 +1908,22 @@ Other Style Guides
     }
     ```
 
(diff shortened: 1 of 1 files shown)


anandfresh (approved):

ljharb:
We already link to https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Statements/let#temporal_dead_zone_tdz for this in the Hoisting section; I think that's enough.

PR #3221, opened 2026-01-25
Docs: clarify cases where traditional functions are preferred

Manvendra0023:
This PR adds a small clarification to the Arrow Functions section describing cases where traditional functions are still preferred, such as function hoisting, dynamic `this` binding, and use of the `arguments` object.

The goal is to improve clarity without changing the existing recommendation.

### DIFF
--- a/README.md
+++ b/README.md
@@ -955,18 +955,16 @@ Other Style Guides
   >
   > **Prefer traditional functions when:**
   >
-  > - You rely on **function hoisting**, such as calling a function before its definition.
+  > - You are defining top-level, reusable functions rather than inline callbacks.
   > - You need a **dynamic `this` binding**, for example in object methods or DOM event handlers.
   > - You require access to the **`arguments` object**, which is not available in arrow functions.
   >
   > **Examples:**
   >
   > ```js
-  > // Function hoisting
-  > initialize();
-  >
-  > function initialize() {
-  >   // setup logic
+  > // Named function for reuse and clearer stack traces
+  > function formatUserName(user) {
+  >   return `${user.firstName} ${user.lastName}`;
   > }
   > ```
   >
--- a/README.md
+++ b/README.md
@@ -949,7 +949,7 @@ Other Style Guides
 
     > Why not? If you have a fairly complicated function, you might move that logic out into its own named function expression.
 
-  > **Note:** Arrow functions are intended primarily for short, anonymous callbacks.
+  > **Note:** Arrow functions are intended primarily for inline, anonymous callbacks.
   > Outside of these cases, prefer traditional function declarations or expressions
   > for clarity and consistency.
   >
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -949,6 +949,45 @@ Other Style Guides
 
     > Why not? If you have a fairly complicated function, you might move that logic out into its own named function expression.
 
+  > **Note:** While arrow functions are recommended for anonymous callbacks, traditional
+  > function declarations or expressions are still preferred in certain scenarios where
+  > their behavior is more appropriate.
+  >
+  > **Prefer traditional functions when:**
+  >
+  > - You rely on **function hoisting**, such as calling a function before its definition.
```
ljharb: Nobody following this guide must ever rely on hoisting.

ljharb (commented):
Arrow functions should only be used for inline callbacks. Everything else should use normal functions.

Manvendra0023:
Thanks for the clarification — that makes sense.

I agree that this guide should not encourage relying on hoisting, and that arrow functions should generally be preferred except for inline callbacks.

My intent here wasn’t to recommend hoisting as a practice, but to document why traditional functions still exist and what technical differences remain, for readers who may encounter them in real codebases.

I’m happy to remove or reword the hoisting mention to avoid implying it’s an acceptable pattern under this guide. Please let me know if you’d prefer the section to focus only on cases like named functions for debugging / stack traces, or if you’d rather drop the section entirely.

ljharb:
I think that instead of "here's when you shouldn't use arrow functions", i think more useful guidance is "here are the only places you should use arrow functions".

Manvendra0023:
That makes sense — I agree that framing this in terms of “where arrow functions should be used” is clearer and more in line with the guide’s intent.

I’ll update the section to focus on a short, explicit list of recommended arrow function use cases (e.g. inline callbacks), and remove or de-emphasise the “when not to use” framing.

I’ll push an update shortly — thanks for the guidance.

README.md
```diff
@@ -949,6 +949,47 @@ Other Style Guides
 
     > Why not? If you have a fairly complicated function, you might move that logic out into its own named function expression.
 
+  > **Note:** Arrow functions are intended primarily for short, anonymous callbacks.
```
```suggestion
  > **Note:** Arrow functions are intended primarily for inline, anonymous callbacks.
```

Manvendra0023:
Thanks for the suggestion! Applied 👍

README.md
```diff
@@ -949,6 +949,47 @@ Other Style Guides
...
+  >   return `${user.firstName} ${user.lastName}`;
+  > }
+  > ```
+  >
+  > ```js
+  > // Dynamic `this` binding (e.g. event handlers)
+  > const button = document.querySelector('button');
+  >
+  > button.addEventListener('click', function () {
+  >   this.classList.add('active');
```
ljharb, line 979: this is actually not a good idea; in event handlers, you should use the event object argument's `target` or `currentTarget` instead of `this`.

ljharb (changes requested):

PR #3244, opened 2026-03-10
Adding support for eslint 9

j-castellanos:
Giving my attempt at eslint 9 support 😅

### DIFF
--- a/packages/eslint-config-airbnb-base/package.json
+++ b/packages/eslint-config-airbnb-base/package.json
@@ -15,6 +15,7 @@
     "./rules/imports": "./rules/imports.js",
     "./rules/strict": "./rules/strict.js",
     "./rules/variables": "./rules/variables.js",
+    "./flat": "./flat.js",
     "./package.json": "./package.json"
   },
   "scripts": {
@@ -72,21 +73,22 @@
     "babel-preset-airbnb": "^4.5.0",
     "babel-tape-runner": "^3.0.0",
     "eclint": "^2.8.1",
-    "eslint": "^7.32.0 || ^8.2.0",
-    "eslint-find-rules": "^4.1.0",
+    "eslint": "^7.32.0 || ^8.2.0 || ^9.0.0",
+    "eslint-find-rules": "^5.0.0",
     "eslint-plugin-import": "^2.30.0",
     "in-publish": "^2.0.1",
     "safe-publish-latest": "^2.0.0",
     "tape": "^5.9.0"
   },
   "peerDependencies": {
-    "eslint": "^7.32.0 || ^8.2.0",
+    "eslint": "^7.32.0 || ^8.2.0 || ^9.0.0",
     "eslint-plugin-import": "^2.30.0"
   },
   "engines": {
     "node": "^10.12.0 || >=12.0.0"
   },
   "dependencies": {
-    "confusing-browser-globals": "^1.0.11"
+    "confusing-browser-globals": "^1.0.11",
+    "globals": "^15.14.0"
   }
 }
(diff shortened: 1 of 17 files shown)


j-castellanos:
Giving it a try @ljharb 🙏

j-castellanos:
@christianvuerings maybe you could take a look when you get a chance?

packages/eslint-config-airbnb-base/package.json
```diff
@@ -72,21 +73,22 @@
     "babel-preset-airbnb": "^4.5.0",
     "babel-tape-runner": "^3.0.0",
     "eclint": "^2.8.1",
-    "eslint": "^7.32.0 || ^8.2.0",
-    "eslint-find-rules": "^4.1.0",
+    "eslint": "^7.32.0 || ^8.2.0 || ^9.0.0",
+    "eslint-find-rules": "^5.0.0",
```
ljharb: eslint-find-rules v5 doesn't support eslint 7, so this will have to be `^4.1.0 || ^5.0.0`

packages/eslint-config-airbnb-base/package.json
```diff
@@ -72,21 +73,22 @@
...
+    "eslint": "^7.32.0 || ^8.2.0 || ^9.0.0",
     "eslint-plugin-import": "^2.30.0"
   },
   "engines": {
     "node": "^10.12.0 || >=12.0.0"
   },
   "dependencies": {
-    "confusing-browser-globals": "^1.0.11"
+    "confusing-browser-globals": "^1.0.11",
+    "globals": "^15.14.0"
```
ljharb: why is `globals` needed? eslint ships with that package and we should be using the same version it uses.

packages/eslint-config-airbnb/package.json
```diff
@@ -73,8 +76,8 @@
     "babel-preset-airbnb": "^4.5.0",
     "babel-tape-runner": "^3.0.0",
     "eclint": "^2.8.1",
-    "eslint": "^7.32.0 || ^8.2.0",
-    "eslint-find-rules": "^4.1.0",
+    "eslint": "^7.32.0 || ^8.2.0 || ^9.0.0",
+    "eslint-find-rules": "^5.0.0",
```
ljharb: same here

ljharb (changes requested):

packages/eslint-config-airbnb-base/package.json
j-castellanos: @ljharb without it being in the package.json we fail the no extraneous dependencies check due to the import. So I put it as a peer dependency. Otherwise I could either disable the rule or use the raw values instead of from globals. Let me know if you want a different change

j-castellanos:
@ljharb let me know your thoughts with the changes

j-castellanos:
@ljharb are the issues with the previous iteration resolved?

daltino (approved):
This PR adds support for ESLint 9 by updating the package version requirements, introducing a new `flat.js` configuration file for rule definitions, and adjusting associated test files for the new file. Implementation looks comprehensive and well-tested with no breaking changes identified. Looking good!

hamzaalikazmi:
just came here dont know what to do but will learn after a while

j-castellanos:
thank you @daltino hopefully it can be approved soon 🙏

j-castellanos:
@ljharb any way you can take a look again?

j-castellanos:
eslint 9 is about to hit EOL

j-castellanos:
@christianvuerings could you take a look at this?

PR #3280, opened 2026-07-03
Add example of array reference and mutation

filforopen-source:


vehicle
```diff
@@ -0,0 +1,6 @@
+const foo = [1, 2];
+const bar = foo;
+
+aar[0.6] = 9;
```
anandfresh, line 4: It is `bar[0.6]` and not `aar[0.6]`

vehicle
```diff
@@ -0,0 +1,6 @@
...
+
+console.log(foo[0.5], bar[0.6]); // => 9, 9
```
anandfresh, line 6: It should be `foo[0.6]` and not `foo[0.5]`

anandfresh (changes requested):

ljharb (changes requested):
i'm confused why we'd want an example of mutation, which should be avoided - either way, adding it in an extensionless file instead of the main guide doesn't make any sense.

